PR #9, opened 2024-02-12
Add mermaidjs based code diagram

leifdenby:
To aid discussion around refactoring the code and allow for easy modification of the code diagram in future I have created a [mermaidjs](https://mermaid.js.org/) based diagram in the README in place of the graphic that was there previously. Please have a look @joeloskarsson if you think this covers what the old diagram did.

Here's a direct link in the README: https://github.com/leifdenby/neural-lam/tree/feat/readme-code-diagram?tab=readme-ov-file#pre-processing, and a screenshot from vscode:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,9 +104,9 @@ source_xy_coordinates["nwp_xy.npy"]:::npy_file
 surface_geopotential["surface_geopotential.npy"]:::npy_file
 grid_features["grid_features.pt"]:::pt_file
 border_mask["border_mask.npy"]:::npy_file
-nwp_mbr_files["nwp_*_mbr*.npy"]:::npy_file
-wtr_files["wtr_*.npy"]:::npy_file
-nwp_toa_files["nwp_toa*.npy"]:::npy_file
+nwp_mbr_files["nwp_*_mbr*.npy: [Nt,Ny,Nx,Nv]"]:::npy_file
+wtr_files["wtr_*.npy: [Ny,Nx]"]:::npy_file
+nwp_toa_files["nwp_toa*.npy: [Nt,Ny,Nx]"]:::npy_file
 
 border_mask -- read by --> training_step
 grid_features -- read by --> training_step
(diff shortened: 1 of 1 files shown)


leifdenby:
@joeloskarsson I updated the diagram because I realised that I had slightly misunderstood the shapes of what is in the `.npy`-files. I made the following notebook which plots each of the inputs you use, just for myself to learn what the shapes of what everything is: https://github.com/leifdenby/neural-lam/blob/examining-inputs/notebooks/examining_inputs.ipynb

README.md
```diff
@@ -83,9 +83,86 @@ It should thus be useful to make sure that your python environment is set up cor
...
+  create_parameter_weights_step -- writes --> parameter_weights
+  
+  surface_geopotential -- read by --> create_grid_features_step
+  source_xy_coordinates -- read by --> create_grid_features_step
+  border_mask -- read by --> create_grid_features_step
+  create_grid_features_step -- writes --> grid_features
+  
+  subgraph parameter_weights
+    direction LR
+    parameter_mean.pt\nparameter_std.pt\ndiff_mean.pt\ndiff_std.pt\nflux_stats.pt:::pt_file
```
joeloskarsson, line 143: The file `parameter_weights.npy` is missing here.

README.md
```diff
@@ -83,9 +83,86 @@ It should thus be useful to make sure that your python environment is set up cor
...
+parameter_weights -- read by --> training_step
+mesh_information -- read by --> training_step
+dynamic_variables -- read by --> training_step
+
+subgraph Dataset
+  subgraph static_variables["static variables [Nx,Ny]"]
+    source_xy_coordinates
+    surface_geopotential
+    border_mask
+    note("stored in `data/{dataset_name}/static/")
```
joeloskarsson, line 122: How tricky would it be to make these notes look more different than the files? It would be nice just to avoid confusion. But if it is not trivial to change the styling a bit it doesn't matter.

joeloskarsson (changes requested):
I think this is great, and much more maintainable than the diagram I had made before. Just fix the missing file and (optionally) see if you can re-style the notes, and I'll merge.

As we refactor I hope to abstract some of these things away, but having this diagram also to change when we are doing such refactoring is very valuable.

joeloskarsson:
Should we fix and merge this, or is this diagram obsolete as we are moving to other ways of handling data loading? I just don't see a need for this to sit here as an open PR.

leifdenby:
I think this is obsolete now :)

PR #10, opened 2024-02-22
Re-define RMSE metric to take sqrt after sample averaging

joeloskarsson:
The current definition of RMSE averages only spatially over the grid points, then takes the sqrt and averages these per-sample RMSEs for the final value. This choice of definition was made to be consistent with weatherbench(1 and 2). However, in the latest version of weatherbench 2 this has been changed, and the RMSE definition is to average over both spatial points and samples before taking the sqrt. Researching this a bit myself I agree that this is a better definition and the one I think we should use. It is additionally more suitable for future use in computations of Spread-Skill-Ratios.

This PR changes the implementation of the RMSE to match this new definition. Note that to achieve this we can not have RMSE as a metric along with others in `metrics..py`, but rather have to store MSE in validation and test steps to later average and take the square root.

### DIFF
--- a/neural_lam/models/ar_model.py
+++ b/neural_lam/models/ar_model.py
@@ -75,10 +75,10 @@ class ARModel(pl.LightningModule):
 
         self.step_length = args.step_length  # Number of hours per pred. step
         self.val_metrics = {
-            "rmse": [],
+            "mse": [],
         }
         self.test_metrics = {
-            "rmse": [],
+            "mse": [],
             "mae": [],
         }
         if self.output_std:
@@ -238,7 +238,9 @@ class ARModel(pl.LightningModule):
         """
         return self.all_gather(tensor_to_gather).flatten(0, 1)
 
-    def validation_step(self, batch):
+    # newer lightning versions requires batch_idx argument, even if unused
+    # pylint: disable-next=unused-argument
+    def validation_step(self, batch, batch_idx):
         """
         Run validation on single batch
         """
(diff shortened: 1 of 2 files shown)


joeloskarsson:
@sadamov can you review this change? Will assign you as reviewer, but I had to invite you to the repository first.

neural_lam/models/ar_model.py
```diff
@@ -508,10 +511,16 @@ def aggregate_and_plot_metrics(self, metrics_dict, prefix):
             )  # (N_eval, pred_steps, d_f)
 
             if self.trainer.is_global_zero:
+                metric_tensor_averaged = torch.mean(metric_tensor, dim=0)
+                # (pred_steps, d_f)
+
+                # Take square root after all averaging to change MSE to RMSE
+                if "mse" in metric_name:
```
sadamov, line 518: Here I wonder whether rmse should always **replace** mse. There might be a verification scenario where we want mse directly, no?

sadamov:
I agree, this is the preferred way to calculate (R)MSE for consistency with other applications and flexibility after validation/test step to calculate additional metrics based on MSE.
Note that the discussion about optimal spread-skill metrics is decades old, but that is a topic for another time (e.g. [here](https://journals.ametsoc.org/view/journals/wefo/19/3/1520-0434_2004_019_0552_aotsru_2_0_co_2.xml)).

The implementation is sound, tested with the `meps_example`. Thanks @joeloskarsson :raised_hands:

sadamov (approved):
I agree, this is the preferred way to calculate (R)MSE for consistency with other applications and flexibility after validation/test step to calculate additional metrics based on MSE.
Note that the discussion about optimal spread-skill metrics is decades old, but that is a topic for another time (e.g. [here](https://journals.ametsoc.org/view/journals/wefo/19/3/1520-0434_2004_019_0552_aotsru_2_0_co_2.xml)).

The implementation is sound, tested with the `meps_example`. Thanks @joeloskarsson :raised_hands:

neural_lam/models/ar_model.py
joeloskarsson: I agree that it is probably not great to always get rid of MSE, if we would be interested in that. My thought process here was that this was the simplest implementation and one can still get MSE from just squaring the saved RMSE values.

Do you think it would be better to log both MSE and RMSE instead of replacing the MSE logging? That should be a simple change.

sadamov: You are right, simply squaring the rmse is easy enough and avoids duplicating the loss tensor. I would go ahead and merge the PR as is.

joeloskarsson:
Thanks for looking at it @sadamov. I'll merge now.

PR #16, opened 2024-05-01
Add Rank-Zero Printing and Improve Wandb Initialization

sadamov:
Description:
This PR introduces `rank_zero_print` and `init_wandb` utility functions to enhance printing behavior in multi-process environments and streamline Weights and Biases (wandb) initialization for logging and experiment tracking.

Changes:
- Added `rank_zero_print` function:
  - Ensures printing only occurs from rank 0 process in multi-process setup.
  - Wraps `print` function for selective printing based on process rank.

- Introduced `init_wandb` function:
  - Initializes wandb based on provided arguments.
  - Handles new runs and resuming existing runs.
  - Generates meaningful run names with relevant information.
  - Creates `WandbLogger` for PyTorch Lightning integration.
  - Saves `constants.py` file to wandb for reference.

- Added necessary imports for time, PyTorch Lightning, wandb, and rank-zero utilities.

Benefits:
- Prevents duplicate or conflicting output in multi-process environments.
- Simplifies wandb setup and configuration for experiment tracking and logging.
- Improves organization and identification of experiments with informative run names.
- Enhances reproducibility by saving `constants.py` file to wandb.


neural_lam/utils.py
```diff
@@ -271,3 +275,49 @@ def init_wandb_metrics(wandb_logger):
...
         experiment.define_metric(f"val_loss_unroll{step}", summary="min")
+
+
+@rank_zero_only
+def rank_zero_print(*args, **kwargs):
+    """Print only from rank 0 process"""
+    print(*args, **kwargs)
+
+
+@rank_zero_only
```
joeloskarsson, line 286: This is super nice, I was not aware that this util exists! Probably other places in the code where this could be used as well, but that does not have to be in this PR.

neural_lam/utils.py
```diff
@@ -271,3 +275,49 @@ def init_wandb_metrics(wandb_logger):
...
+def init_wandb(args):
+    """Initialize wandb"""
+    if args.resume_run is None:
+        prefix = f"subset-{args.subset_ds}-" if args.subset_ds else ""
+        if args.eval:
+            prefix = prefix + f"eval-{args.eval}-"
+        run_name = (
+            f"{prefix}{args.model}-{args.processor_layers}x{args.hidden_dim}-"
+            f"{time.strftime('%m_%d_%H_%M_%S')}"
```
joeloskarsson, line 295: Can we keep an actual random number as part of run name? I have experienced that runs with different configs start at the same second on SLURM, resulting in naming conflicts and checkpoints being over-written. Or is there a better solution to this (having a random number in there is not super nice)?

neural_lam/utils.py
```diff
@@ -271,3 +275,49 @@ def init_wandb_metrics(wandb_logger):
...
+        wandb.init(
+            name=run_name,
+            project=constants.WANDB_PROJECT,
+            config=args,
+        )
+        logger = pl.loggers.WandbLogger(
+            project=constants.WANDB_PROJECT,
+            name=run_name,
+            config=args,
+            log_model=True,
```
joeloskarsson, line 306: Not sure about defaulting this to true. If large models are trained that is a lot of data going to wandb. Can we make it an option in some way?

neural_lam/utils.py
```diff
@@ -271,3 +275,49 @@ def init_wandb_metrics(wandb_logger):
...
+        )
+        wandb.save("neural_lam/constants.py")
```
joeloskarsson, line 308: This is good, can easily replace by saving the yaml config once that is in place.

neural_lam/utils.py
```diff
@@ -271,3 +275,49 @@ def init_wandb_metrics(wandb_logger):
...
+
+
+@rank_zero_only
+def rank_zero_print(*args, **kwargs):
+    """Print only from rank 0 process"""
+    print(*args, **kwargs)
+
+
+@rank_zero_only
+def init_wandb(args):
```
joeloskarsson, line 287: Could the metric initialization from `utils.init_wandb_metrics` also be included in this function?

joeloskarsson (changes requested):
Overall a nice improvement. I had some small comments related to my use cases.

neural_lam/utils.py
sadamov: I set this to False (default) again, the checkpointing should be considered in a seperate PR #17

neural_lam/utils.py
sadamov: ah yes, I had this issue as well. I think adding a randint is totally appropriate here.

sadamov:
Okay I requested another review, and from my side this PR should be ready for merge.

neural_lam/utils.py
```diff
@@ -260,11 +265,50 @@ def fractional_plot_bundle(fraction):
...
+        run_name = (
+            f"{prefix}{args.model}-{args.processor_layers}x{args.hidden_dim}-"
+            f"{time.strftime('%m_%d_%H_%M_%S')}-{random_int}"
+        )
+        wandb.init(
+            name=run_name,
+            project=args.wandb_project,
+            config=args,
+        )
+        logger = pl.loggers.WandbLogger(
```
joeloskarsson, line 291: Now we get a warning every time the WandbLogger is created, as wandb is already initialized. Is it really necessary to first do the `wandb.init` call? Would the same thing not be achieved by just creating the `WandbLogger`? (and not have that warning) `WandbLogger` should be able to take all the same arguments as `wandb.init`.

neural_lam/utils.py
```diff
@@ -260,11 +265,50 @@ def fractional_plot_bundle(fraction):
...
+        )
+        wandb.save("neural_lam/data_config.yaml")
+    else:
+        wandb.init(
+            project=args.wandb_project,
+            config=args,
+            id=args.resume_run,
+            resume="must",
+        )
+        logger = pl.loggers.WandbLogger(
```
joeloskarsson, line 304: Same thing here, do we need both `wandb.init` and `WandbLogger`?

joeloskarsson (changes requested):
I tested this on multi-gpu and it works great. I do however get a warning when the logger is created, that I think we maybe can avoid.

sadamov:
@joeloskarsson the issue was that wandb is only initialized once the trainer.fit() started. So a nice way to solve this issue is: removing the `wandb.init()` as you suggested and then saving the configs in the ar_model:

```
    def on_run_end(self):
            wandb.save("neural_lam/data_config.yaml")
```

sadamov:
@leifdenby can I add this PR to the roadmap for v.0.2.0? and then add it as a feature to the changelog as well?

neural_lam/models/ar_model.py
```diff
@@ -597,3 +597,7 @@ def on_load_checkpoint(self, checkpoint):
         if not self.restore_opt:
             opt = self.configure_optimizers()
             checkpoint["optimizer_states"] = [opt.state_dict()]
+
+    def on_run_end(self):
```
joeloskarsson, line 601: When is this hook called? I can't find it in the Lightning docs. I am thinking that we should save the config as soon as possible, if the training run would crash at some point. But maybe that is what this does? Otherwise I guess `on_train_start` and `on_test_start` would be good hooks for this.

Also minor thing: Add docstring

joeloskarsson (changes requested):
Right, I know remember that the wandb initialization timing is a bit annoying. But I think this is a great solution! I just don't understand the `on_run_end` hook, so want to double check that before we merge.

Also: You can add this to the changelog already now.

neural_lam/models/ar_model.py
sadamov: Good idea, I switched the hooks to the ones you suggest. CHANGELOG is ready as well.

leifdenby:
LGTM!

sadamov:
Apologies I have been quite slow reacting to this PR, I merged latest `main` into PR-branch. Okay with @leifdenby if I merge this now into v0.2.0?

sadamov:
implemented in https://github.com/mllam/neural-lam/pull/103

Issue #7, opened 2024-01-27
Keeping a changelog and versioning


leifdenby:
I would like to propose that we start keeping a changelog. This would mean that from now on changes are only added through pull-requests, and every pull-request should add an entry to `CHANGELOG.md` (typically this addition will be done as a final commit once a pull-request has passed tests and review).

My notes below are based on https://keepachangelog.com/en/1.0.0/, please read it. It is a very succinct. 

**Why keep a changelog?**

As the code keeps evolving it will be important to be able to communicate with each other how the codebase is changing. Having a single point of reference, where changes are grouped by **feature additions**, **bugfixes**, **breaking changes** and **maintenance** will make it a lot easier (and a lot more fun!) to work with the codebase. 

For reference here is xarray's changelog: https://github.com/pydata/xarray/blob/main/doc/whats-new.rst, here is one for a different project I've worked on: https://github.com/cloudsci/cloudmetrics/blob/master/CHANGELOG.md

Connected to this I would also like to introduce versioning, using [semantic versioning](https://semver.org/). I would to tag the commit https://github.com/joeloskarsson/neural-lam/commit/2378ed7eddf8da5bfec6f57c41cadf310d191dee @joeloskarsson as `v0.1.0`, and create a changelog relative to there. My reasoning for this is that I think that commit was the most recent commit when you shared this repository publicly. We can the start the changelog by including the commit that you made for recent loss-function additions (new feature) and the maintenance @sadamov is doing in #6 by setting up linting.

In that case the CHANGELOG would look something like:

```
# Changelog

## [Unreleased](https://github.com/joeloskarsson/neural-lam/HEAD)

[Full Changelog](https://github.com/joeloskarsson/neural-lam/compare/v0.1.0...HEAD)

*new features*

- additional loss-functions 
  [2378ed7](https://github.com/joeloskarsson/neural-lam/commit/2378ed7eddf8da5bfec6f57c41cadf310d191dee)](https://github.com/joeloskarsson/neural-lam/commit/c14b6b4323e6b56f1f18632b6ca8b0d65c3ce36a) By Joel Oskarsson
  (@joeloskarsson )

*maintenance*

- set up linting with pyflake8, black with pre-commit with github action to run in cicd [\#1](https://github.com/joeloskarsson/neural-lam/pull/6/) by Simon Adamov (@sadamov)

## [v0.1.0](https://github.com/joeloskarsson/neural-lam/HEAD)

[...]

PR #28, opened 2024-05-13
Add changelog

leifdenby:
This PR adds a changelog as I proposed in #7. I have tried to go through the commits (https://github.com/joeloskarsson/neural-lam/compare/v0.1.0...HEAD) since v0.1.0

@joeloskarsson and @sadamov would you able to check this with the commits/PRs that you are tagged on to check that you are happy with what I've written? Otherwise maybe you could suggest some changes?

You can view the rendered changelog here: https://github.com/leifdenby/neural-lam/blob/maint/changelog/CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,14 +32,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   [c14b6b4](https://github.com/joeloskarsson/neural-lam/commit/c14b6b4323e6b56f1f18632b6ca8b0d65c3ce36a)
   @joeloskarsson
 
-- compute `rmse` after spatial averaging
+- change RMSE definition to compute sqrt after all averaging
   [\#10](https://github.com/joeloskarsson/neural-lam/pull/10)
   @joeloskarsson
 
 ### Removed
 
-- `WeatherDataset(torch.Dataset)` no longer returns "static" component of
-  training item (only `prev_state`, `target_state` and `forcing`), static is
+- `WeatherDataset(torch.Dataset)` no longer returns "batch-static" component of
+  training item (only `prev_state`, `target_state` and `forcing`), the batch static features are
   instead included in forcing
   [\#13](https://github.com/joeloskarsson/neural-lam/pull/13)
   @joeloskarsson
@@ -47,5 +47,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [v0.1.0](https://github.com/joeloskarsson/neural-lam/releases/tag/v0.1.0)
 
-First tagged release of `neural-lam`, matching Oscarsson et al 2023 publication
+First tagged release of `neural-lam`, matching Oskarsson et al 2023 publication
 (https://arxiv.org/abs/2309.17370)


CHANGELOG.md
```diff
@@ -0,0 +1,51 @@
...
+- `WeatherDataset(torch.Dataset)` no longer returns "static" component of
+  training item (only `prev_state`, `target_state` and `forcing`), static is
+  instead included in forcing
+  [\#13](https://github.com/joeloskarsson/neural-lam/pull/13)
+  @joeloskarsson
+
+
+## [v0.1.0](https://github.com/joeloskarsson/neural-lam/releases/tag/v0.1.0)
+
+First tagged release of `neural-lam`, matching Oscarsson et al 2023 publication
```
joeloskarsson, line 50: \* Oskarsson :)

CHANGELOG.md
```diff
@@ -0,0 +1,51 @@
...
+
+- moved batch-static features ("water cover") into forcing component return by `WeatherDataset`
+  [\#13](https://github.com/joeloskarsson/neural-lam/pull/13)
+  @joeloskarsson
+
+- change validation metric from `mae` to `rmse`
+  [c14b6b4](https://github.com/joeloskarsson/neural-lam/commit/c14b6b4323e6b56f1f18632b6ca8b0d65c3ce36a)
+  @joeloskarsson
+
+- compute `rmse` after spatial averaging
```
joeloskarsson, line 35: The change is to take sqrt after averaging all the forecasts. It was always after spatial averaging. See e.g. difference between v1 and v2 of WB2 https://arxiv.org/abs/2308.15560.

CHANGELOG.md
```diff
@@ -0,0 +1,51 @@
...
+  [\#10](https://github.com/joeloskarsson/neural-lam/pull/10)
+  @joeloskarsson
+
+### Removed
+
+- `WeatherDataset(torch.Dataset)` no longer returns "static" component of
```
joeloskarsson, line 41: Maybe better to refer to this as batch-static here as well, to avoid confusing it with actual static features, which were never returned from `WeatherDataset`.

joeloskarsson (changes requested):
A few tiny things, but otherwise this is great. Let's try to merge this first, so we can add the changes from other PRs to the changelog.

CHANGELOG.md
leifdenby: Oh no, sorry! I keep doing this... so sorry

CHANGELOG.md
leifdenby: I don't quite understand what you mean by "averaging all the forecasts". Is there a figure in the WeatherBench2 article you're referring? Or is it something discussed in v2 that was missing in v1 of that article?

CHANGELOG.md
leifdenby: yes, good point, I'll change that

CHANGELOG.md
joeloskarsson: Sorry, I realize that was not very clear. I was referencing the RMSE definition in the WB2 paper (eq.2 in both versions). It is somewhat different in v1 and v2. A good description for this change would be "change RMSE definition to compute sqrt after all averaging".

leifdenby: thank you :)

leifdenby:
thanks for your feedback @joeloskarsson! I've adapted the changelog with your suggestions.

@sadamov are you happy with the changelog as i stands?

joeloskarsson (approved):
All looks good now

sadamov (approved):
Thanks for adding the CHANGELOG, looks good! Ready to merge from my side.
Do we agree to squash commits per default?

joeloskarsson:
Go ahead and merge. Yes, we should definitely squash as a default. If there is a setting to to strictly enforce this that would be good.

sadamov:
I changed the settings to this for now: 
image

PR #31, opened 2024-05-14
Replace constants.py with data_config.yaml

sadamov:
**Summary**
This PR replaces the `constants.py` file with a `data_config.yaml` file. Dataset related settings can be defined by the user in the new yaml file. Training specific settings were added as additional flags to the `train_model.py` routine. All respective calls to the old files were replaced.

**Rationale**

- Using a Yaml file for data config gives much more flexibility for various datasets used in the community. It also facilitates the future use of forcing and boundary datasets. In a follow-up PR the dataset paths will be defined in the yaml file, removing the dependency on a pre-structured `/data` folder. 
- It is best practice to define user input in a yaml file, the usage of python scripts for that purpose is not common.
- The old `constants.py` actually combined both constants and variables, many "constants" should rather be flags to `train_models.py`
- The introduction of a new ConfigClass in `utils.py` allows for very specific queries of the yaml and calculations based thereon. This branch shows future possibilities of such a class https://github.com/joeloskarsson/neural-lam/tree/feature_dataset_yaml

**Testing**
Both training and evaluation of the model were succesfully tested with the `meps_example` dataset.

**Note**
@leifdenby Could you invite Thomas R. to this repo, in case he wanted to give his input on the yaml file? This PR should mostly serve as a basis for discussion. Maybe we should add more information to the yaml file as you outline in https://github.com/mllam/mllam-data-prep. I think we should always keep in mind how the repository will look like with realistic boundary conditions  and zarr-archives as data-input.

This PR solves parts of https://github.com/joeloskarsson/neural-lam/issues/23


neural_lam/utils.py
```diff
@@ -263,11 +262,65 @@ def fractional_plot_bundle(fraction):
...
+        return value
+
+    def __contains__(self, key):
+        return key in self.values
+
+    def num_data_vars(self):
+        """Return the number of data variables for a given key."""
+        return len(self.dataset.vars)
+
+    def projection(self):
```
joeloskarsson, line 320: Maybe @property would be nice for this?

joeloskarsson: Or maybe it would actually make more sense to pre-compute this in the constructor and store it then. So we do not have to re-instantiate it every time a plot should be created.

leifdenby:
From a quick glance this looks simply amazing @sadamov! Thanks for doing this work. I will give a thorough review later today/tomorrow. Just tagging @SimonKamuk to have a read and give your thoughts too. I've added @ThomasRieutord to the organisation too. I'll also send Thomas an email so that he definitely sees the PR.

neural_lam/utils.py
leifdenby: Yes, I think a decorator would be a nice idea since this is an attribute of the dataset rather than something that is derived based on a set of arguments. Rather than instantiate in `__init__` I would cache with [functools.cache](https://docs.python.org/3/library/functools.html#functools.cache). Also, I think it would be nice to rename as `coords_projection` or something like that, to make it explicitly that this is about the projection of the dataset coordinates.

neural_lam/utils.py
```diff
@@ -263,11 +262,65 @@ def fractional_plot_bundle(fraction):
...
+
+class ConfigLoader:
+    """
+    Class for loading configuration files.
+
+    This class loads a YAML configuration file and provides a way to access
+    its values as attributes.
+    """
+
+    def __init__(self, config_path, values=None):
```
leifdenby, line 283: you could create a classmethod called something like `Config.from_file(filepath)`, e.g.

```python
class Config
    def __init__(self, values):
        self.values = values

class YamlConfig(Config):
    @classmethod
    def from_file(cls, filepath):
        with open(self.config_path, encoding="utf-8", mode="r") as file:
            return cls(values=yaml.safe_load(file))
```

This makes it possible to create a `Config` directly from a `dict` for example and allows us to easily implement other storage formats. Or alternatively:

```python
class Config
    def __init__(self, values):
        self.values = values

    @classmethod
    def from_file(cls, filepath):
        if filepath.endswith(".yaml"):
            with open(self.config_path, encoding="utf-8", mode="r") as file:
                return cls(values=yaml.safe_load(file))
        else:
             raise NotImplementedError(Path(filepath).suffix))
```

I think I prefer the latter actually since it means fewer classes, and it probably isn't important to know exactly what format the config was loaded from.

neural_lam/weather_dataset.py
```diff
@@ -218,9 +218,10 @@ def __getitem__(self, idx):
         # can roll over to next year, ok because periodicity
 
         # Encode as sin/cos
+        seconds_in_year = 365 * 24 * 3600
```
leifdenby, line 221: maybe we should add a `# NB: this assumes that all years contain 365 days, may want to change this in the future` or something like that?

neural_lam/models/ar_model.py
```diff
@@ -246,7 +246,7 @@ def validation_step(self, batch, batch_idx):
         # Log loss per time step forward and mean
         val_log_dict = {
             f"val_loss_unroll{step}": time_step_loss[step - 1]
-            for step in constants.VAL_STEP_LOG_ERRORS
+            for step in self.args.val_steps_log
```
leifdenby, line 249: This is another one I would consider renaming. Maybe `val_steps_to_log`? I think `val_steps_log` is better than `VAL_STEP_LOG_ERRORS` because the latter could either be interpreted as a verb (*log these errors") or the "log errors of val step"

neural_lam/models/ar_model.py
```diff
@@ -328,7 +328,9 @@ def test_step(self, batch, batch_idx):
         spatial_loss = self.loss(
             prediction, target, pred_std, average_grid=False
         )  # (B, pred_steps, num_grid_nodes)
-        log_spatial_losses = spatial_loss[:, constants.VAL_STEP_LOG_ERRORS - 1]
+        log_spatial_losses = spatial_loss[
+            :, [step - 1 for step in self.args.val_steps_log]
```
leifdenby, line 332: is the list comprehension here because `constants.VAL_STEP_LOG_ERRORS` was previously a numpy array but now is a list?

leifdenby (commented):
I really like this work @sadamov! You've really caught all the bits here (which I'm impressed you've done considering we don't have any tests right now!)

I have just made a few comments/suggestions. Let me know what you think :)

README.md
```diff
@@ -45,7 +45,7 @@ Still, some restrictions are inevitable:
 ## A note on the limited area setting
 Currently we are using these models on a limited area covering the Nordic region, the so called MEPS area (see [paper](https://arxiv.org/abs/2309.17370)).
 There are still some parts of the code that is quite specific for the MEPS area use case.
-This is in particular true for the mesh graph creation (`create_mesh.py`) and some of the constants used (`neural_lam/constants.py`).
+This is in particular true for the mesh graph creation (`create_mesh.py`) and some of the constants used (`neural_lam/data_config.yaml`).
```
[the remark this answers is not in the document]
joeloskarsson: I think it would be best if it is completely up to the user how and where the config is stored. Then you just give any path to the yaml-file as input to the main script.

neural_lam/weather_dataset.py
joeloskarsson: This whole computation has to be handled better, but I agree that we can just make a note of this for now

README.md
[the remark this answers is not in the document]
leifdenby: > I think it would be best if it is completely up to the user how and where the config is stored. Then you just give any path to the yaml-file as input to the main script.

yes, that was what I was thinking too :) I was just being a bit more long-winded about it. The README here suggests to me that it _should_ reside inside `neural_lam/`. Maybe we should change this to `data_config.yaml` instead rather than `neural_lam/data_config.yaml` then?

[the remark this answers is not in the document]
sadamov: Agreed, I changed the Readme accordingly.

neural_lam/data_config.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - t_isobaricInhPa_500_instant
+    - t_isobaricInhPa_850_instant
+    - u_hybrid_65_instant
+    - u_isobaricInhPa_850_instant
+    - v_hybrid_65_instant
+    - v_isobaricInhPa_850_instant
+    - wvint_entireAtmosphere_0_instant
+    - z_isobaricInhPa_1000_instant
+    - z_isobaricInhPa_500_instant
+  forcing_dim: 16
```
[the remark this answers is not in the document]
sadamov: In a future PR the forcings will be provided by a path to a zarr archive containing forcing features. Since in the current MEPS implementation the calculation of forcings is heavily integrated into the Dataset/Dataloader, I suggest to change the name to `num_forcing_dim` for now and implement the fundamental changes "naming forcing variables" once the zarr-based approach was merged into main. See https://github.com/mllam/neural-lam/tree/feature_dataset_yaml

neural_lam/data_config.yaml
```diff
@@ -0,0 +1,64 @@
+dataset:
+  name: meps_example
+  vars:
```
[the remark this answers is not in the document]
sadamov: good idea

neural_lam/weather_dataset.py
sadamov: Since this is only required to calculate forcings, I suggest to handle it directly in a seperate script that provides forcings #34. If you are okay with it I would leave it as is for now.

neural_lam/models/ar_model.py
sadamov: yes exactly, alternatively we could convert the list from args.parse into numpy array beforehand.

neural_lam/utils.py
sadamov: great idea

sadamov:
I implemented most requested changes in the latest commit and requested one more review. From my side we are clear to merge. The latest changes were again tested for model training and evaluation.

neural_lam/data_config.yaml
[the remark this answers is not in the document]
leifdenby: sounds good to me! Happy to have the only change here be changed of name to `num_forcing_dim`

neural_lam/weather_dataset.py
leifdenby: yes, maybe you could just add a comment so that we have a note for later?

neural_lam/models/ar_model.py
leifdenby: I see, not this is good :) just wanted to understand

neural_lam/config.py
leifdenby, line 1: I'm not saying we should do this now, but I learnt more about the Meteo-France work on `neural-lam` this morning and they make quite heavy use of [python dataclasses]() for configuration storage and schema. This could be something to consider when we want to make the config content more explicit.

leifdenby (approved):
Looks great! Thanks again @sadamov !

leifdenby:
Remember to update the [changelog](https://github.com/mllam/neural-lam/blob/main/CHANGELOG.md) before you merge @sadamov! I think I would call this a new (very useful!) feature. A few things also change here (where constants are stored and thus how they are accessed in the code)

leifdenby:
Hurraay! :partying_face:

train_model.py
```diff
@@ -183,8 +180,36 @@ def main():
...
+    parser.add_argument(
+        "--metrics_watch",
+        type=list,
+        default=[],
+        help="List of metrics to watch, including any prefix (e.g. val_rmse)",
+    )
+    parser.add_argument(
+        "--var_leads_metrics_watch",
+        type=dict,
+        default={},
```
joeloskarsson, line 206: @sadamov Can you pass a dict as input on the command line? I could not figure out a way to use this option

joeloskarsson:
Wonderful job with this @sadamov!

train_model.py
sadamov: Ah no you can't, I'll make a short PR to fix three bugs that I introduced in this PR. One of them the dictionary here.

PR #38, opened 2024-05-22
Feature: add tests for meps dataset

### LINKED ISSUE #30: Add test for MEPS data example
Aim: Create a test setup where we ensure that npy-file based datasets can be read into neural-lam and the training items contain tensors of the right shape.

TODO

- [ ] reduce number of variables, size of domain etc in Joel's MEPS data example so that the zip file is less than 500MB. Calling it `meps_example_reduced`
- [ ] create test-data zip file and upload to EWC (credentials from @leifdenby)
- [ ] implement test using pytorch to download and unpack testdata using [pooch](https://pypi.org/project/pooch/)
- [ ] Implement testing of:
   - [ ] initiation of `neural_lam.weather_dataset.WeatherDataset` from downloaded data
   - [ ] check shapes of returned parts of training item
   - [ ] create new graph in tests for reduced dataset
   - [ ] feed single batch through model and check shape of output
- [ ] add github action to run tests during ci/cd

Links:
- uploading to EWC: https://github.com/mllam/mllam-docs/blob/main/hosting_source_datasets_on_ewc.md
- example of using `meps_example` for testing from Leif's branch: https://github.com/leifdenby/neural-lam/blob/mllam-dataloader/tests/test_mllam_dataset.py#L78
- leif's branch with dependencies handled with pdm: https://github.com/leifdenby/neural-lam/tree/maint/refactor-as-package

SimonKamuk:
Implemeted tests for loading a reduced size meps example dataset, creating graphs, and training model.

Work TODO:

- [x] reduce number of variables, size of domain etc in Joel's MEPS data example so that the zip file is less than 500MB. Calling it `meps_example_reduced`
- [x] create test-data zip file and upload to EWC (credentials from @leifdenby)
- [x] implement test using pytorch to download and unpack testdata using [pooch](https://pypi.org/project/pooch/)
- Implement testing of:
   - [x] initiation of `neural_lam.weather_dataset.WeatherDataset` from downloaded data
   - [x] check shapes of returned parts of training item
   - [x] create new graph in tests for reduced dataset
   - [x] feed single batch through model and check shape of output
- [x] add github action to run tests during ci/cd

closes #30


tests/test_mllam_dataset.py
```diff
@@ -0,0 +1,131 @@
...
+
+    var_names = config.values["dataset"]["var_names"]
+    var_units = config.values["dataset"]["var_units"]
+    var_longnames = config.values["dataset"]["var_longnames"]
+
+    assert len(var_names) == len(var_longnames)
+    assert len(var_names) == len(var_units)
+
+    # TODO: can these two variables be loaded from elsewhere?
+    n_grid_static_features = 4
```
leifdenby, line 51: I think you can assume here the number of "grid static features" that is present in the `grid_static.pt` file https://github.com/mllam/neural-lam/blob/5b71be3c68d815e0e376ee651c14f09d801f86de/neural_lam/utils.py#L56

The `WeatherDataset` doesn't currently expose number of grid-features, static or dynamic. Maybe that is something we should consider in future @sadamov?

leifdenby: or maybe these config file should at least explicitly state which grid static features there are?

neural_lam/vis.py
```diff
@@ -87,7 +87,7 @@ def plot_prediction(
         1,
         2,
         figsize=(13, 7),
-        subplot_kw={"projection": data_config.coords_projection()},
```
leifdenby, line 90: I think this will be fixed by @sadamov in #40, maybe we should wait for that to be merged? What do you think @sadamov?

neural_lam/utils.py
```diff
@@ -250,7 +251,8 @@ def fractional_plot_bundle(fraction):
     Get the tueplots bundle, but with figure width as a fraction of
     the page width.
     """
-    bundle = bundles.neurips2023(usetex=True, family="serif")
+    usetex = True if shutil.which("latex") else False
```
leifdenby, line 254: this seems resonable, are you ok with adding this @joeloskarsson?

requirements.txt
```diff
@@ -10,6 +10,18 @@ Cartopy>=0.22.0
...
 plotly>=5.15.0
+torch-geometric>=2.5.2
+loguru>=0.7.2
+xarray>=2024.3.0
+zarr>=2.17.2
+dask>=2024.4.2
 
 # for dev
 pre-commit>=2.15.0
+codespell>=2.0.0
```
leifdenby, line 21: these deps for linting are already defined and pinned in `.pre-commit-config.yaml`, it shouldn't be necessary to include them here too. Did you have problems with `pre-commit`?

tests/test_mllam_dataset.py
leifdenby, line 1: simply beautiful!

leifdenby (changes requested):
Looks great to me! It will be really nice to have this automated testing on a real dataset. What do you think to creating a file called `DEVELOPING.md` in the repo root where you detail the contents of the test data a bit more?

@sadamov could you give this a read through too please?

requirements.txt
SimonKamuk: I did, but it must have been something else, because it works now without these dependencies.

SimonKamuk:
> Looks great to me! It will be really nice to have this automated testing on a real dataset. What do you think to creating a file called `DEVELOPING.md` in the repo root where you detail the contents of the test data a bit more?
> 
> @sadamov could you give this a read through too please?

I added this as a notebook instead, since I could then include the code directly. Let me know if you would still prefer just a md file :)

tests/test_mllam_dataset.py
sadamov: We can define paths to zarr archives from static, forcing, boundary, state directly in the future (see  https://github.com/mllam/neural-lam/blob/feature_dataset_yaml/neural_lam/data_config.yaml)

`num_data_vars` is already implemented in the Config class: https://github.com/mllam/neural-lam/blob/5b71be3c68d815e0e376ee651c14f09d801f86de/neural_lam/config.py#L51

neural_lam/vis.py
sadamov: yes lets fix the bugs I introduced, I also had to manually fix this in multiple branches which is annoying. @leifdenby could you give a final review on #40 and then @SimonKamuk merge this PR-branch with main later.

neural_lam/utils.py
sadamov: Colleagues had issues with usetex=True on some machines, I appreciate this addition.

sadamov:
> > Looks great to me! It will be really nice to have this automated testing on a real dataset. What do you think to creating a file called `DEVELOPING.md` in the repo root where you detail the contents of the test data a bit more?
> > @sadamov could you give this a read through too please?
> 
> I added this as a notebook instead, since I could then include the code directly. Let me know if you would still prefer just a md file :)

Could we move this and all future notebooks to a `neural-lam/notebooks` folder? Then I would also give this notebook a more desciptive name like `create_reduced_meps_dataset.ipynb`

DEVELOPING.ipynb
sadamov, line 1: I would reference the `README.md` here on how to retrivee the original meps_example.

tests/test_mllam_dataset.py
```diff
@@ -0,0 +1,136 @@
+# Standard library
+import os
+
+# Third-party
+import pooch
```
sadamov, line 5: nice, I didn't know this. Very useful!

tests/test_mllam_dataset.py
```diff
@@ -0,0 +1,136 @@
...
+    var_names = config.values["dataset"]["var_names"]
+    var_units = config.values["dataset"]["var_units"]
+    var_longnames = config.values["dataset"]["var_longnames"]
+
+    assert len(var_names) == len(var_longnames)
+    assert len(var_names) == len(var_units)
+
+    # TODO: can these two variables be loaded from elsewhere?
+    n_grid_static_features = 4
+    n_input_steps = 2
```
sadamov, line 57: My understanding is that this is hardcoded in the repo (prev_prev + prev -> target), meaning the previous two time steps are used to predict the current. So it's fine to hardcode it here as well.

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      - name: Test with pytest
+        run: |
+          pytest  -v -s
```
sadamov, line 32: Should we cache the downloaded dataset to avoid re-downloads on every push? Or is that not necessary?
```suggestion
          pytest  -v -s
      - name: Cache data
        uses: actions/cache@v2
        with:
        path: data
        key: ${{ runner.os }}-data-${{ hashFiles('requirements.txt') }}
        restore-keys: |
          ${{ runner.os }}-data-
```

sadamov (changes requested):
Very nice @SimonKamuk, we definitely needed tests! Well done :rocket: 
I tested this with a push to mllam-repo and all tests run smoothly.

- I think you should add a sentence or two about the new testing to https://github.com/mllam/neural-lam/blob/main/README.md#development-and-contributing
- These tests take 150s (with download) and 100s(without) on my machine. I think that's okay since they mostly run in a seperate GitHub runner on push. In the future we could also use [Mock objects](https://docs.python.org/3/library/unittest.mock.html) for long-running tests  (e.g. `@mock.patch('train_model.main')`)

Otherwise, I only requested some smaller changes. Thanks again Simon!

neural_lam/vis.py
joeloskarsson: Should be fixed now with the merge of #17

neural_lam/utils.py
joeloskarsson: I also struggle with this when sometimes working on machines without a tex installation. I think this is a good solution. If you don't want to use tex you can use something like ² in units and I think that will look ok. Let's leave further improvements to a more robust rework of the visualization code.

tests/test_mllam_dataset.py
joeloskarsson: Using 2 input steps is very much assumed throughout all the code. As this is the standard assumption for ML weather models right now I reckon we will not have to think about generalizing this in the near future.

tests/test_mllam_dataset.py
leifdenby: ok, in that case I think this comment in the code can stay, or you could maybe change it to say `in future the number of grid static features will be provided by the Dataset class itself` or something like that. Sound ok?

neural_lam/vis.py
leifdenby: > Should be fixed now with the merge of #17

Ok, sounds good. Maybe merge upstream into your branch @SimonKamuk and push this branch to github again?

neural_lam/utils.py
leifdenby: Ok, maybe in the case add a comment @SimonKamuk that with this change plots might render incorrectly (but at least not cause an exception to be raised?)

DEVELOPING.ipynb
leifdenby: Good idea! Otherwise this notebook looks great @SimonKamuk!

I would love to have JupyterBook based documentation for neural-lam (like for [weather-model-graphs](https://github.com/mllam/weather-model-graphs)) in future, maybe for `v0.3.0` if I (or someone else is keen) to add this :smile:

tests/test_mllam_dataset.py
leifdenby: yeah it is a really nice package. Can download directly from zenodo with a DOI too

leifdenby (approved):
I only have two minor comments about comments you could add to the code. Otherwise this is good to go.

Remember to update the changelog to say that you have added testing of dataset loading, graph creation and training based on reduced MEPS dataset stored on AWS S3! This is a big step!!

.github/workflows/run_tests.yml
leifdenby: good idea, although the caching should maybe be done on some identifier unique to the dataset. If you're with the testdata you have on AWS now @SimonKamuk you could maybe change the cache key to `meps-reduced-example-data-v0.1.0` or something like that, rather than based the key on the contents of `requirements.txt`

tests/test_mllam_dataset.py
leifdenby: Yes, good point. Maybe you could add a comment to say as much here as well @SimonKamuk?

requirements.txt
```diff
@@ -10,6 +10,13 @@ Cartopy>=0.22.0
 pyproj>=3.4.1
 tueplots>=0.0.8
 plotly>=5.15.0
+torch-geometric>=2.5.2
```
joeloskarsson, line 13: Will this play well with all the `pytorch-geometric` depdendencies (see point 5 in install instructions in the readme). If so I think it makes sense to include the `pytorch-geometric` dependency here, but we should also adjust the install instructions in the README.

requirements.txt
```diff
@@ -10,6 +10,13 @@ Cartopy>=0.22.0
...
+loguru>=0.7.2
+xarray>=2024.3.0
+zarr>=2.17.2
+dask>=2024.4.2
```
joeloskarsson, line 17: Why are these introduced with this change?

neural_lam/utils.py
joeloskarsson: Alternatively we just change the units to use ² , together with this change? I think that should look ok both with and without tex (but you might want to double-check that).

DEVELOPING.ipynb
joeloskarsson: While I agree that it is nice to have this documented somewhere, I think the current name and placement in the repo is a bit misleading and could confuse people finding the repo. A file called `DEVELOPING` in the root of the repo I would assume is about information for people who want to do development of the code. But this notebook is documentation about a very specific part of the testing setup. I would much rather have this as something like `docs/notebooks/meps_reduced_creation.ipynb` or similar.

joeloskarsson (changes requested):
This is very valuable to have! I had some questions about the added dependencies in `requirements.py` and also some follow-ups to earlier discussions that might be good to take a look at.

joeloskarsson:
> I only have two minor comments about comments you could add to the code. Otherwise this is good to go.
> 
> Remember to update the changelog to say that you have added testing of dataset loading, graph creation and training based on reduced MEPS dataset stored on AWS S3! This is a big step!!
> 
> Also I think it would be nice to add to the top of the README cicd badges for the testing if @sadamov and @joeloskarsson you are ok with that?

I have no opinion on cicd badges, so please go ahead if you'd like

DEVELOPING.ipynb
leifdenby: > While I agree that it is nice to have this documented somewhere, I think the current name and placement in the repo is a bit misleading and could confuse people finding the repo. A file called `DEVELOPING` in the root of the repo I would assume is about information for people who want to do development of the code. But this notebook is documentation about a very specific part of the testing setup. I would much rather have this as something like `docs/notebooks/meps_reduced_creation.ipynb` or similar.

Yes, I agree. Ok, let's add a `docs/` folder and we can discuss later what we like have in there over time. But having a `docs/` for all documentation related stuff sounds good to me. @sadamov and @SimonKamuk sound good?

requirements.txt
leifdenby: I'm not sure to be honest. In #37 am planning to implement install and testing on both cpu-only and gpu setups. So we can maybe postpone making a decision on where `pytorch-geometric` sits until then?

If so I would suggest you remove this from `requirements.txt` @SimonKamuk and instead put the installation into the github action following the steps that @joeloskarsson has put in the README. Sound ok?

joeloskarsson: Note that the `torch_geometric` package itself does not need to be installed differently depending on if you are on CPU or GPU (see https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html). That is only the case for its optional dependencies these days. So I do not see a problem with just having it listed in `requirements.txt`. 

We should probably go over those optional dependencies as well, since I am not sure if they all are needed. Maybe something to do for #37 as well?

leifdenby: Just to double check you're agreeing @joeloskarsson with my suggestion that the `torch-geometric` should be removed from `requirements.txt`, right? Just so @SimonKamuk knows what to do :)

leifdenby (changes requested):
I think we just need to remove `pytorch-geometric` from the requirements.txt and then this is good to go!

requirements.txt
joeloskarsson: Yes, I think that is the easiest for now, and handle any restructuring of these dependencies in #37

joeloskarsson (approved):
All looking good now I think. I have only read the code, but trust that others have tested that everything runs. Great work on this!

leifdenby (approved):
Ready!

SimonKamuk:
@sadamov Do you have any further requests or are we ready to merge? :)

sadamov:
Ready go merge! :rocket:

sadamov (approved):

joeloskarsson:
@SimonKamuk at the moment I am seeing "Unit Tests: Failing" on the badge, even thought this is not the case for the main branch. I was looking at https://github.com/mllam/neural-lam/actions/workflows/run_tests.yml and it seems to be on some other branch they have failed. I really don't know much about how these badges work, so not sure how it is supposed to function :sweat_smile: Do you know if there is a way to tweak it so that it only applies to the main branch? Or would that require large changes to the way the github workflows are set up?

SimonKamuk:
> @SimonKamuk at the moment I am seeing "Unit Tests: Failing" on the badge, even thought this is not the case for the main branch. I was looking at https://github.com/mllam/neural-lam/actions/workflows/run_tests.yml and it seems to be on some other branch they have failed. I really don't know much about how these badges work, so not sure how it is supposed to function 😅 Do you know if there is a way to tweak it so that it only applies to the main branch? Or would that require large changes to the way the github workflows are set up?

Yeah, I noticed that as well. I could specify that the badge should only refer to the main branch, but the issue is that the workflows are not running on pushes to main, so there are no runs to refer to. In the recent commit, I made these changes, but I am not sure what the implication of no longer ignoring pushes to main (why were they ignored in the first place @leifdenby?)

joeloskarsson:
Ok, that's great! I guess running on push to main is generally unnecessary since we run everything before merging. However I don't think it matters if we run it an extra time just for the badge.

Issue #23, opened 2024-05-03
Replace constants.py with data + region specification from yaml-file


joeloskarsson:
This supersedes #2 and #3.

## Motivation
It is currently very hard to work with neural lam on different regions due to everything related to data and the forecast region being specified hard-coded in `constants.py`. It would be much better to specify this in a config file that you can then point to. Yaml seems like a suitable format for this.

## Proposition
The main training/eval script takes a flag `--spec` that should be given a path to a yaml-file. This yaml file specifies all the data that goes into the model and information about the region the model should be working with.

Current options in `constants.py` that relate to what to plot should all be turned into flags.

The yaml file should be read in and turned into a single object that contains all useful information and can be passed around in the code (since this is needed almost everywhere). Having this as an object means that it can also compute things not directly in the yaml file, such as units of variables that can be retrieved from loaded xarray data.

## Design
Here is an idea for how the yaml file could be laid out with examples:

### Start of file:
```yaml
# Data config for MEPS dataset
---
```
Some comments to keep track of what this specification is about. We don't enforce any content in there.

### Forecast area data configuration
This describes the data configuration for the actual limited area that you are training on. Explicitly, the "inner region", not the "boundary". What is specified is what zarrs to load state, forcing and static grid features from and which variables in these to use for each. 

```yaml
forecast_area:
  zarrs: # List of zarrs containing fields related to state
    fields: # Names on this level are arbitrary
      path: data/meps_example/fields.zarr # Path to zarr
      dims: # Name of dimensions in zarr, to be used for indexing
        time: time
        level: level
        grid: grid_node # Either give "grid" (flattened) dimension or "x" and "y"
    forcing: 
      path: data/meps_example/forcing.zarr
      dims:
        time: time
        level: level
        x: x_coord 
        y: y_coord
  state: # Variables forecasted by the model
    surface: # Single-field variables
      -  2t
      - 10u
      - ...
    atmospheric: # Variables with vertical levels
      - z
      - u
      - v
      - ...
    levels: # Levels to use for atmospheric variables 
      - 700
      - 850
      - ...
    units:
      z: m2/s2
      2t: K
      ...
  forcing: # Forcing variables, dynamic inputs to the model
      surface:
        - land_sea_mask # Dynamic in MEPS
        - ...
      atmospheric:
        - ... # Nothing for MEPS, but we allow it
      levels:
        - ...
  static: # Static inputs
      surface:
        - topography 
        - ...
      atmospheric:
[...]

PR #40, opened 2024-05-25
Three minor bugfixes for data_config.yaml workflow

sadamov:
### Summary
https://github.com/mllam/neural-lam/pull/31 introduced three minor bugs that are fixed with this PR:

- coords_projection is a cached_property and as such is not callable (removed `()`)
- r"" strings are not required in units of `data_config.yaml`
- dictionaries cannot be passed as argsparse, rather JSON strings. This bug is related to the flag `var_leads_metrics_watch`

### DIFF
--- a/train_model.py
+++ b/train_model.py
@@ -197,7 +197,7 @@ def main():
     )
     parser.add_argument(
         "--metrics_watch",
-        type=list,
+        nargs="+",
         default=[],
         help="List of metrics to watch, including any prefix (e.g. val_rmse)",
     )
@@ -205,11 +205,13 @@ def main():
         "--var_leads_metrics_watch",
         type=str,
         default="{}",
-        help="JSON string with variables and lead times to log watched metrics"
-        # e.g. '{"var1": [1, 2], "var2": [3, 4]}'
+        help="""JSON string with variable-IDs and lead times to log watched
+             metrics (e.g. '{"1": [1, 2], "3": [3, 4]}')""",
     )
     args = parser.parse_args()
-    args.var_leads_metrics_watch = json.loads(args.var_leads_metrics_watch)
+    args.var_leads_metrics_watch = {
+        int(k): v for k, v in json.loads(args.var_leads_metrics_watch).items()
+    }
 
     config_loader = config.Config.from_file(args.data_config)
 
--- a/train_model.py
+++ b/train_model.py
@@ -1,4 +1,5 @@
 # Standard library
+import json
 import random
 import time
 from argparse import ArgumentParser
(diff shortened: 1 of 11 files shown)


leifdenby:
While you are doing this I thought of something else with the config: Shouldn't `num_forcing_features` be derived as the length of the `dataset.var_names` list in the config  https://github.com/mllam/neural-lam/blob/main/neural_lam/data_config.yaml#L57? I mean rather than being defined separately.

sadamov:
Hmm, the dataset.var_names are of the **state** not of the **forcings** though. 
In the future when our yaml file covers state,forcings and boundaries I suggest something along these lines 
https://github.com/mllam/neural-lam/blob/feature_dataset_yaml/neural_lam/data_config.yaml 
```        
      self.grid_dim = (
            2 * self.config_loader.num_data_vars("state")
            + grid_static_dim
            + self.config_loader.num_data_vars("forcing")
            * self.config_loader.forcing.window
        )
```
Note, that currently the forcings are calculate in parts in the dataset and in otherparts loaded from .npy files.

leifdenby:
> Hmm, the dataset.var_names are of the **state** not of the **forcings** though.

Ah yes, of course you're right! But but isn't `num_forcing_features == 16` too large then? I thought the forcing features were only radiation etc.

Also, I really like your suggestion about introducing a `num_data_vars` representation somewhere. Should I start a separate issue for this so we can discuss it in more detail?

sadamov:
the forcings should be `GRID_FORCING_DIM = 5 * 3 + 1  # 5 feat. for 3 time-step window + 1 batch-static` (from original `constants.py`)

**=16** 

1. land sea mask (batch-static)
2. top of atmosphere radiation (window)
3. sin hour of day (window)
4. cos hour of day (window) 
5. sin day in year (window)
6. cos day in year (window)

sadamov:
`num_data_vars` is already implemented in the Config class: https://github.com/mllam/neural-lam/blob/5b71be3c68d815e0e376ee651c14f09d801f86de/neural_lam/config.py#L51

train_model.py
```diff
@@ -202,11 +203,13 @@ def main():
     )
     parser.add_argument(
         "--var_leads_metrics_watch",
-        type=dict,
-        default={},
-        help="Dict with variables and lead times to log watched metrics for",
+        type=str,
+        default="{}",
+        help="JSON string with variables and lead times to log watched metrics"
+        # e.g. '{"var1": [1, 2], "var2": [3, 4]}'
```
joeloskarsson, line 209: Can we move the example comment into the help message? Would be useful to see when running `python train_model.py --help`

joeloskarsson (changes requested):
Good fixes, just a small thing about the dict argument.

joeloskarsson:
The coords_projection problem was fixed by merging #17, so merged main in here so we don't see that change.

train_model.py
joeloskarsson: I can fix this myself

joeloskarsson: Actually nvm. I can't get this argument to work, so you'll have to take a look at this @sadamov. It used to be the case that this was a dict mapping from variable index to a list of lead times, rather than variable names. Was this changed to names now?

This is used here, where the dict keys index `var_nums`. I guess this should be the var names?
https://github.com/mllam/neural-lam/blob/879cfec1b49d0255ed963f44b3a9f55d42c9920a/neural_lam/models/ar_model.py#L474-L476

sadamov: image

joeloskarsson (approved):
All good now, tested and it works!

joeloskarsson:
Did not add explicitly to changelog as these are only fixes to #23

PR #51, opened 2024-06-03
[WIP]: Add cirun to github actions to enable gpu instance

khintz:


leifdenby:
I'm closing this as @khintz and I included GPU-based testing in #37 (which moves dependencies to `pyproject.toml`) so that we can test that the new installation procedures work for both cpu and gpu based installs

PR #52, opened 2024-06-03
Fix swapped x and y dimensions in comments and variable names for MEPS data

### LINKED ISSUE #46: Grid Dimension Ordering
Hi, I ran into some issues using `create_mesh.py` for my dataset. I'm wondering if there might be a bug in the documentation for the code? 

The code comments consistently say that a sample has dimensions (N_t', dim_x, dim_y, d_features'). 
https://github.com/mllam/neural-lam/blob/9d558d1f0d343cfe6e0babaa8d9e6c45b852fe21/neural_lam/weather_dataset.py#L88-L90

Similarly, `nwp_xy.npy`, the coordinates of each grid point also has dimensions in the same order (2, N_x, N_y).
https://github.com/mllam/neural-lam/blob/9d558d1f0d343cfe6e0babaa8d9e6c45b852fe21/create_grid_features.py#L29-L32

Now, these lines in `create_mesh.py` imply that the first slice of `nwp_xy.npy` hold the x coordinates, and vice versa. 
https://github.com/mllam/neural-lam/blob/9d558d1f0d343cfe6e0babaa8d9e6c45b852fe21/create_mesh.py#L110-L112

I now refer to `nwp_xy.npy` as `grid_xy`. 
Therefore, given the dimension ordering (2, N_x, N_y), I would reason that  `grid_xy[0][0]` is a length N_y vector holding the x coordinates of *vertical* slice of the dataset. I would expect `grid_xy[0][0]` to have the same x coordinate. However, I see that isn't the case.

```
grid_xy = np.load("data/meps_example/static/nwp_xy.npy")
print(grid_xy.shape)
print(grid_xy[0][0].shape)
print(grid_xy[0][0])

>>>
(2, 268, 238)
(238,)
array([-1.0601221e+06, -1.0501221e+06, -1.0401222e+06, -1.0301222e+06,
...
```

Trying to create a mesh with my own dataset with the ordering (2, N_x, N_y) gives me a buggy graph. However, when changing the ordering to (2, N_y, N_x), the graph seems fine.

Is this an issue with the code comments? Should the correct dimension ordering be N_y, N_x in all cases instead of N_x, N_y?

joeloskarsson:
The x- and y-dimensions for the MEPS data are swapped in comments describing tensor shapes, and also in some variable names. This change swaps from (x, y) ordering to the correct (y, x) ordering.

This fixes #46. See the issue for a more clear description.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   [\#47](https://github.com/mllam/neural-lam/pull/47)
   @joeloskarsson
 
+- Fix incorrect ordering of x- and y-dimensions in comments describing tensor
+  shapes for MEPS data
+  [\#52](https://github.com/mllam/neural-lam/pull/52)
+  @joeloskarsson
+
 ## [v0.1.0](https://github.com/joeloskarsson/neural-lam/releases/tag/v0.1.0)
 
 First tagged release of `neural-lam`, matching Oskarsson et al 2023 publication
--- a/create_grid_features.py
+++ b/create_grid_features.py
@@ -29,14 +29,14 @@ def main():
     # -- Static grid node features --
     grid_xy = torch.tensor(
         np.load(os.path.join(static_dir_path, "nwp_xy.npy"))
-    )  # (2, N_x, N_y)
+    )  # (2, N_y, N_x)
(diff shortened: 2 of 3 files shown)


bet20ICL (approved):
lgtm, I think this covers everything

leifdenby:
LGTM as well! Thanks for this @joeloskarsson

sadamov:
Ah yes this did confuse me back then :) thanks for fixing the comments!

PR #55, opened 2024-06-06
Run ci/cd tests on push to main

SimonKamuk:
This is a minor change for ci/cd to also run on pushes to main (which then includes the push created when a branch is merged). Also changed ci/cd badges to only look at the main branch.


leifdenby (approved):
Looks good to me. Maybe update the changelog to add this PR to your previous one about adding testing too?

Issue #34, opened 2024-05-17
Creation of Static and Forcing Fields


sadamov:
## Summary

For international collaboration a selection of state variables, as well as static and forcing features were selected (see here for reference: [Sheet](https://liuonline-my.sharepoint.com/:x:/g/personal/joeos82_liu_se/Efy1uJuJYHBNi9inM49Ly5wBFQM3_tZB0EOTPBLeNb2V2w?rtime=tUt9a1Ng3Eg)). While the variables are provided by the user in a zarr-archive (usually NWP-model output), the additional forcing and static features are often not available. I suggest to provide the user with the option to generate these fields using a new script  [create_forcings.py](https://github.com/mllam/neural-lam/blob/feature_calculate_forcings/create_forcings.py). The script works for state and boundary domains alike, simple visualizations were provided to make this issue easier to understand. Please feel free to improve on any part of the script and help writing PRs :rocket: 

### New fields

- datetime forcings, 
- land-sea masks, 
- boundary masks, 
- top-of-atmosphere (TOA) radiation forcings. 

## Questions / Issues:

- Does such a script belong in the neuralLAM repo or in a pre-processing step?
- Should it be one or multiple scripts?
- I have a script version for COSMO and MEPS data - should become domain agnostic
- See below for questions specific to each field
- Currently the fields are stored in one .zarr archive. It probably makes sense to specify the zarr archive where each field should append to. 

### `calculate_datetime_forcing(ds, args)`

This function calculates the datetime forcing for the neural LAM model by generating sinusoidal representations of the hour of the day and the time of the year. These features help the model understand diurnal and seasonal variations. Previously these were part of the PyTorch Dataset, provided in `__get_item__`; it certainly makes more sense to calculate these in advance, should be more efficient and allows for cleaner Dataset code.

![image](https://github.com/mllam/neural-lam/assets/45732287/b433cddf-ff9d-4fed-aa0c-c231846fdd21)

### `generate_land_sea_mask(xy_state, tempdir, projection, high_res_factor=10)`

This function generates a land-sea mask for the neural LAM model. The shapefile for the land-sea mask comes from natural-earth at 50m resolution. Percentage of gridcell covered by land is calculated by rasterization of high-res shapefile. This generates static fields, for some regions this mask will vary in time as parts of the sea freeze.
![image](https://github.com/mllam/neural-lam/assets/45732287/6ac99fe7-6312-4825-9ba1-f073e246242f)

[...]

PR #57, opened 2024-06-08
Contributions to #34

TomasLandelius:
I added alternative functions for calculation of toa irradiance and border masks using similar input and output as in the existing ones. Also removed need for pyproj by reformulating projections using cartopy.

### DIFF
--- a/create_forcings.py
+++ b/create_forcings.py
@@ -16,7 +16,6 @@ import requests
 import xarray as xr
 from affine import Affine
 from anemoi.datasets.grids import cutout_mask
-from pyproj import Proj, Transformer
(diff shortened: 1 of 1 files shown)


TomasLandelius:
@sadamov  please have a look.

sadamov (approved):
Thanks a lot @TomasLandelius for your work on this and your patience while I was traveling. I do agree that having simpler solutions that are still robust but faster make a lot of sense for our usecase. I ran your code and it works great.

I  will merge this PR and we should discuss with @leifdenby and the Dev-Team how we want to add these forcing calculations to the roadmap.

PR #65, opened 2024-06-24
Introduce graph classes for graph-based models

joeloskarsson:
## Describe your changes
The actual graph constructions (mesh graphs + m2g and g2m edges) currently take the form as a set of tensors in the code. 
All these tensors are passed around and used in the different model classes in a somewhat unstructured way. See for example: https://github.com/mllam/neural-lam/blob/066efe0a77f1ed741d9eb2f303ee920927d79ad2/neural_lam/models/base_hi_graph_model.py#L43-L45

This also means that there are some more or less hacky ways that we differentiate between hierarchical and "flat" graphs (see `utils.load_graph`): https://github.com/mllam/neural-lam/blob/066efe0a77f1ed741d9eb2f303ee920927d79ad2/neural_lam/utils.py#L111

The idea of this change is to keep the full representation of a graph together in one object. This means that all tensors (edge_index and features) related to the graph sits together in a coherent structure.

At the moment this loads the graph tensors from the same kind of saved `.pt` files as in the current implementation. There has been some discussion about just directly serializing these graph objects, and then using them also in https://github.com/mllam/weather-model-graphs.

Some TODOs:

[ ] Try to turn into dataclasses, while also extending `nn.Module`. Test that moving to GPU still works.
[ ] Implement hierarchical graph class

## Issue Link
Does not directly solve an issue, but should be used in #4.

## Type of change

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging)
- [ ] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form).
## Checklist for reviewers

- [ ] the code readable
- [ ] the code well tested
- [ ] the code documented
- [ ] the code easy to maintain

- [ ] I have added a line to the CHANGELOG describing this change (in section
  reflecting type of change, for example "bug fixes", add section where
  missing)

## Checklist for assignee

- [ ] the tests passing
- Once the PR ready to be merged, squash commits and merge the PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ slurm_log*
 saved_models
 lightning_logs
 data
-graphs
+graphs/*
 *.sif
 sweeps
 test_*.sh
--- a/neural_lam/graphs/base_weather_graph.py
+++ b/neural_lam/graphs/base_weather_graph.py
@@ -1,0 +1,134 @@
+# Standard library
+import os
+
+# Third-party
+import torch
+import torch.nn as nn
+
+
+class BaseWeatherGraph(nn.Module):
+    """
+    Graph object representing weather graph consisting of grid and mesh nodes
+    """
+
+    def __init__(
+        self,
+        g2m_edge_index,
+        g2m_edge_features,
+        m2g_edge_index,
+        m2g_edge_features,
+    ):
+        """
+        Create a new graph from tensors
+        """
+        super().__init__()
+
+        # Store edge indices
+        self.g2m_edge_index = g2m_edge_index
+        self.m2g_edge_index = m2g_edge_index
+
+        # Store edge features
+        self.g2m_edge_features = g2m_edge_features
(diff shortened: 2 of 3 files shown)


joeloskarsson:
From discussion at MLLAM dev meeting 16/6:

- There is a need for one single object representing a graph. This will be increasingly useful for any extensions that train on multiple regions with separate graphs, and make it a lot easier to make the graph itself part of the batch.
- Unclear if this type of self-made class is best solution, or if we should just use a [pyg.HeteroData](https://pytorch-geometric.readthedocs.io/en/2.5.0/generated/torch_geometric.data.HeteroData.html). That could also give better interoperability with https://github.com/ecmwf/anemoi-core/tree/main/graphs.
  - Key question here is if/where to do any checks that indexing is correct. Also need to make sure that solution works with the Inet implementation.

PR #84, opened 2024-10-31
Separate interior state and boundary forcing to only predict state

joeloskarsson:
## Describe your changes

The goal of this PR is to establish a clear separation of the (predicted) state in the interior region and the boundary forcing coming from outside (and potentially overlapping with) the limited area. This PR performs these changes on the *modeling* side. That is from each batch is fetched from the dataset class and onward. Including how the tensors are propagated through the model, loss calculation, evaluation and plotting. This should be complemented with a separate PR for handling the data-loading side of things, where the boundary forcing could come from a separate dataset. That change should then build upon #66, once merged.

*Note:* In order to allow for working on this before the change has been done on the data loading side this currently includes changes in the MEPS npy dataset class that separates state and boundary already in there. This defines the interface between the dataset and model (currently missing https://github.com/mllam/neural-lam/issues/64 from #66, but that can easily be added later) and allows for working on these separately.

### After this change:

* The model will only predict outputs within the limited area considered
* Plots will only include points within the limited area (this could be expanded in a future PR to plot also boundary fields, but would require a mapping between state and boundary forcing dimensions to plot together)
* Graphs will have to be created using a boundary mask as in https://github.com/mllam/weather-model-graphs/pull/34 to make sure that the g2m-component only maps to the interior nodes.

## Issue Link

No issue. This relates to the overall rework of Reading Training Data, but would be good to put as separate point on roadmap.

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [ ] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [ ] I have performed a self-review of my code
- [ ] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [ ] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
[...]

Issue #89, opened 2024-11-21
Output predictions as zarr dataset


khintz:
When running evaluation mode figures are being created. Currently, there is no option to output the predictions.
I would like to implement the option to output predictions as a Zarr dataset.
(WIP: https://github.com/khintz/neural-lam/tree/feature/output).

This should be possible by converting the `predictions` tensor from the function `test_step` to a Zarr dataset.
```python
    def test_step(self, batch, batch_idx):
        """
        Run test on single batch
        """
        prediction, target, pred_std, batch_times = self.common_step(batch)
```
This could be done by utilising the existing function `create_dataarray_from_tensor`
https://github.com/mllam/neural-lam/blob/main/neural_lam/weather_dataset.py#L509
and then simply using xarray to output to zarr.

sadamov:
This is a great idea. Some thoughts that come to mind:

- Given we have implemented a flexible datastore setup for various input datatypes, it might make sense to give the same flexibility in the output. So basically the `mdp` datastore returns a zarr, the `npy` datastore returns a numpy-array, by default. The user can however always override this default and output to any given format.
- There actually is a way to output predictions right now, it is hardcoded as pytorch tensors: https://github.com/mllam/neural-lam/blob/c3c1722f5130bfe374acdbad01e57fc76893e5cd/neural_lam/models/ar_model.py#L491 I don't think it's good to have this inside `plot_example` but rather this could be replaced by your new functionality. (i.e. _Output as tensor without any metadata for debugging._)
- We are currently running inference and auto-regressive rollouts within pytorch's `on_test_...` logic. But actually pytorch would provide a `on_predict_...` set of functions. This PR suggested changing this and also making sure that `predict_step()` is properly used. https://github.com/mllam/neural-lam/pull/45/files Currently, neural-lam overwrites `predict_step` and uses it during training.
- Here a first attempt at prediction output: https://github.com/mllam/neural-lam/pull/44/files It is based on GRIB specifically, but maybe still somewhat useful.

So to summarize, this is clearly required. I suggest to make the output format flexible with sensible defaults. Also to adhere to pytorch's `predict_...` logic (maybe in a separate PR). And finally get rid of the two Draft PRs listed above.

joeloskarsson:
Yes this would be great! (And is very much needed for our ongoing research work, we just haven't had time to put time on it). I agree that it could be nice to allow for different output formats, but I think we should start with zarr+xarray. That can also easily be converted to other things. 

[...]

PR #90, opened 2024-11-21
Feat/boundary dataloader

sadamov:
This PR introduces support for handling boundary data through an additional _optional_ datastore. This allows forcing the model's interior state with boundary conditions during training and evaluation. This PR introduces boundary data handling up until `WeatherDataset.__getitem__` which now returns 5 elements `init_states, target_states, forcing, **boundary**, target_times` (#84 will later implement boundary handling on the model side).
 **Currently boundary datastore must be of type mdp containing forcing category data.** The code wa developed in a way that this requirement can be easily relaxed in the future.

### Motivation and Context:
In weather modeling incorporating realistic boundary conditions is crucial for accurately simulating atmospheric processes. By providing an optional boundary datastore, the model gains flexibility to accept external boundary data, enhancing prediction accuracy and allowing for more complex simulations. The current workflow using n outmost gridcells as boundary will be fully replaced.
The creation of a new ERA5 WeatherBench MDP example datastore demonstrates the implementation of these features and serves as a template for future models.

**Key Changes:**

- Introduced `num_past/future_forcing_steps` and `num_past/future_boundary_steps` as separate arguments and CLI flags.
- Created a new ERA5 WeatherBench MDP example datastore: Developed an example datastore using ERA5 WeatherBench data within the Model-Data-Platform (MDP) framework. 
- Ensured num_grid_points also works for the forcing datastore (i.e. not containing state vars)
- Combined time slicing for state and forcing and make it timestep-based. The same function now handles state forcing and boundary data and matched the corresponding samples based on actual `time` (using minimal time differences between state and forcing/boundary). 
- Updated global configuration to support two datastores: Modified the global configuration to accommodate both state and boundary/forcing datastores simultaneously.
- Added temporal embedding of time_step_diffs: Introduced temporal embeddings by adding time step differences as additional input features to improve the model's temporal awareness.
- Removed interior/exterior masks from the codebase: Eliminated the use of data masks to simplify the codebase. This is now handled by having two separate datastores
- Accept boundary data in common_step but do not return it: Modified the common_step method to accept boundary data as input without returning it, preparing for future boundary/state implementation on the model side.
- Moved the handling of ensemble members to the initialization of WeatherDataset to prevent duplicate time steps in the dataset.

**Bugfix:**
- Fixed a bug in analysis time retrieval in the MEPS store: Corrected an issue with retrieving analysis_time from the MEPS data store where duplicate time steps for one member were returned.

**Notes:**
- I have introduced strict assertion that `time` , `analysis_time` and `elapsed_forecast_time` all have a consistent step size. This is not strictly necessary but probably what most user want and it helps with the temporal encoding.

## Introduced Dependencies
- "gcsfs>=2021.10.0" Where should we download era5 from, do we want this dependency to Google Cloud Storage?

## TODOs:
- [ ] Upload MEPS example with dates close to Danra and next to each other for test/val/train @leifdenby This allows for testing both npy and mdp datastores with the same boundary datastore and limits the required data.

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


neural_lam/models/ar_model.py
```diff
@@ -108,25 +110,15 @@ def __init__(
         self.grid_dim = (
             2 * self.grid_output_dim
             + grid_static_dim
-            + num_forcing_vars
+            # Factor 2 because of temporal embedding or windowed features
+            + 2
+            * num_forcing_vars
             * (num_past_forcing_steps + num_future_forcing_steps + 1)
```
joeloskarsson, line 116: Could you explain more why there is a factor 2 here? I understand that there is an additional temporal embedding variable (time delta between state and forcing), but would that not just be `num_forcing_vars + 1`? What am I missing? The different forcing variables should not have different time deltas so you need twice as many vars?

joeloskarsson: Oh right, it's more thatn `+ 1` since there is the windowing as well. But then I would expect `(num_forcing_vars + 1) * (num_past_forcing_steps + num_future_forcing_steps + 1)` rather than the `* 2`.

pyproject.toml
```diff
@@ -25,7 +25,8 @@ dependencies = [
     "torch>=2.3.0",
     "torch-geometric==2.3.1",
     "parse>=1.20.2",
-    "dataclass-wizard>=0.22.3",
+    "dataclass-wizard<0.31.0",
+    "gcsfs>=2021.10.0",
```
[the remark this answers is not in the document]
sadamov: moved to dev-group in 4c175452af54fa4833fd9ac67bb4b1b36cdaa777

neural_lam/weather_dataset.py
```diff
@@ -199,82 +354,64 @@ def _slice_state_time(self, da_state, idx, n_steps: int):
...
+        # might be dealing with a datastore_boundary
+        if "analysis_time" in da_forcing_boundary.dims:
+            # Select the closest analysis time in the forcing/boundary data
+            # This is mostly relevant for boundary data where the time steps
+            # are not necessarily the same as the state data. But still fast
+            # enough for forcing data where the time steps are the same.
+            idx = np.abs(
+                da_forcing_boundary.analysis_time.values
+                - self.da_state.analysis_time.values[idx]
+            ).argmin()
```
joeloskarsson, line 399: For the case where boundary forcing time steps are different than state time steps:

Thinking about how this will be used downstream in the model I am not sure about the choice to find the *closest* analysis time. I would rather for state time t find the previous or equal boundary forcing time (<= t). The point being that the same index of boundary forcing features always should correspond to "the previous boundary forcing" or "the next boundary forcing". This relates to the discussed idea of the modeling learning to implicitly interpolate between the boundary forcing at different times.

Consider this setup, with `num_past_boundary_steps = 0` and `num_future_boundary_steps = 1`.
boundary_forcing_timescales
And let's say we have more $X$ in between 0 h and 6 h. If I am not misunderstanding this, with the current approach $X_\text{1h}$ would be paired with a window of $F_\text{0h}$ and $F_\text{6h}$ (as desired). But for $X_\text{5h}$, the closest boundary forcing time is 6 h, meaning it would be paired with $F_\text{6h}$ and $F_\text{12h}$. This makes this "implicit interpolation" to boundary forcing at 5 h impossible, as the information from $F_\text{0h}$ is not included. Now one could technically solve this by setting `num_past_boundary_steps = 1`, but we would make the problem harder for the network, as it has to change between "interpolating" between the first boundary forcings in the window and the last two. I hope this makes sense, I realize this got a bit involved.

I think that what I propose could potentially also simplify the argmin-expressions, if you just keep track of the timedeltas for both the boundary forcing and state time series. It seems to me that this should be doable without comparing to  all times, with just `.sel`.

neural_lam/weather_dataset.py
```diff
@@ -294,46 +431,115 @@ def _slice_forcing_time(self, da_forcing, idx, n_steps: int):
...
+            # Stack the 'feature' and 'window' dimensions and add the
+            # time step differences to the existing features as a temporal
+            # embedding
+            da_windowed = da_windowed.stack(
+                {stacked_dim: ("forcing_feature", "window")}
+            )
+            da_windowed = xr.concat(
+                [da_windowed, da_windowed.time_diff_steps],
+                dim="forcing_feature_windowed",
+            )
```
joeloskarsson, line 525: How will the concatenation of these time deltas work with e.g. `BaseDatastore.get_vars_names(category="forcing")`? I guess they will not be included there. But that also does not reflect the forcing windowing, so maybe one should not make the assumption that what is returned there maps to the indices of the forcing tensor returned from the dataset? That would require the datastore to have knowledge of the WeatherDataset config, which we probably don't want :thinking: 

I fully agree that we need these time-delta features, but it's tricky how to handle them. We've been talking about doing some form of sinusoidal embeddings of them later in the model. Would it even make sense to return these separately? Since if we  want to do something special (embedding) with these features we would otherwise have to extract them from the boundary forcing tensor (assuming that they are the last features). If we would return them separately we should maybe create some form of custom batch-object, as it would start to be a little unstructured to have this tuple of 7 objects.

neural_lam/models/ar_model.py
sadamov: You are right, all forcing/boundary features share the same temporal embedding. I fixed this in https://github.com/mllam/neural-lam/pull/90/commits/16d5d04bbd9e49a1fe53e56ff95e22d565326c67 and updated the time slicing test to cover also forecast-type data: https://github.com/mllam/neural-lam/pull/90/commits/8fd7a10fc1d01c998641baf228dfa66f8630280c @leifdenby you were also asking for this

PS: What is in terms of memory/computational efficiency the best way to return such a tensor that repeats over 2 dimensions `time, grid_index`? Should we enforce `signed Integers` here for improved memory management, even though the `time_step_diff` can be `float`?

neural_lam/weather_dataset.py
sadamov: I see your point, but what if the closest time step from the boundary is +1h in the future of state_time[idx]. With `num_past_boundary_steps = 1; num_future_boundary_steps = 0`. In your setup the two boundary steps would be -5 and -11h away from state_time[idx] giving the model a much harder problem than just using the boundary at +1h and the one at -5h.

I agree with .sel being a much simpler and faster solution than my global time diff matrices. :heavy_check_mark:

neural_lam/weather_dataset.py
sadamov: Yes, BaseDatastore.get_vars_names(category="forcing") will not return the windowed_features + temporal_embeddings. This is important for saving model predictions in https://github.com/mllam/neural-lam/issues/89, too.  We don't need them in the model output, but want to keep the original features exactly in order.

- ordering the forcing and boundary return, to always have the temporal embedding as the last feature, sound like an easy option for now
- I am still a bit confused how exactly the temporal embedding will be used in the model. The temporal embedding is the same for all `grid_points, time` and maps to 'windowed_forcing_features` based on the window-size. Can one just leave these additional temporal embeddings as additional input feature channels (sin(embedding)), or should we directly _merge_ them with the forcing_feature in some sort of batch-norm in ar_model? Is there a world where we can already merge the embedding and forcing_feature in weatherDataset?

neural_lam/models/ar_model.py
joeloskarsson: Nice! Should the lines I commented on in `ar_model.py` also be updated then?

Regarding memory management: You can have a tensor with large dimensions that share the same memory, e.g. for all grid points. If you construct the tensor on GPU you can achieve this with https://pytorch.org/docs/stable/generated/torch.Tensor.expand.html (and I think also other ways). But I am unsure what happens when you create such a tensor on CPU and then pass it o GPU. Maybe something that should be tested (e.g. by checking that memory is still shared when it's on the GPU by doing some in-place operation).

neural_lam/weather_dataset.py
joeloskarsson: > I see your point, but what if the closest time step from the boundary is +1h in the future of state_time[idx]. With num_past_boundary_steps = 1; num_future_boundary_steps = 0. In your setup the two boundary steps would be -5 and -11h away from state_time[idx] giving the model a much harder problem than just using the boundary at +1h and the one at -5h.

Yes, that's a bad situation, but I think this just means that the user has made a bad choice with `num_future_boundary_steps = 0` given their data. Likely you would always want to set that to at least 1 when you have this kind of time-step mismatch. Even in that situation I think it's still preferable over the model having to adjust to the forcing features at one specific index sometimes being from the future and sometimes being from the past. That information is in the delta-time features, but using those deltas in that way seems like a harder problem for the model to untangle.

sadamov: so basically you are saying its easier for the model to have forcing either (1) in the past or (2) in the past and future. Than it is to have forcing features in (1) + (2) and (3) in the future. Okay I can see that. and there might be some scenario where it is technically of scientifically not valid to choose data from the future. So having the default to be the closest past time step is sound. In addition we can use xarray.sel() which is more efficient and lazy. ok I'm convinced. will get to work on this.

neural_lam/weather_dataset.py
joeloskarsson: > Can one just leave these additional temporal embeddings as additional input feature channels (sin(embedding))

I guess we should leave it open to different model implementations what to do with these. What I am thinking of first is to just do the sin(embedding) and feed them as additional input feature channels. Then we could expand on that later as needed.

> or should we directly merge them with the forcing_feature in some sort of batch-norm in ar_model?

That's another option, to map from these features to the mean and std-dev used in LayerNorms (akin to how time-embeddings are often used in diffusion-models and others). It's a bit trickier to implement, and might not really be neccesary here, but still potentially interesting.

> Is there a world where we can already merge the embedding and windowed_forcing_features in weatherDataset?

I don't think that would be possible (considering the LayerNorm setup discussed above). As the use of theses features in the LayerNorms involves trainable parameters, and they act on intermediate representations in the model, this would have to be done on the model side of things.

I suppose we could do the sin(embedding) already in WeatherDataset, if that does not involve trainable parameters. But I don't think we want to, as it should be better to do those computations on GPU (similar to how we want to move standardization to GPU).

sadamov: okay thanks for the explanations. While not totally relevent here, it's still good to have the bigger picture in mind. so for now I just focus un returning all temporal embedding features as the last channels in `forcing` or `boundary`

neural_lam/weather_dataset.py
joeloskarsson: I guess another simple way to put what I mean is: If I look at one specific feature dimension in the boundary forcing tensor (e.g. boundary_forcing[..., 21]) , I think it's good if the value there either always comes from the past or always comes from the future. That seems easier for the model to handle.

neural_lam/models/ar_model.py
joeloskarsson: Great!

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
joeloskarsson, line 1: I think this file needs a couple more adjustments:

* Other instances of WeatherDataset requires the boundary datastore argument
* All batches now contain 5 objects, so an additional boundary forcing has to be ignored when unpacking batches in the loops

joeloskarsson: Can cherry-pick this one: https://github.com/joeloskarsson/neural-lam-dev/commit/ebfd0bd2f883301939aafd72d08909a976ec0cea

neural_lam/models/ar_model.py
joeloskarsson: @sadamov Does this actually work correctly now? I am getting dimension mismatches when building MLPs with the dimensions as given after https://github.com/mllam/neural-lam/commit/252a33cd5903aca793f480e871912cc8b8616df2. In `_process_windowed_data` I am going from 6 forcing features (2 features windowed over 3 times) to 12 features. If we only include the delta time once per time in the window this should be 6 features to 6+3 = 9.

tests/datastore_examples/mdp/era5_1000hPa_danra_100m_winds/config.yaml
```diff
@@ -0,0 +1,12 @@
+datastore:
+  kind: mdp
+  config_path: danra.datastore.yaml
+datastore_boundary:
+  kind: mdp
+  config_path: era5.datastore.yaml
```
leifdenby, line 6: Could we maybe do something like following instead:

```suggestion
boundary:
  datastore:
    kind: mdp
    config_path: era5.datastore.yaml
```

```suggestion
boundary:
  point_width: 30
```

And expand the config spec so that either `boundary.point_width` or `boundary.datastore` should be defined? Then we could keep the existing option where we force with the LAM domain data on the boundary. Just in cases where people haven't got boundary data or want to examine the effect of changing the two. Just an idea :)

joeloskarsson: If you want that setup you can do the splitting of the boundary from the interior first, then build two mdp-zarrs based on those and load in this setup. I don't think we need to preserve the option to load the boundary forcing from the same datastore here.

sadamov: I think removing all the code related to these existing `interior_masks` is a good thing, as it makes the code easier to understand. Creating two zarrs seems straight-forward enough. And for those quickly wanting to test something they can:
- use the future hello_world example
- run an experiment with boundary=0, which should technically be possible, right @joeloskarsson

neural_lam/weather_dataset.py
sadamov: Okay I fixed this now in https://github.com/mllam/neural-lam/pull/90/commits/cfe1e278ae16a4ec8e19dc3f2db79e976e8014d6 
the temporal embedding has size `window` and is not repeated for each feature.
the embedding is concatenated to the existing `forcing_feature_windowed` at the very end

neural_lam/weather_dataset.py
sadamov: Agree with you, implemented this with xarray's `.sel` and `method=pad`, which does exactly what you describe.
https://github.com/mllam/neural-lam/pull/90/commits/cfe1e278ae16a4ec8e19dc3f2db79e976e8014d6
Sorry this commit is a bit overloaded, introducing more than one change. Let me know if confusing!

neural_lam/models/ar_model.py
sadamov: This now works and is extensively tested using new and old tests:
Tests: https://github.com/mllam/neural-lam/pull/90/commits/1d53ce7b86ee8c936a1c8c2fd9bad58bb672b844
Fixing slicing/embedding: https://github.com/mllam/neural-lam/pull/90/commits/cfe1e278ae16a4ec8e19dc3f2db79e976e8014d6

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
sadamov: True I forgot about this file, pardon. Cherry-pick complete.

tests/datastore_examples/mdp/era5_1000hPa_danra_100m_winds/config.yaml
joeloskarsson: Yes, I have made the boundary forcing completely optional, so everything can be ran without it (tested with the DANRA example). This will be the behavior if no boundary datastore is given.

neural_lam/models/ar_model.py
joeloskarsson: Now it seems correct, yes :smile:

joeloskarsson:
I think that if we merge `main` into here the tests will pass. They seem to have failed for the same  reason as on main, which was fixed with #94.

tests/test_time_slicing.py
```diff
@@ -16,40 +16,76 @@ class SinglePointDummyDatastore(BaseDatastore):
     root_path = None
 
     def __init__(self, time_values, state_data, forcing_data, is_forecast):
```
leifdenby, line 18: thank you for making this datastore for testing more complete!

tests/test_time_slicing.py
```diff
@@ -125,22 +190,138 @@ def test_time_slicing_analysis(
...
+    sample = dataset[0]
+
+    init_states, target_states, forcing, _, _ = [
+        tensor.numpy() for tensor in sample
+    ]
+
+    # Compute expected initial states and target states based on ar_steps
+    offset = max(0, num_past_forcing_steps - INIT_STEPS)
+    init_idx = INIT_STEPS + offset
+    expected_init_states = STATE_VALUES_FORECAST[0][offset:init_idx]
```
leifdenby, line 285: These tests are so great. I was just thinking it might be nice to make it explicit here that this `0`-index is due to the fact we only have implemented sampling from the first ensemble member. That is the reason why? Then it will be clearer how to update the tests once we add sampling more ensemble members

leifdenby (commented):
First read through: this was joy to read 😊 I have a few comments, but in general I think this is very close to done.

neural_lam/weather_dataset.py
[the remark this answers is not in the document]
joeloskarsson: Looking through the updated version of https://github.com/sadamov/neural-lam/blob/feat/boundary_dataloader/tests/test_time_slicing.py was quite helpful for me when reading through this :) That is still only for one Datastore at a time, but good for seeing how the forcing windowing works both for analysis and forecast data.

neural_lam/weather_dataset.py
```diff
@@ -294,46 +444,133 @@ def _slice_forcing_time(self, da_forcing, idx, n_steps: int):
...
+            time_diff_steps,
+        )
 
-        return da_concat
+        return da_state_sliced, da_forcing_matched
+
+    def _process_windowed_data(self, da_windowed, da_state, da_target_times):
+        """Helper function to process windowed data. This function stacks the
+        'forcing_feature' and 'window' dimensions and adds the time step
+        differences to the existing features as a temporal embedding.
```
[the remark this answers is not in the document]
joeloskarsson: I think this is highly related. The temporal embedding (which we should probably not refer to as such, as this is not an embedding, so from here on: the delta times) is the time differences for the different entries in the window. After these have been stacked it is not clear to me how you would know what the delta-times are, since you've got rid of the window dimension.

tests/test_datasets.py
```diff
@@ -66,10 +82,20 @@ def test_dataset_item_shapes(datastore_name):
     assert forcing.ndim == 3
     assert forcing.shape[0] == N_pred_steps
     assert forcing.shape[1] == N_gridpoints
-    assert forcing.shape[2] == datastore.get_num_data_vars("forcing") * (
+    # each time step in the window has one corresponding temporal embedding
```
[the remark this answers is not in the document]
joeloskarsson: Agree that this should not be referred to as an embedding (this is rather what we will use to create an embedding later). I've also been sloppily calling it this :sweat_smile: I think something like time delta is more clear (forcing increments makes me think it is the increment in forcing values, rather than in time).

neural_lam/weather_dataset.py
```diff
@@ -94,6 +125,98 @@ def __init__(
...
+
+                # Calculate required bounds for boundary using its time step
+                boundary_required_time_min = (
+                    state_time_min
+                    - self.num_past_forcing_steps * self.time_step_boundary
+                )
+                boundary_required_time_max = (
+                    state_time_max
+                    + self.num_future_forcing_steps * self.time_step_boundary
+                )
```
joeloskarsson, line 204: I think there is a typo here w.r.t. which variables to use

```suggestion
                boundary_required_time_min = (
                    state_time_min
                    - self.num_past_boundary_steps * self.time_step_boundary
                )
                boundary_required_time_max = (
                    state_time_max
                    + self.num_future_boundary_steps * self.time_step_boundary
                )
```

neural_lam/config.py
```diff
@@ -168,4 +172,15 @@ def load_config_and_datastore(
         datastore_kind=config.datastore.kind, config_path=datastore_config_path
     )
 
-    return config, datastore
+    if config.datastore_boundary is not None:
```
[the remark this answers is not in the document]
sadamov: Fixed in 8cc608dfc4f8b4e97a931c58e5170c2578281fef

neural_lam/datastore/mdp.py
```diff
@@ -415,8 +380,17 @@ def grid_shape_state(self):
...
         """
-        ds_state = self.unstack_grid_coords(self._ds["state"])
-        da_x, da_y = ds_state.x, ds_state.y
+        # Boundary data often has no state features
+        if "state" not in self._ds:
+            warnings.warn(
+                "no state data found in datastore"
+                "returning grid shape from forcing data"
+            )
+            ds_forcing = self.unstack_grid_coords(self._ds["forcing"])
```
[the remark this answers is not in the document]
sadamov: fixed in 857f7482e9e04c90a7112e8063bc6e449c9971c6

neural_lam/weather_dataset.py
```diff
@@ -22,6 +22,8 @@ class WeatherDataset(torch.utils.data.Dataset):
     ----------
     datastore : BaseDatastore
         The datastore to load the data from (e.g. mdp).
+    datastore_boundary : BaseDatastore
+        The boundary datastore to load the data from (e.g. mdp).
```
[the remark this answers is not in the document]
sadamov: fixed in d0a6f2425f473b06db99740f295c0d452d003281

neural_lam/weather_dataset.py
```diff
@@ -36,33 +38,62 @@ class WeatherDataset(torch.utils.data.Dataset):
...
+                "A non-empty state dataarray must be provided. "
+                "The datastore.get_dataarray() returned None or empty array "
+                "for category='state'"
+            )
         self.da_forcing = self.datastore.get_dataarray(
             category="forcing", split=self.split
         )
+        # XXX For now boundary data is always considered mdp-forcing data
+        if self.datastore_boundary is not None:
+            self.da_boundary = self.datastore_boundary.get_dataarray(
```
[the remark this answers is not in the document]
sadamov: fixed in d0a6f2425f473b06db99740f295c0d452d003281

neural_lam/weather_dataset.py
```diff
@@ -36,33 +38,62 @@ class WeatherDataset(torch.utils.data.Dataset):
...
+        self.datastore_boundary = datastore_boundary
         self.num_past_forcing_steps = num_past_forcing_steps
         self.num_future_forcing_steps = num_future_forcing_steps
+        self.num_past_boundary_steps = num_past_boundary_steps
+        self.num_future_boundary_steps = num_future_boundary_steps
 
         self.da_state = self.datastore.get_dataarray(
             category="state", split=self.split
         )
+        if self.da_state is None:
```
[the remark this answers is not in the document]
sadamov: fixed in 71b52b248caef8ffb15a0147988f60db9e41fcda

neural_lam/weather_dataset.py
```diff
@@ -94,6 +125,98 @@ def __init__(
...
+                        f"Required start: {boundary_required_time_min}, "
+                        f"but boundary starts at {boundary_time_min}."
+                    )
+
+                if boundary_time_max < boundary_required_time_max:
+                    raise ValueError(
+                        f"Boundary data ends too early."
+                        f"Required end: {boundary_required_time_max}, "
+                        f"but boundary ends at {boundary_time_max}."
+                    )
```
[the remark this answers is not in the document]
sadamov: I moved `get_time_step` into `utils.py` where I also created a new stateless function called `check_time_overlap`.
Currently the logic remains uchanged. I will however in a future commit change this from 1. to 2.

1. tell the user to include sufficient time-steps and have a very clear requirement and warning
2. OR: expect most users to chose the same datetimes for boundary and forcing and give a warning how many time steps were removed from interior

neural_lam/weather_dataset.py
```diff
@@ -160,32 +293,75 @@ def __len__(self):
                 - self.num_future_forcing_steps
             )
 
-    def _slice_state_time(self, da_state, idx, n_steps: int):
+    def _get_time_step(self, times):
```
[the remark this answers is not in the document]
sadamov: fixed in https://github.com/mllam/neural-lam/commit/b69056341eb288c5a439fcdf038ae3172aee52c3

neural_lam/weather_dataset.py
```diff
@@ -126,7 +259,7 @@ def __len__(self):
                 warnings.warn(
                     "only using first ensemble member, so dataset size is "
                     " effectively reduced by the number of ensemble members "
-                    f"({self.da_state.ensemble_member.size})",
+                    f"({self.datastore._num_ensemble_members})",
```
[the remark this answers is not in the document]
joeloskarsson: While we do not handle the case of ensemble reanalysis data atm, it might be nice to allow for extending to that in the future. Therefore I would probably prefer to keep `is_ensemble` and `is_forecast` as orthogonal properties. And not do `is_forecast = (num_ensemble_members > 0)` (if that is how you propose to do that derivation?).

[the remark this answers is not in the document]
joeloskarsson: I anyhow think that better handling of `is_forecast`/`is_ensemble` is orthogonal to this PR, and is better addressed separately to prevent scope creep of this change.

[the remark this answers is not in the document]
sadamov: I have implemented this in BaseDatastore.num_ensemble_members
Having  `BaseDatastore.num_ensemble_members` allows to treat `is_ensemble` and `is_forecast` as 2 truly independent attributes.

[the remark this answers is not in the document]
sadamov: Haha Joel I think we had the same ideas there. Hope my solution is a good compromise that allows for future mdp-ensembles, while keeping this PR as simple as possible.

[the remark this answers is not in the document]
sadamov: Haha Joel I think we had the same ideas there. Hope my solution is a good compromise that allows for future mdp-ensembles, while keeping this PR as simple as possible.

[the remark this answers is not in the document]
sadamov: I forgot to declare the new abstract method in the mdp datastore: num_ensemble_members

neural_lam/weather_dataset.py
```diff
@@ -294,46 +444,133 @@ def _slice_forcing_time(self, da_forcing, idx, n_steps: int):
...
+                forcing_time_step = self.forecast_step_forcing
+                state_time_step = self.forecast_step_state
+            else:
+                forcing_time_step = self.time_step_forcing
+                state_time_step = self.time_step_state
+            time_diff_steps = (
+                da_forcing_matched["window"]
+                * (forcing_time_step / state_time_step),
+            )
+        time_diff_steps = da_forcing_matched.isel(
```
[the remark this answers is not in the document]
sadamov: I think `time_deltas` is a nice name. Resolving here as we have the same discussion in the comment above.

tests/test_time_slicing.py
sadamov: No these are indexing `analysis_time` as can be seen here: 
https://github.com/sadamov/neural-lam/blob/98c54d9856f4dead1ed8fd8951a69795931cb5e6/tests/test_time_slicing.py#L111

I added some comments to the tests to make this indexing clear. 4a278fd9dea3fd8ce7bce91e88cb6259fa42cc6a Ensemble functionality must be added at a later date.

neural_lam/weather_dataset.py
sadamov: OOps, thank you :pray: I have in the meantime defined a new function that takes over the min/max time calculations: https://github.com/sadamov/neural-lam/blob/c82d22ba9749c8f1fcd04c7bd7e1301881951622/neural_lam/weather_dataset.py#L180 
This issue should be fixed

neural_lam/weather_dataset.py
```diff
@@ -94,6 +126,66 @@ def __init__(
...
+        # The boundary data is allowed to have a different time_step
+        # Check that the boundary data covers the required time range
+        check_time_overlap(
+            self.da_state,
+            self.da_boundary_forcing,
+            da1_is_forecast=self.datastore.is_forecast,
+            da2_is_forecast=self.datastore_boundary.is_forecast,
+            num_past_steps=self.num_past_boundary_steps,
+            num_future_steps=self.num_future_boundary_steps,
+        )
```
sadamov, line 187: Joel Oskarsson
  Tuesday at 2:16 PM
I've been trying out loading data in neural-lam that I have created using the DANRA and ERA5 configs from https://github.com/leifdenby/mllam-data-prep/tree/feat/crop-with-other-dataset. In those configs both datasets cover exactly the same time period. Running neural-lam with num_past_boundary_steps > 0 raises this error https://github.com/sadamov/neural-lam/blob/f0a7046b7ce0c5c955762b67ebe017fdef4eae4e/neural_lam/weather_dataset.py#L207 (even with the fix https://github.com/mllam/neural-lam/pull/90#discussion_r1888481106). This makes sense, as there is no boundary data before the first time point for the interior. However, an alternative behavior would be to adjust the first time point for the interior (e.g. ignoring the first time steps of the interior data, until data loading is possible for the set num_past_boundary_steps ), to still allow for training in this case. What's your thoughts on this? Now one has to always leave in some extra time steps in the boundary forcing, depending on num_past_boundary_steps , which has to be decided beforehand when the dataset is created. But maybe that makes more sense than allowing for having a different number of samples depending on num_past_boundary_steps ?
------------------------------------------------------
Simon Adamov:
true this is indeed by design.
tell the user to include sufficient time-steps and have a very clear requirement and warning
OR: expect most users to chose the same datetimes for boundary and forcing and give a warning how many time steps were removed from interior
As long as only a few timesteps are removed from interior, I think option 2 is nicer to handle. But no user would want half their training data to be ignored "quietly".
So, I think both options are fine, as long as we have clear warnings and error messages. Is 2 your preference?
----------------------------------------------------------------
Joel Oskarsson
  Wednesday at 5:30 PM
That's good ideas. I agree that we should really avoid cutting away much of the training data without any proper warning. But cutting away a few timesteps is not really a problem (but could still show a warning explaining how much is being removed). Yea, option 2 is quite natural I think, since it would be nice to prepare the data by just specifying the exact same time period for the interior and boundary when building corresponding zarrs :grin: I'm thinking that this might be easy to achieve by "just" slicing away a few time steps in the interior xr.Dataarray used in WeatherDataset, but not I have not looked at any details of what such an implementation would look like.

neural_lam/weather_dataset.py
```diff
@@ -11,6 +11,7 @@
 
 # First-party
 from neural_lam.datastore.base import BaseDatastore
+from neural_lam.utils import check_time_overlap, get_time_step
 
 
 class WeatherDataset(torch.utils.data.Dataset):
```
sadamov, line 17: Implement the relevant changes on the WeatherDataset side that were required for succesfull experiments:
> Here is the exact changes that I made to neural-lam to make everything works despite issue 1 above: https://github.com/joeloskarsson/neural-lam-dev/commit/2f0e6a0137244d1073ced1db6165de1bca396be3

neural_lam/weather_dataset.py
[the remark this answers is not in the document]
joeloskarsson: I see, we were writing at the same time :smile: good! Would it make sense to already in BaseDatastore default `num_ensemble_members` to 1 (not an ensemble) and derive `is_ensemble` from `num_ensemble_members > 1`? One could then override `num_ensemble_members` in subclasses that allow for working with ensemble data.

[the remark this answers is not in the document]
sadamov: Yep, that makes more sense, implemented in: c0c50d5e44580c6a8e486d44cee19be6ed1fd847

neural_lam/weather_dataset.py
sadamov: I suggest to implement an additional stateless function called `crop_time_if_needed` in 7e5797e18f5a251441e98ef9ad63f412c06ef409. This function gets called automatically before any time-slicing operations in WeatherDataset. I left all the existing checks and warning in the code, can't hurt. Added additional print statement about how many time steps were removed from the state data. Let me know what you think!

sadamov:
@leifdenby @joeloskarsson I have given this a thorough rework and implemented all your feedback from here and from slack. there are two threads unresolved, mostly for you to have a chance to intervene if not in agreement with my solution. Please give this your final review so we can merge soon. It is my understanding that everything up to the return of WeatherDataset should now be ready to run "realistic" experiments with neural-lam.

PS: locally the tests pass for me
image

neural_lam/weather_dataset.py
joeloskarsson: Something that struck me when looking at this again is that it would often actually be possible to create all the boundary forcing with num_past_boundary_steps > 0 even in cases when the interior and boundary covers exactly the same period. This  because boundary forcing is really only needed from time step 3 of the interior data. The first two time steps of the interior data are only used as initial conditions. However, whether the needed time step is included in the data or not for num_past_boundary_steps=1 also depends on the length of time steps of interior data vs boundary data.

So overall I think handling this gets terribly overcomplicated, for just getting one or two extra samples. So much better to stick with this solution with `crop_time_if_needed` :)

joeloskarsson:
As we continued work on this, and there is a lot of discussion here (some of it still relevant, some not), it is probably better to make a new PR for this and look over it once again. I changed this to a draft PR for now.

sadamov:
Closing this PR in favor of a fresh implementation rebased on current `main`.

Since this was opened, `main` has undergone significant refactoring (notably the Forecaster/StepPredictor decomposition in #208), making this branch unmergeable without a full rewrite. As @joeloskarsson noted in https://github.com/mllam/neural-lam/pull/90#issuecomment-3703909477, a new PR is the right path forward.

A new PR for boundary datastore support in `WeatherDataset` and config will be opened shortly, based on the work in the [research branch](https://github.com/joeloskarsson/neural-lam-dev/tree/research) and rebased on current `main`. Tracking continues in #138.

PR #103, opened 2025-01-22
Add multi-node-training

SimonKamuk:
## Describe your changes

This PR adds support for multi-node GPU training using the SLURM job scheduler. The changes allow setting the number of nodes with the cli argument `num_nodes`. It is also possible to select a subset of visible GPU's using the argument `devices` (only when not using SLURM)

Replaces https://github.com/mllam/neural-lam/pull/26 with a simpler method based on advice from @sadamov 

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/tests/datastore_examples/mdp/danra_100m_winds/config.yaml
+++ b/tests/datastore_examples/mdp/danra_100m_winds/config.yaml
@@ -7,5 +7,3 @@ training:
     weights:
       u100m: 1.0
       v100m: 1.0
-      r2m: 1.0
-      t2m: 1.0
--- a/tests/datastore_examples/mdp/danra_100m_winds/config.yaml
+++ b/tests/datastore_examples/mdp/danra_100m_winds/config.yaml
@@ -7,3 +7,5 @@ training:
     weights:
       u100m: 1.0
       v100m: 1.0
+      r2m: 1.0
+      t2m: 1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased](https://github.com/mllam/neural-lam/compare/v0.3.0...HEAD)
 
+### Added
+- Add support for multi-node training.
+[\#103](https://github.com/mllam/neural-lam/pull/103) @simonkamuk @sadamov
+
 ## [v0.3.0](https://github.com/mllam/neural-lam/releases/tag/v0.3.0)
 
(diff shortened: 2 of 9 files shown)


tests/datastore_examples/mdp/danra_100m_winds/config.yaml
```diff
@@ -7,3 +7,5 @@ training:
     weights:
       u100m: 1.0
       v100m: 1.0
+      r2m: 1.0
```
sadamov, line 10: This seems unrelated, maybe better remove it?

sadamov (approved):
Thank @SimonKamuk for putting this together! I like this simple approach a lot. @joeloskarsson you previously were not too impressed by myself adding `device` as a new flag. But would you agree that in a non-slurm world it makes sense?

sadamov:
I have closed the legacy #26 PR and linked the corresponding issue #1 and milestone here. I am not sure if we have a procedure in place, but you mentioned reducing the scope of v0.4.0. We should at least post this in the slack-channel, if we can't wait until the next dev-meeting.

joeloskarsson:
I suppose I have viewed it as better to use `CUDA_VISIBLE_DEVICES` environment variable than to integrate device selection into neural-lam, but I don't know if there are any advantages or disadvantages to this. It gets tricky when we start talking HPC setups as it can be quite different for different users and I'd prefer to not make assumptions about the users env. But as I see this change I don't think it makes any more assumptions than lightning does, which should be the minimal.

tests/datastore_examples/mdp/danra_100m_winds/config.yaml
SimonKamuk: You're right, I added this to use some existing data I had for testing. I forgot to remove it :)

SimonKamuk:
The reason why I wanted to select devices, is because we at DMI also have some shared GPU's which are not on a SLURM cluster, and I wanted to be able to run some things there without reserving the entire system. I didn't think of using CUDA_VISIBLE_DEVICES, and I have no strong opinion either way, so if you prefer, I'll just remove the devices argument again, and specify the way to do it in the readme 😄

joeloskarsson:
No strong opinions on my side either really, it doesn't hurt to have the --devices option as long as it defaults to auto.

SimonKamuk:
Alright cool, then I think this is ready to merge. FYI I also changed to only print on rank 0 everywhere to keep logs clearer.

sadamov:
looks good to merge. The whole printing/logging will soon be redone anyways. So it makes sense to keep it simple here. Please go ahead with the merge :+1:

PR #104, opened 2025-01-24
Add writing to zarr dataset for eval-mode of trained models

### LINKED ISSUE #89: Output predictions as zarr dataset

leifdenby:
## Describe your changes

Adds new CLI flag to `neural_lam.train_model` called `--save-eval-to-zarr-path <path-to-dataset>` which can be added when running neural-lam in `eval` mode (i.e. `neural_lan.train_model --eval ...`) to write the predictions to a zarr dataset stored in `<path-to-dataset>`. This functionality is motivated by our want to be able to store model predictions for later verification.

Example of usage:

Model trained with 

```bash
$> pdm run python -m neural_lam.train_model --config_path tests/datastore_examples/mdp/danra_100m_winds/config.yaml --hidden_dim 2 --epochs 1 --ar_steps_train 3 --ar_steps_eval 3 --graph 1level
```

used for inference with

```bash
$> pdm run python -m neural_lam.train_model \
   --config_path tests/datastore_examples/mdp/danra_100m_winds/config.yaml \
   --hidden_dim 2 --epochs 1 --ar_steps_train 1 --ar_steps_eval 3 --eval val \
   --load saved_models/train-graph_lam-4x2-01_24_17-2502/min_val_loss.ckpt \
   --val_steps_to_log 3 --graph 1level --save-eval-to-zarr-path state_predictions.zarr/
```

results in:
```bash
$> zarrdump state_predictions.zarr
<xarray.Dataset> Size: 123MB
Dimensions:                    (elapsed_forecast_duration: 3, start_time: 11,
                                state_feature: 2, x: 789, y: 589)
Coordinates:
  * elapsed_forecast_duration  (elapsed_forecast_duration) timedelta64[ns] 24B ...
  * start_time                 (start_time) datetime64[ns] 88B 1990-09-07T06:...
  * state_feature              (state_feature) <U5 40B 'u100m' 'v100m'
    time                       (start_time, elapsed_forecast_duration) datetime64[ns] 264B dask.array<chunksize=(4, 3), meta=np.ndarray>
  * x                          (x) float64 6kB -1.999e+06 ... -2.925e+04
  * y                          (y) float64 5kB -6.095e+05 ... 8.605e+05
Data variables:
    state                      (start_time, elapsed_forecast_duration, state_feature, x, y) float32 123MB dask.array<chunksize=(4, 3, 2, 789, 589), meta=np.ndarray>
```

NB: This does not implement the inversion of the transformations that take place in `mllam-data-prep` (e.g. splitting individual features back into separate variables and levels. Also, the zarr datasets store time as `[start_time, elapsed_forecast_duration]` rather than `[start_time, sample]` to avoid producing a large array with many empty-values (NaNs) which would otherwise happen because each sample has a different start time. In the snippet below I have demonstrated how one could return to absolute time (probably there is a better way to do this...):

```python
import xarray as xr
import matplotlib.pyplot as plt
ds = xr.open_zarr("state_predictions.zarr/", chunks={})

ds.state.isel(x=0, y=0, start_time=slice(0, 4)).plot(hue="start_time", col="state_feature")
plt.savefig("state_predictions_relative_time.png")

ds_abs_time = xr.concat([
    ds.isel(start_time=i).swap_dims(dict(elapsed_forecast_duration="time")) for i in range(len(ds.start_time))
], dim="sample")
ds_abs_time.state.isel(x=0, y=0).plot(hue="sample", col="state_feature")
plt.savefig("state_predictions_absolute_time.png")
```

Example plot with time-axis showing elapsed time:
state_predictions_relative_time

Example plot with time-axis showing absolute time:
state_predictions_absolute_time

This probably needs more work, but I think it is ready for people to try it out and let me know what they think 😄 

## Issue Link

Implements https://github.com/mllam/neural-lam/issues/89

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


SimonKamuk:
I haven't tested the code myself yet, but this looks great! One question, is test_step called on all gpus in an multi-gpu setup? If so, i guess multiple gpus writing to the same zarr is not an issue as long as the chunks are different, but the metadata is only written correctly, if the GPU which handles batch_idx 0 finishes first, right?

leifdenby:
> ! One question, is test_step called on all gpus in an multi-gpu setup? If so, i guess multiple gpus writing to the same zarr is not an issue as long as the chunks are different, but the metadata is only written correctly, if the GPU which handles batch_idx 0 finishes first, right?

That is a good point. I haven't actually tested writing on multiple with interference run across multiple GPUs in parallel. There could be a race condition with the fact that I create the zarr dataset (i.e. write the meta info with `ds.to_zarr(path, mode="w")` only for `batch_idx==0` and for other `batch_idx` I append with `ds.to_zarr(path, mode="a")` https://github.com/mllam/neural-lam/pull/104/files#diff-043440a2d7a2cd62bb349e74c9fe4f55e69a8d0e4801f6ab2e150d381b11c74dR429. So if `batch_idx != 0` is ready before `batch_idx == 0` then I think this could might cause an exception to be raised (since there will be no dataset to append to). I am actually not sure how to do this correctly. @observingClouds do you know? I started by looking into the `xr.to_zarr(.., range=...)` `range` argument, but for that I think you also have to write the zarr meta data first.

Could we get away with issuing a warning to say this has only been tested for single GPU inference so far?

observingClouds:
@leifdenby do we know all available output times at the time of starting the write process? If so we can create the metadata first and chunks can be written independently without conflicting.

leifdenby:
> do we know all available output times at the time of starting the write process

Yes, we should be able to get these from the datastore. So you would go with using the `range` argument then? Also, does that mean that we'd create the metadata on say the rank `0` process first and then each separate process would write its own ranges? That would mean introducing something like an MPI barrier after the metadata write, right @observingClouds?

joeloskarsson:
> does that mean that we'd create the metadata on say the rank 0 process first and then each separate process would write its own ranges

I think this sounds like a good idea. You could use something like https://pytorch.org/docs/stable/distributed.html#torch.distributed.barrier for that. 

Am interested in having this work with multi-gpu, so I don't think we should just issue a warning and ignore the multi-gpu case. Something that might complicate things (if we allow batch_size > 1) is the note from https://github.com/joeloskarsson/neural-lam-dev?tab=readme-ov-file#evaluate-models.

> Note: While it is technically possible to use multiple GPUs for running evaluation, this is strongly discouraged. If using multiple devices the DistributedSampler will replicate some samples to make sure all devices have the same batch size, meaning that evaluation metrics will be unreliable. A possible workaround is to just use batch size 1 during evaluation. This issue stems from PyTorch Lightning. See for example https://github.com/Lightning-AI/torchmetrics/pull/1886 for more discussion.

 If samples are duplicated you could end up with different processes writing to the same region. So that is something to think about.

leifdenby:
> Am interested in having this work with multi-gpu, so I don't think we should just issue a warning and ignore the multi-gpu case. Something that might complicate things (if we allow batch_size > 1) is the note from https://github.com/joeloskarsson/neural-lam-dev?tab=readme-ov-file#evaluate-models.

Ok, things are never as easy as they seem 😆 

What I am proposing is that we merge this single-GPU implementation (so that people can start using it) and add multi-GPU in a later PR when we have figured out how to do that. If we did that I would issue a warning with just the single-GPU implementation

sadamov:
I have implemented a version that works with multi-gpu, using region="auto" after writing the full metadata initially. you can find the code here: https://github.com/sadamov/neural-lam/tree/write_zarr 

And here is the output of a `--eval test` on the danra test datastore:
image

Shortcomings:
- The chunking breaks down for large datastores and need some work (manual chunking?) - could be related to the reindexing step
- the logging is too verbose and not helpful

Thought this might be useful here, or in the follow-up PR.

neural_lam/models/ar_model.py
```diff
@@ -368,15 +370,79 @@ def on_validation_epoch_end(self):
...
+            )
+            # Unstack grid coords if necessary, this also avoids the need to
+            # try to store a MultiIndex zarr dataset which is not supported by
+            # xarray
+            if isinstance(self._datastore, BaseRegularGridDatastore):
+                da_pred = self._datastore.unstack_grid_coords(da_pred)
+
+            t0 = da_pred.coords["time"].values[0]
+            da_pred.coords["start_time"] = t0
+            da_pred.coords["elapsed_forecast_duration"] = da_pred.time - t0
```
joeloskarsson, line 418: Please have another look over how target times are used here. batch_times being fed in here are not [analysis_time, analysis_time+time_step, analysis_time+2time_step, ...], but rather [analysis_time+time_step, analysis_time+2time_step, ...]. This also because batch_predictions does not include the state at the analysis time.

This means that start_time is currently not when the forecast was started.

neural_lam/models/ar_model.py
```diff
@@ -368,15 +370,79 @@ def on_validation_epoch_end(self):
...
+            the number of prediction steps, num_grid_nodes is the number of
+            grid nodes, and d_f is the number of state features.
+        batch_idx : int
+            The index of the batch in the current epoch.
+        """
+        batch_size = batch_predictions.shape[0]
+        # Convert predictions to DataArray using _create_dataarray_from_tensor
+        das_pred = []
+        for i in range(len(batch_times)):
+            da_pred = self._create_dataarray_from_tensor(
```
joeloskarsson, line 404: Now that _create_dataarray_from_tensor is also used here it seems unreasonable to not properly deal with the hack where a WeatherDataset is instantiatied at each call (https://github.com/mllam/neural-lam/blob/1c281a2fe3f681154fb0ea883bf8142e10b9bb9d/neural_lam/models/ar_model.py#L182-L185). As is we are doing instatiation of O(NT) WeatherDatasets when saving to zarr (not a memory problem, as we throw them away, but very wastefull). This will quickly become a problem, as the WeatherDatasets will grow when we merge in more boundar-related changes.

neural_lam/models/ar_model.py
```diff
@@ -368,15 +370,79 @@ def on_validation_epoch_end(self):
...
+
+        da_pred_batch = xr.concat(das_pred, dim="start_time")
+
+        # Apply chunking along analysis_time so that each batch is saved as a
+        # separate chunk
+        da_pred_batch = da_pred_batch.chunk({"start_time": batch_size})
+
+        if batch_idx == 0:
+            logger.info(f"Saving predictions to {zarr_output_path}")
+            da_pred_batch.to_zarr(zarr_output_path, mode="w", consolidated=True)
```
joeloskarsson, line 431: I am getting erroneous start_times with this, due to problems with their encoding. This seems to fix it for me
 ```suggestion
            da_pred_batch.to_zarr(
                zarr_output_path,
                mode="w",
                consolidated=True,
                encoding={
                    "start_time": {
                        "units": "Seconds since 1970-01-01 00:00:00",
                        "dtype": "int64",
                    },
                },
            )
```
, setting it to use unix standard time. But I don't know if that is the best solution.

File: neural_lam/models/ar_model.py (hunk 2 above)
Comment (joeloskarsson) [reply neural_lam/models/ar_model.py:404]: This is an alternative hack to avoid constantly making new datasets: https://github.com/joeloskarsson/neural-lam-dev/pull/9/commits/d277b08b105ad3b8513311387c243557c94f123c however, this is still very much a hack and the TODO remains that we should handle this some proper way.

File: neural_lam/models/ar_model.py [hunk 4]
```diff
@@ -368,15 +370,79 @@ def on_validation_epoch_end(self):
...
+            # Unstack grid coords if necessary, this also avoids the need to
+            # try to store a MultiIndex zarr dataset which is not supported by
+            # xarray
+            if isinstance(self._datastore, BaseRegularGridDatastore):
+                da_pred = self._datastore.unstack_grid_coords(da_pred)
+
+            t0 = da_pred.coords["time"].values[0]
+            da_pred.coords["start_time"] = t0
+            da_pred.coords["elapsed_forecast_duration"] = da_pred.time - t0
+            da_pred = da_pred.swap_dims({"time": "elapsed_forecast_duration"})
```
joeloskarsson, line 419: Do we really want to leave the time coordinate in this DataArray? It looks to me like those values will only be valid for one of the forecasts (and I am not entirely sure which one).

neural_lam/models/ar_model.py
joeloskarsson: Just adjusting t0 would fix this: https://github.com/joeloskarsson/neural-lam-dev/commit/721ac5ec23ec19a1402946074194043072088c26 However, you probably need to get the step length from somewhere else than in that commit, as self.step_length is not present on main.

neural_lam/models/ar_model.py
```diff
@@ -432,6 +498,14 @@ def test_step(self, batch, batch_idx):
         self.spatial_loss_maps.append(log_spatial_losses)
         # (B, N_log, num_grid_nodes)
 
+        if self.args.save_eval_to_zarr_path:
+            self._save_predictions_to_zarr(
+                batch_times=batch_times,
+                batch_predictions=prediction,
```
joeloskarsson, line 504: These predictions are in the standardized scale. At some point before these are written to disk in the zarr they should be rescaled to the original data scale.

neural_lam/models/ar_model.py
observingClouds: @joeloskarsson which zarr version are you using?

neural_lam/models/ar_model.py
joeloskarsson: Can be done as https://github.com/joeloskarsson/neural-lam-dev/pull/9/commits/d3f636e9ec573b063ed0fcd236f572c5dfe0a5b7

neural_lam/models/ar_model.py
```diff
@@ -372,15 +375,109 @@ def on_validation_epoch_end(self):
...
+                    ds_pred_batch, idxnames=[idx_name]
+                )
+
+        if batch_idx == 0:
+            logger.info(f"Saving predictions to {zarr_output_path}")
+            ds_pred_batch.to_zarr(zarr_output_path, mode="w", consolidated=True)
+        else:
+            ds_pred_batch.to_zarr(
+                zarr_output_path, mode="a", append_dim="start_time"
+            )
```
observingClouds, line 470: we might need to be a bit careful with the time encoding here. In my research branch I get errors like:
```
UserWarning: Times can't be serialized faithfully to int64 with requested units 'days since 2020-02-12T01:20:00'. Serializing with units 'hours since 2020-02-12T01:20:00' instead. Set encoding['dtype'] to floating point dtype to serialize with units 'days since 2020-02-12T01:20:00'. Set encoding['units'] to 'hours since 2020-02-12T01:20:00' to silence this warning.
```
Basically, the first time step is saved as days, because it might be 00:00UTC, but consecutive steps might be in e.g. hours.

Jayant-kernel:
I opened a clean follow-up PR that addresses all review items from #104:

- Multi-GPU safe: rank-0 writes metadata template (compute=False), then barrier(), then all ranks write via region="auto"
- Correct start_time: analysis_time = batch_times[:,0] - step_length
- Rescale predictions to original scale in test_step before writing
- Drop raw time coord; explicit int64 encoding for time variables
- Cache WeatherDataset per (split, category) to avoid O(N*T) re-instantiation
- Add --save-eval-to-zarr-path CLI flag
- Add tests/test_zarr_eval.py single-GPU integration test

New PR: <https://github.com/mllam/neural-lam/pull/289>

joeloskarsson:
@leifdenby please answer in #289 if you want to continue working on this or if we should close this and move efforts to #289.

sadamov:
@leifdenby @observingClouds is this superceded by #289?

PR #106, opened 2025-01-26
Fix type error when calling `create_dataarray_from_tensor`

observingClouds:
## Describe your changes

The `create_dataarray_from_tensor` is now called with the expected type (tensor instead of numpy array).

## Issue Link

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


joeloskarsson:
Thanks for fixing this! I wonder why this was not caught by any tests? Maybe something to try to also change here to avoid similar bugs in the future. We have a test that should run an epoch or so of training on the example data. I wonder if we have managed to configure that one so that it will not run validation. Because this bug should have crashed the validation step I think, making any test including that fail.

leifdenby:
> I wonder why this was not caught by any tests?

I think because the tests don't run through the cli interface, but instead create a `Trainer` object directly. The cli interface code needs refactoring so that it can be called directly when a list of arguments to be treated as `argv` from the command line. I think the tests should call the cli interface though, that would be better.

observingClouds:
My idea was also to use `mypy` typing checks which could be able to detect that too.

leifdenby:
> My idea was also to use `mypy` typing checks which could be able to detect that too.

Interesting! I hadn't thought of that :) If you have time and inclination I would review a PR on that. But maybe not a priority for now? What do you think?

joeloskarsson:
Another simple temporary fix to the tests for this is to just specify options to the instantiated `pl.Trainer` so that it actually runs one validation step.

observingClouds:
> Interesting! I hadn't thought of that :) If you have time and inclination I would review a PR on that. But maybe not a priority for now? What do you think?

I have included mypy now and fixed all the issues that were raised (~40). It is not in strict mode yet as many functions are not yet typed in neural-lam and since this results in about 460 issues, I will not attempt that now 😄

sadamov:
Whenever we talk about static typing and/or CLI I like to mention this library that elegantly combines both: https://typer.tiangolo.com/

observingClouds:
To keep this PR rather clean, I reverted the mypy test and additional type annotations now.

leifdenby (approved):
Perfect! Thanks for fixing this.

PR #107, opened 2025-01-27
Add Learning Rate Scheduler to neurallamConfig

### LINKED ISSUE #97: Learning rate scheduler
Learning rate scheduling is something useful that we should have as an option. As setting up a way to configure arbitrary pytorch schedulers becomes over-complicated, we should start by just adding in options for cosine schedule. @ErikLarssonDev has a prototype for this.

matschreiner:
## Describe your changes

< Summary of the changes.>

I have added an optimization field to the `TrainingConfig` in `neural_lam/config.py`. This field takes a `lr`, `lr_scheduler` and `lr_scheduler_kwargs`. 

The `lr_scheduler` is checked against available schedulers in `pytorch` after the field is initialized. If the scheduler doesn't exist it will throw an `InvalidConfigError`.

The optimization config is added as a member to the `ARModel class`, so we can access it in `configure_optimizers`.
I think it would have been better to add the entire config, but maybe its a conscious choice that the config is not an attribute of the model. 

I have included tests to check if the configuration instantiation fails with an invalid scheduler and a valid scheduler. 
I also test if the learning rate is decreasing as expected on the optimizer when calling the learning_rate scheduler.

It is a breaking change because I have removed the `--lr` option in the arguments parser and moved it to the configuration file. 

If for example we wanted to add a `CosineAnnealingLR` with `T_max 1000` and `eta_min =` we would add the following to the config yaml.

```yaml
datastore:
     ...
training:
  state_feature_weighting:
    ...
  optimization:
    lr_scheduler: CosineAnnealingLR
    lr_scheduler_kwargs:
      T_max: 1000
      eta_min: 0.0
```

Also update some tests to use a fixture for model_args

< Please also include relevant motivation and context. >
I was tasked with exploring different schedulers and learning rates and thought it would be nice to log these things in the configuration files. 

< List any dependencies that are required for this change. >
None

## Issue Link
https://github.com/mllam/neural-lam/issues/97

## Type of change

- [X] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)

- [X] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [X] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [X] I have updated the [README](README.MD) to cover introduced code changes
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).

## Checklist for reviewers

- [X] the code is readable
- [x] the code is well tested
- [X] the code is documented (including return types and parameters)
- [X] the code is easy to maintain

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/neural_lam/config.py
+++ b/neural_lam/config.py
@@ -5,7 +5,6 @@ from typing import Dict, Union
 
 # Third-party
 import dataclass_wizard
-import torch
 
 # Local
 from .datastore import (
@@ -58,14 +57,33 @@ class Optimization:
     """
 
     lr: float = 1e-3
-    scheduler: str = ""
-    scheduler_kwargs: Dict[str, Union[int, float]] = dataclasses.field(
+    lr_scheduler: str = ""
+    lr_scheduler_kwargs: Dict[str, Union[int, float]] = dataclasses.field(
         default_factory=dict
     )
 
     def __post_init__(self):
-        if self.scheduler:
-            assert self.scheduler in dir(torch.optim.lr_scheduler)
+        valid_schedulers = [
+            "ConstantLR",
+            "CosineAnnealingLR",
+            "CosineAnnealingWarmRestarts",
+            "CyclicLR",
+            "ExponentialLR",
+            "LRScheduler",
+            "LambdaLR",
+            "LinearLR",
+            "MultiStepLR",
+            "MultiplicativeLR",
+            "OneCycleLR",
+            "PolynomialLR",
+            "ReduceLROnPlateau",
+            "SequentialLR",
+            "StepLR",
+        ]
+        if self.lr_scheduler and self.lr_scheduler not in valid_schedulers:
+            raise InvalidConfigError(
+                f"Invalid learning rate scheduler: {self.lr_scheduler}"
+            )
 
 
 @dataclasses.dataclass
--- a/neural_lam/config.py
+++ b/neural_lam/config.py
@@ -5,6 +5,7 @@ from typing import Dict, Union
 
 # Third-party
 import dataclass_wizard
+import torch
 
 # Local
 from .datastore import (
@@ -61,6 +62,10 @@ class Optimization:
     scheduler_kwargs: Dict[str, Union[int, float]] = dataclasses.field(
         default_factory=dict
     )
+
+    def __post_init__(self):
+        if self.scheduler:
+            assert self.scheduler in dir(torch.optim.lr_scheduler)
 
 
 @dataclasses.dataclass
--- a/neural_lam/config.py
+++ b/neural_lam/config.py
@@ -50,14 +50,19 @@ class Optimization:
 
     Attributes
     ----------
-    scheduler : str
+    lr: float
+        The initial learning rate.
+    lr_scheduler : str
         The type of learning rate scheduler to use
-    scheduler_args : Dict[str, Union[int, float]]
+    lr_scheduler_scheduler_step_freq : int
+        The frequency at which to call the learning rate scheduler.
+    lr_scheduler_args : Dict[str, Union[int, float]]
         Arguments to pass to the learning rate scheduler.
     """
 
     lr: float = 1e-3
     lr_scheduler: str = ""
+    lr_scheduler_scheduler_step_freq: int = 1
     lr_scheduler_kwargs: Dict[str, Union[int, float]] = dataclasses.field(
         default_factory=dict
     )
--- a/neural_lam/config.py
+++ b/neural_lam/config.py
@@ -44,6 +44,26 @@ class DatastoreSelection:
 
 
 @dataclasses.dataclass
+class Optimization:
+    """
+    Configuration for the learning rate scheduler.
+
+    Attributes
+    ----------
+    scheduler : str
+        The type of learning rate scheduler to use
+    scheduler_args : Dict[str, Union[int, float]]
+        Arguments to pass to the learning rate scheduler.
+    """
+
+    lr: float = 1e-3
+    scheduler: str = ""
+    scheduler_kwargs: Dict[str, Union[int, float]] = dataclasses.field(
+        default_factory=dict
+    )
+
+
+@dataclasses.dataclass
 class ManualStateFeatureWeighting:
     """
     Configuration for weighting the state features in the loss function where
@@ -85,6 +105,7 @@ class TrainingConfig:
     state_feature_weighting: Union[
         ManualStateFeatureWeighting, UniformFeatureWeighting
     ] = dataclasses.field(default_factory=UniformFeatureWeighting)
+    optimization: Optimization = dataclasses.field(default_factory=Optimization)
 
 
 @dataclasses.dataclass
--- a/neural_lam/config.py
+++ b/neural_lam/config.py
@@ -87,7 +87,7 @@ class Optimization:
         ]
         if self.lr_scheduler and self.lr_scheduler not in valid_schedulers:
             raise InvalidConfigError(
-                f"Invalid learning rate scheduler: {self.lr_scheduler}"
+                f"Invalid learning rate scheduler: {self.lr_scheduler}."
             )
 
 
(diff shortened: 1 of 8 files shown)


matschreiner:
@khintz 
@leifdenby

leifdenby:
From a quick glance this is looking really great! Thanks for doing this @matschreiner! Since this is looking quite close to complete and is quite an isolated feature, do you want to propose this for the next release (`v0.4.0`, edit: I can see @joeloskarsson already proposed this for `v0.5.0` on #97 so maybe we keep this feature for that release)?

It looks like the linting is complaining right now. We use `pre-commit` for linting. You can install the development tools with `python -m pip install -e ".[dev]"` or `python -m pip install -e ".[dev]"`. And you can install a git hook so pre-commit is automatically run with `pre-commit install` :)

Maybe if you fit the linting issues and then I can give this a proper testing and review?

joeloskarsson:
This is really nice to have! A couple of thoughts:

1. It would probably be good to handle lr-scheduler parameters similar to https://github.com/mllam/neural-lam/blob/807e47639dd7022a889a8195c9a2a4306c7ea37b/neural_lam/models/ar_model.py#L755-L757 , to have some control if they should be loaded from a previous checkpoint or not.
2. With this the lr is configured in the config script rather than with argparse flags. I think there is a larger discussion to be had about what should be in the config vs command line arguments. To me it is a bit unintuitive why e.g. lr should be somewhere else than the number of epochs I train for, or the ar_steps. I would always adjust these together, meaning I would have to change the config file and the bash-script I use for starting training runs. Up until now the setup in practice has been that you only write the config once, to tell neural-lam where to read data and how to weight variables, and then the actual training setup being ran is specified using command line args. There is definitely an argument to be made that all/most things should be in a config file. For now I just want us to have a thought-through idea of what goes where.

matschreiner:
@joeloskarsson 

Thanks for having a look at my code :)

1) I've changed it now so that we can restore lr_schedulers same way as we can restore the optimizer :) 

2) I agree that training time and learning rate scheduler setup is related, and I think that the issue is partly in the way pytorch is set up. But since the model is responsible for instantiating optimizers and schedulers, I think it makes sense to also include them in the model configuration file. The number of epochs/training steps on the other hand are arguments for the trainer so I'd say they belong in the arguments for the training script.

matschreiner:
Also I can't pass the tests because of timeout errors.  Maybe it could be fixed by merging in  [Leifs PR](https://github.com/mllam/neural-lam/pull/110).

joeloskarsson:
Yea the way these things are handled in pytorch and lightning really is not optimal (from a software design standpoint). I'm not sure if I have any idea for a good solution. Just wanted to add that I don't think it is suitable to think of the config file as a model config, as the model architecture is mainly defined through command line arguments. I think of it more as a config of how the model should read and weight data at the moment, but maybe that is something that should actually change (what the config is for).

sadamov:
I realized today that restoring checkpoints can be a bit tricky with LR schedulers. Currently, neural-lam allows to load a checkpoint with/without optimizer state. If the LR has a scheduler, this checkpoint logic has to be expanded as you and Joel already discussed above. Pytorch-Lightning has rather specific opinions about how a checkpoint is loaded. In the research branch we are always loading the full checkpoint and then resetting scheduler and optimizer manually. Setting epoch and step to 0 prevents pytorch-lightning from throwing warnings. I don't think my current hack is a good solution, but it should give some indication to what might be necessary: https://github.com/joeloskarsson/neural-lam-dev/blob/3cdbcc546f27fd843f5e3e86f4a512a67f7e2a61/neural_lam/models/ar_model.py#L806

observingClouds:
I recently needed to change the LR scheduler and thought "it would be really nice if one could provide this within a config", so thanks for having created this PR.

There is however one case that I think the current suggestion is not covering and that is the [torch.optim.lr_scheduler.ChainedScheduler](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ChainedScheduler.html), which is not uncommon to use.

I imagine e.g. the following config:

```
datastore:
     ...
training:
  state_feature_weighting:
    ...
  optimization:
    lr_scheduler: ChainedScheduler
    lr_scheduler_kwargs:
      schedulers:
          - scheduler: ConstantLR
            lr_scheduler_kwargs:
               factor:0.1
               total_iters:20
          - scheduler: CosineAnnealingLR
            lr_scheduler_kwargs:
              T_max: 1000
              eta_min: 0.0
```

Kirankhoushik:
Hi! I’m Kiran Khoushik, a CS student. I’m new to this repo and also new to LR schedulers.
Before I start implementing, could you please clarify how you want this to behave ?
1.Should the scheduler update every batch or every epoch ?
2.When resuming from a checkpoint, should the scheduler continue from the saved state or restart?
3.Should these settings live only in the YAML config, or do you also want CLI overrides ?

abhaygoudannavar:
Thanks @observingClouds for pointing me here from #237.

Looking at the current state of this PR — the main gap seems to be ChainedScheduler / SequentialLR support, where the config needs to handle a list of sub-schedulers with their own kwargs.

One approach that could work: rather than special-casing ChainedScheduler in the config parsing, we could recursively instantiate schedulers from a list. Something like:
``` yaml
training:
  optimization:
    lr_scheduler: SequentialLR
    lr_scheduler_kwargs:
      milestones: [1000]
      schedulers:
        - scheduler: LinearLR
          kwargs:
            start_factor: 0.001
            end_factor: 1.0
            total_iters: 1000
        - scheduler: CosineAnnealingLR
          kwargs:
            T_max: 50000
            eta_min: 1.0e-6
 ```
On the code side, the instantiation logic in 
configure_optimizers would check if any kwarg value is a list of dicts with a scheduler key, and recursively build those sub-schedulers before passing them to the parent.

I would be happy to take a stab at implementing this if the approach sounds reasonable.

**Comment by nsg365** [comment]
Hi @observingClouds , I'd like to take this up please

**Comment by joeloskarsson** [comment]
@matschreiner and @observingClouds is there interest from you to keep working on this? Otherwise I think it would be better to start with the solution from here, bring it up to date with main and then build this out in a new PR.

**Comment by observingClouds** [comment]
Hi @joeloskarsson,
I think it would be nice to keep discussing how we can support chained schedulers. I think a single scheduler itself is often not sufficient for a whole experiment. So including this very important case in the design for the beginning seems worth the wait for me.

**Comment by observingClouds** [comment]
How about always defining the schedulers in a list and wrap them on the code side into the [ChainedScheduler](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ChainedScheduler.html). I don't see any harm in always wrapping the scheduler into a ChainedScheduler, even if it is just a single scheduler.


```
datastore:
     ...
training:
  state_feature_weighting:
    ...
  optimization:
    lr_schedulers:
      - scheduler: ConstantLR
         factor:0.1
         total_iters:20
      - scheduler: CosineAnnealingLR
        T_max: 1000
        eta_min: 0.0
```

joeloskarsson:
Ok, makes sense @observingClouds 😄 
I think that is a nice and clean approach. Also don't see any issue with always wrapping in a ChainedScheduler.

sadamov:
Picking this back up. @matschreiner has been silent since 2025-02-04. The latest design direction (from @joeloskarsson and @observingClouds above) is: always wrap schedulers in `ChainedScheduler` so single-scheduler and multi-scheduler configs share one code path. I think that's the right design.

@observingClouds you've been the most invested. Do you want to take ownership? Either rebase this branch onto main with the ChainedScheduler design, or follow Joel's earlier suggestion and open a fresh PR with this as the reference.

@Kirankhoushik and @nsg365, you both volunteered earlier. If @observingClouds doesn't have bandwidth, are either of you still interested?

Closing #134 (GraphCast scheduler) separately since ChainedScheduler covers that case.

observingClouds:
Thanks for the ping/push @sadamov. I will implement the latest design.

PR #113, opened 2025-02-04
Feature: mypy testing

observingClouds:
## Describe your changes

This introduces mypy as a type checker and fixes some issues/inconsistencies that were found while implementing it.
This is only a start of type checking as many functions do not have type annotations yet, but this could help us to
extend the type checking more and more.

< Please also include relevant motivation and context. >
This was originally motivated by #106 where type-checking could have helped to identify the issue earlier. However, in that particular case it does not help because of other un-typed functions/missing type information of external packages.

< List any dependencies that are required for this change. >
- mypy as pre-commit hook

## Issue Link

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


joeloskarsson:
I think this could be a nice addition. As you say, we don't have type hints for major parts of the codebase still, but we could  implement this for the parts that have. And it would be useful when going over existing code and adding such type hints (I think doing that for all code is a bit much of a project for this PR). 

Do you want to resolve the conflicts and I could review this? If it is ready for that?

observingClouds:
With the CI being fixed, this PR is now ready to be reviewed @joeloskarsson. I agree that it would be good to start with the type testing somewhere and build from there.

neural_lam/datastore/base.py
```diff
@@ -285,6 +288,12 @@ def get_xy(self, category: str) -> np.ndarray:
         ----------
         category : str
             The category of the dataset (state/forcing/static).
+        stacked : bool
+            Whether to stack the x, y coordinates. The parameter `stacked` has
+            been introduced in this class. Parent class `BaseDatastore` has the
+            same method signature but without the `stacked` parameter. Defaults
+            to `True` to match the behaviour of `BaseDatastore.get_xy()` which
+            always returns the coordinates stacked.
```
joeloskarsson: Description seems wrong as this is in the BaseDatastore class
```suggestion
            Whether to stack the x, y coordinates.
```

neural_lam/datastore/base.py
```diff
@@ -276,7 +279,7 @@ def boundary_mask(self) -> xr.DataArray:
         pass
 
     @abc.abstractmethod
-    def get_xy(self, category: str) -> np.ndarray:
+    def get_xy(self, category: str, stacked: bool) -> np.ndarray:
```
joeloskarsson, line 282: Do we need to introduce `stacked` already here? I think the original idea was that the `BaseDatastore` does not have to assume that grid points are on a regular 2d grid (that is the assumption introduced in `BaseRegularGridDatastore`). Without that assumption the notion of stacked/not does not really make sense, i.e. we could not have `stacked=False` for an irregular grid.

observingClouds: The problem is that in `def get_xy_extent` we call `xy = self.get_xy(category, stacked=True)`.

observingClouds: The other issue is that adding new method arguments to a subclass function violates some design principles.

joeloskarsson: Ah, I see. Yea that is a problem. I also agree that it is not great that the subclass methods have an additional argument :thinking: It would be better if the "non-stacked" version of `get_xy` that the subclasses have would just be a different method. Anyhow, that's outside the scope of this PR, so you don't have to dive into that.

I suggest either:

1. In `get_xy_extent` change `xy = self.get_xy(category, stacked=True)` to just `xy = self.get_xy(category)`. And set the sub-classes to `stacked=True` as default argument.
2. Keep it as is, with the `stacked` argument present in `BaseDataStore.get_xy`, but add a comment that the `stacked=False` case does not make sense in the general case where grid points are not laid out in a regular 2d grid.

You can decide.

observingClouds: I opted for 2 and adjusted the docstring of both the parent and the sub-class.

neural_lam/datastore/base.py
```diff
@@ -472,19 +480,17 @@ def grid_shape_state(self) -> CartesianGridShape:
...
             The category of the dataset (state/forcing/static).
         stacked : bool
-            Whether to stack the x, y coordinates. The parameter `stacked` has
-            been introduced in this class. Parent class `BaseDatastore` has the
-            same method signature but without the `stacked` parameter. Defaults
-            to `True` to match the behaviour of `BaseDatastore.get_xy()` which
-            always returns the coordinates stacked.
+            Whether to stack the x, y coordinates. Defaults to `True` to match
+            the behaviour of `BaseDatastore.get_xy()` which always returns the
+            coordinates stacked.
```
joeloskarsson: It doesn't default to True anymore. That change makes sense, but we should just fix the docstring to be correct.

```suggestion
            Whether to stack the x, y coordinates.
```

PR #122, opened 2025-02-11
Standardize state diff stats in mdp datastore. Change variable name.

### LINKED ISSUE #112: Wrong scaling of 1-step diff_std, should be standardized
After having issues with my one variable having really high loss I realized that my issues stem from it having a very low value for diff_std in ar_model. This is the std-dev of 1-step differences. Digging a bit, I realize this is in mllam-data-prep computed for the non-standardized variables https://github.com/mllam/mllam-data-prep/blob/3a48c997144e66ef2478413668e669e124400cdd/mllam_data_prep/ops/statistics.py#L45 whereas it was earlier (and still for the MEPS datastore) for the standardized variables https://github.com/mllam/neural-lam/blob/f233f87c31ad7240644344f01e7cdb164044209e/neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py#L283 These differences should be converted to the standardized scale before being used in the loss. It is not necessarily wrong to compute them for the non-standard vars, they just have to be re-scaled. And we should be consistent about the scaling for these that are returned from the datastore.

Having the wrong scaling of these is actually a significant problem, and lines like https://github.com/mllam/neural-lam/blob/f233f87c31ad7240644344f01e7cdb164044209e/neural_lam/models/base_graph_model.py#L174 will make it harder for the model rather than help. So this is something that needs fixing. 

I don't have any strong opinions about where the standardization should happen: 1. when computing the stats and storing them on disk 2. when loading these in the datastore or 3. in the model itself. We just need to decide on some strategy to keep this clear and consistent.

SimonKamuk:
## Describe your changes

This merge request updates the `MDPDatastore` to return `state_diff_std_standardized` and `state_diff_mean_standardized` in addition to the existing non-standardized versions. The `NpyFilesDatastoreMEPS` already only returned the standardized versions, so here the variable name has just been changed.

Also updated where `get_standardization_dataarray` is called to get the standardized versions.

## Issue Link

Closes #112 

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/neural_lam/datastore/mdp.py
+++ b/neural_lam/datastore/mdp.py
@@ -301,7 +301,8 @@ class MDPDatastore(BaseRegularGridDatastore):
         should contain a `{category}_mean` and `{category}_std` variable for
         each variable in the category. For `category=="state"`, the dataarray
         should also contain a `state_diff_mean` and `state_diff_std` variable
-        for the one- step differences of the state variables.
+        for the one- step differences of the state variables along with their
+        standardized versions appended with `_standardized`.
 
         Parameters
         ----------
@@ -327,6 +328,19 @@ class MDPDatastore(BaseRegularGridDatastore):
             )
 
         ds_stats = self._ds[stats_variables.keys()].rename(stats_variables)
+
+        if category == "state":
+            standardized_state_diff = {
+                op: ds_stats[f"state_diff_{op}"] / ds_stats["state_std"]
+                for op in ops
+            }
+            ds_stats = ds_stats.assign(
+                **{
+                    f"state_diff_{op}_standardized": standardized_state_diff[op]
+                    for op in ops
+                }
+            )
+
         return ds_stats
 
     @cached_property
--- a/neural_lam/models/ar_model.py
+++ b/neural_lam/models/ar_model.py
@@ -65,10 +65,12 @@ class ARModel(pl.LightningModule):
                 da_state_stats.state_std.values, dtype=torch.float32
             ),
             "diff_mean": torch.tensor(
-                da_state_stats.state_diff_mean.values, dtype=torch.float32
+                da_state_stats.state_diff_mean_standardized.values,
+                dtype=torch.float32,
             ),
             "diff_std": torch.tensor(
-                da_state_stats.state_diff_std.values, dtype=torch.float32
+                da_state_stats.state_diff_std_standardized.values,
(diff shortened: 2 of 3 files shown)


SimonKamuk:
I think this is ready for review. One question I have is whether I should keep both the normalized and non-normalized state_diff stats (this is what I've done), or if I should just remove them entirely.

SimonKamuk:
Also, this goes in v0.4.0, right?

neural_lam/models/ar_model.py
```diff
@@ -65,10 +65,12 @@ def __init__(
                 da_state_stats.state_std.values, dtype=torch.float32
             ),
             "diff_mean": torch.tensor(
```
joeloskarsson, line 67: Just an additional comment, to be extra clear about this
```suggestion
            # Note that the one-step-diff stats (diff_mean and diff_std) are for differences 
            # computed on standardized data
            "diff_mean": torch.tensor(
```

neural_lam/datastore/mdp.py
```diff
@@ -321,12 +322,27 @@ def get_standardization_dataarray(self, category: str) -> xr.Dataset:
...
         if category == "state":
             stats_variables.update(
                 {f"state__{split}__diff_{op}": f"state_diff_{op}" for op in ops}
             )
 
         ds_stats = self._ds[stats_variables.keys()].rename(stats_variables)
+
+        # Add standardized state diff stats
+        if category == "state":
+            ds_stats = ds_stats.assign(
```
joeloskarsson, line 336: I think it is enough to return only the standardized, I don't think we need the non-standardized version. And if I understand correctly the mdp datastore returns both non-standardized and standardized versions, while the npyfilesmeps returns only standardized. That is an inconsistency I would like to avoid.

joeloskarsson (changes requested):
See small comments above (forgot to make them part of review)

neural_lam/datastore/mdp.py
SimonKamuk: Done, removed the non-standardized diff stats.

SimonKamuk:
Done 👍 care to have a final look and then I'll merge?

joeloskarsson (approved):
Perfect! lgtm

PR #160, opened 2025-10-28
Introduce `WeatherDatasetWithGraph` to return data and graph in training/inference samples

### LINKED ISSUE #70: Implement Model Sharding
### Description:
We should implement model sharding in neural-lam to allow for training with larger batch sizes without exhausting GPU vRAM. This feature will enable users to scale to larger models and improve training efficiency. Using high-res datasets with many input feature channels quickly exhausts even GPUs with 100GB vRAM. Currently, the batch-size must be reduced to 1-4 on many systems for such datasets.

### Proposed implementation:

1. Add sharding logic to model definition (see here for [bipartite_subgraph](https://github.com/ecmwf/anemoi-models/blob/86732c284f123d4d70d02aa857282f83bba04ba2/src/anemoi/models/distributed/khop_edges.py) & [here](https://github.com/ecmwf/anemoi-models/blob/86732c284f123d4d70d02aa857282f83bba04ba2/src/anemoi/models/distributed/graph.py)) 
2. Provide configuration options for sharding strategy as `train_model` flag

### Benefits:
- Train larger models with limited GPU resources
- Increase batch sizes for improved training + speed
- Enable future compatibility with Anemoi-Models (exchange models between frameworks)
- Allow training on high-res datasets with many features
- Allow training with large boundary areas

### Technical considerations:
- PyTorch's `DistributedDataParallel` should play well with `bipartite_graph`
- Ensure compatibility with existing model architectures (graph_lam and hi_lam)
- Add documentation and examples for using sharded models

This feature will significantly enhance neural-lam's capabilities for large-scale atmospheric modeling.

leifdenby:
## Describe your changes

This PR changes where the graph is loaded from disk, so that rather than being loaded within and belonging within the torch module of the forecasting model, instead *the graph is now contained as part of a training/inference batch*. Practically, this means that the PR makes the following changes:

- moves the loading of graphs from within `neural_lam.models.ar_model.ARModel` and into a new function in `neural_lam.graph_data` called `load_graph` that returns a new dataclass called `GraphEdgesAndFeatures`
- alters `neural_lam.interaction_net.InteractionNet` so that rather than taking adjacency lists as arguments when initialising an `Interactionet` (which make the edges of the graph fixed throughout training) the adjacency list should now be provided when calling `InteractionNet.forward(...)`. This means we can use different sets of edges (i.e. different graphs) during training
- introduces a dataclass called `GraphSizes` in `neural_lam.graph_data` which contains the different size-aspects of the graph which are required when creating the message-passing and embedding/de-embedding MLPs. A `GraphSizes` object can be derived from a `GraphEdgesAndFeatures` object, and this representation of the sizes is then passed in when creating the forecasting models `torch.nn.Module`.
- introduces new pytorch `Dataset` class called `neural_lam.weather_dataset.WeatherDatasetWithGraph` which is constructed from `neural_lam.weather_dataset.WeatherDataset` and a graph name. This pytorch dataset can then be passed to the data-loader and used during training/inference.

*So why did I do all of this?* This does seem like quite a significant change - I know!

There are two strong reasons for moving the graph itself (as in the edges it is made of) out of `ARModel` and instead treating it as part of the "data" to train on, specifically:

- it gives ability to train a model on different datasets (for example different geographical domains), i.e. this would be used in the case where we have different geographical domains but with the same physical variables and we want the model to learn how to do forecasting on both, and therefore learning message passing that is skillful for both geographical domains at the same time.
- enable work to use data-sharding already implemented in [pytorch-geometric](https://pytorch-geometric.readthedocs.io/en/latest/tutorial/distributed_pyg.html#graph-partitioning)

There are no dependency changes for this PR.

I have ensured that all tests execute, but I haven't completed any training yet. Would be very interested to hear thoughts by @matschreiner, @SimonKamuk and @observingClouds  on this! Marking this as draft for now until I get some feedback. Caveat: I did use openai codex for some of the refactoring to speed up my work, so I still need to go through carefully and check some of the docstrings before this is ready for proper review. I am looking forward feedback on the general approach first and foremost - thanks! 🥳 

## Issue Link

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


joeloskarsson:
In general I think this would be a great change, thanks for starting to look at this! :) 

A considerations for this is if we want to always pass _the same_ graph object as part of the batch (for now, when there is only one graph). Looking quickly there seems to now be a deepcopy of the graph in the code, but is that really necessary? The downside of this change is that we have to keep moving the (same) graph from CPU memory to GPU memory in each batch. Optimally we could just put the graph in GPU memory, and make sure that the batch always contains a reference to it. But not sure if that is possible without a too messy solution. It would also only work for the current case, with one static graph (although, I expect this to still be the main use case in the future). The moving from CPU to GPU should incur with some overhead, but maybe it's not too bad. Should also depend on the size of the graph.

neural_lam/weather_dataset.py
```diff
@@ -601,6 +605,68 @@ def _is_listlike(obj):
...
+        Dimensional metadata describing the graph.
+    graph_payload : dict
+        Dictionary view of ``graph_edges_and_features`` used when returning
+        batches.
+    """
+
+    def __init__(
+        self,
+        weather_dataset: WeatherDataset,
+        graph_name: str,
```
matschreiner, line 631: I think that the graph logic should be isolated in a GraphModule or GraphProvider where all the attributes defined below lives. 

How I envision this would be something like

```
class GraphModuleBase(abc.ABC):
@abstractmethod
  def get_graph(idx?): 
     ...
   
class WMGModule(GraphModuleBase):
  def __init__(self, graph_name):
         graph_edges_and_features = load_graph(
              graph_dir_path=self.graph_dir_path, device=self.device
          )
          self.graph_edges_and_features = graph_edges_and_features
          self.graph_sizes = build_graph_sizes(graph_edges_and_features)
          self.graph_payload = graph_edges_and_features.as_batch_dict()
          self.hierarchical = self.graph_sizes.hierarchical
          ...
  
  def get_graph():
         return copy.deepcopy(self.graph_payload)

class WeatherDatasetWithGraph(...):
    def __init__( self,
            weather_dataset: WeatherDataset,
            graph_module: GraphModuleBase,
            device: str = "cpu",
        ):
        self.graph_module = graph_module
        self.dataset = weather_dataset
        ...

    def __getitem__(self, idx):
        sample = self.weather_dataset[idx]
        graph = self.graph_provider.get_graph(idx)
        return (sample, graph)
```

Then we can have any implementation of the GraphBaseModule.

neural_lam/weather_dataset.py
```diff
@@ -601,6 +605,68 @@ def _is_listlike(obj):
...
+        self.graph_sizes = build_graph_sizes(graph_edges_and_features)
+        self.graph_payload = graph_edges_and_features.as_batch_dict()
+        self.hierarchical = self.graph_sizes.hierarchical
+
+    def __len__(self):
+        return len(self.weather_dataset)
+
+    def __getitem__(self, idx):
+        sample = self.weather_dataset[idx]
+        return (*sample, copy.deepcopy(self.graph_payload))
```
matschreiner: Maybe for another PR, but I think that in general it would be nicer to pass batches around as dictionaries

neural_lam/models/ar_model.py
```diff
@@ -277,7 +277,25 @@ def common_step(self, batch):
         num_grid_nodes, d_forcing),
             where index 0 corresponds to index 1 of init_states
         """
-        (init_states, target_states, forcing_features, batch_times) = batch
+        if len(batch) == 5:
```
matschreiner: I think that this conditional unpacking based on len(batch) is a little brittle and also coupled to a specific batch structure. If we use the batches as dictionaries the check would be easier and more robust.

matschreiner: Maybe it is a little late in the pipeline to figure out how the model should work on graph data (this is part of the forward pass somewhere in the inheritance tree, right?). Could this be decided at instantiation time instead?

neural_lam/weather_dataset.py
leifdenby: yes I agree. But yes, maybe we should leave this for a separate PR

neural_lam/weather_dataset.py
leifdenby: ok, thanks for this! Let's talk more about it offline, because I need to understand a bit more your thoughts :)

leifdenby:
> A considerations for this is if we want to always pass _the same_ graph object as part of the batch (for now, when there is only one graph). Looking quickly there seems to now be a deepcopy of the graph in the code, but is that really necessary?

I completely agree, this shouldn't be required, no. It might be that I could refactor this using @matschreiner's `GraphProvider` idea where one could have implementations like `StaticGraphProvider` and `DynamicGraphProvider` which internally take care of whether to allocate once on given device (gpu or cpu) or on each call for dynamic graphs. I will think a bit more about this. Glad you think this is going in the right direction @joeloskarsson !

neural_lam/models/ar_model.py
leifdenby: Yes, I agree, it would be better with a dict. I will change that and see what the others think to this too

leifdenby: yes, I will remove the option to have a batch that doesn't not include the graph (the 4-element tuple), so that the graph is always required. Actually, I will just make this a dict like you suggested above

neural_lam/train_model.py
```diff
@@ -248,6 +249,18 @@ def main(input_args=None):
...
+    ModelClass = MODELS[args.model]
+
+    # Prepare graph metadata for graph-based models
+    graph_sizes = None
+    graph_name = args.graph if issubclass(ModelClass, BaseGraphModel) else None
+    if graph_name is not None:
+        graph_dir_path = datastore.root_path / "graph" / graph_name
+        _, graph_sizes = WeatherDatasetWithGraph.load_graph(
+            graph_dir_path=graph_dir_path
+        )
```
observingClouds: This does not work as WeatherDatasetWithGraph has no load_graph function. The error passed our cli test as it is currently very limited. It should probably be similar to the code in `test_training.py`:

```suggestion
        graph_features_and_edges = load_graph(graph_dir_path)
        graph_sizes = build_graph_sizes(graph_features_and_edges)
```

neural_lam/models/base_graph_model.py
```diff
@@ -15,33 +19,42 @@ class BaseGraphModel(ARModel):
...
-            else:
-                setattr(self, name, attr_value)
-
-        # Specify dimensions of data
-        self.num_mesh_nodes, _ = self.get_num_mesh()
+        self.graph_sizes = graph_sizes
+        self.hierarchical = graph_sizes.hierarchical
+        self.current_graph: Union[dict[str, Any], None] = None
+
+        self.num_mesh_nodes = graph_sizes.num_mesh_nodes
```
observingClouds: I know you want rather conceptual suggestions, but let me post some issues I found.

We set `self.graph_sizes = graph_sizes` and the property `self.num_mesh_nodes` is nothing else than returning `self.graph_sizes.num_mesh_nodes`, we can therefore delete the line
```suggestion
```

Issue #161, opened 2025-10-28
LaTeX error when running neural_lam.train_model


lorenzo30salgado:
When execution the command: `pdm run python -m neural_lam.train_model --config_path data/config.yaml --model graph_lam --graph multiscale --epochs 2 --processor_layers 4 --ar_steps_train 6 --ar_steps_eval 6 --num_past_forcing_steps 0 --num_future_forcing_steps 0 --val_steps_to_log 1 3 6`, I’m having this error message:

```
RuntimeError: latex was not able to process the following string:
b'lp'

Here is the full command invocation and its output:

latex -interaction=nonstopmode --halt-on-error --output-directory=tmp3jwbe_rz 94709c666295ff3d987a4eac032e9551.tex

This is pdfTeX, Version 3.14159265-2.6-1.40.19 (TeX Live 2018) (preloaded format=latex)
 restricted \write18 enabled.
entering extended mode
(./94709c666295ff3d987a4eac032e9551.tex
LaTeX2e <2017-04-15>
Babel <3.17> and hyphenation patterns for 3 language(s) loaded.
(/usr/share/texlive/texmf-dist/tex/latex/base/article.cls
Document Class: article 2014/09/29 v1.4h Standard LaTeX document class
(/usr/share/texlive/texmf-dist/tex/latex/base/size10.clo))

! LaTeX Error: File `type1cm.sty' not found.

Type X to quit or <RETURN> to proceed,
or enter new name. (Default extension: sty)

Enter file name: 
! Emergency stop.
<read *> 
         
l.7 \usepackage
               {type1ec}^^M
No pages of output.
Transcript written on tmp3jwbe_rz/94709c666295ff3d987a4eac032e9551.log.
```

I was wondering if anyone has had the same problem? For context, I’m using an ERA5 subset of data for my training, and I followed the workflow described in the README.

sadamov:
So what's happening is that whenever a plotting routine is called during training, certain labels and titles are allowed to use latex. From Slack:
> When running neural_lam.train_model, I’m having this error message after the first epochs: RuntimeError: latex was not able to process the following string: b'lp'

After a few epochs, I assume the first time validation is triggered, some plots are generated and you run into latex dependency issues. 

Now we don't expect the user to have latex installed and have implemented this check: https://github.com/mllam/neural-lam/blob/579efc8050d0f51c04c6b369f34126ac29e41106/neural_lam/utils.py#L232

[...]

PR #162, opened 2025-10-30
Implement more robust LaTeX availability check function

lorenzo30salgado:
Added a function to check for LaTeX availability and its toolchain, enhancing visualization capabilities.

## Describe your changes

Added a more robust function to check if LaTeX availability and its toolchains.
 
Please let me know if any changes are needed :)

## Issue Link

this makes reference to the issue #161.

## Type of change

-  [X]  🐛 Bug fix (non-breaking change that fixes an issue)

- [X] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [X] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).

## Checklist for reviewers

- [X] the code is readable
- [X] the code is well tested
- [X] the code is documented (including return types and parameters)
- [X] the code is easy to maintain

## Checklist for assignee

- [x] the tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix the order in create_graph.py which caused wrong G2M and M2G
 [\#150](https://github.com/mllam/neural-lam/pull/150) @YUTAIPAN
 
+- Adding a more robust LaTeX availability check function [\#162](https://github.com/mllam/neural-lam/pull/162) @lorenzo30salgado
+
 
 ### Maintenance
 


neural_lam/utils.py
```diff
@@ -221,15 +221,51 @@ def make_mlp(blueprint, layer_norm=True):
     return nn.Sequential(*layers)
 
 
+def has_working_latex():
```
joeloskarsson, line 229: This is a great check! As this is a bit heavier to run now, and since it should always return the same thing during the course of the program running, I think we should cache the output of this function.

```suggestion
@cache
def has_working_latex():
```

(also requires importing [`functools.cache`](https://docs.python.org/dev/library/functools.html#functools.cache))

neural_lam/utils.py
```diff
@@ -221,15 +221,51 @@ def make_mlp(blueprint, layer_norm=True):
...
+$E=mc^2$ \LaTeX\ ok
+\end{document}
+""".lstrip()
+
+    try:
+        with tempfile.TemporaryDirectory() as td:
+            td = Path(td)
+            (td / "test.tex").write_text(tex_src, encoding="utf-8")
+            cmd = ["latex", "-interaction=nonstopmode", "-halt-on-error", "test.tex"]
+            subprocess.run(cmd, cwd=td, stdout=subprocess.PIPE, stderr=subprocess.PIPE, timeout=20, check=True)
```
sadamov: Did you check that 20s timeout is enough? The first latex run can be really slow on some machines. How much time did the subprocess roughly require when you tested it?

sadamov (commented):
thanks @lorenzo30salgado, this should really cover many more latex installations a user could have. 🙏 
The linting is currently failing in the checks below, could you run `pre-commit run --all-files` (see [here](https://github.com/mllam/neural-lam?tab=readme-ov-file#development-and-contributing)) in your branch and push?

sadamov:
> Introduce an option (e.g. --tex ?) to enable tex-based plotitng, but have it default off. That way any user that wants to use tex-based plotting have to explicitly make sure that they have an environment where that works.

Copied this comment over from #161. I would say we merge this bugfix now. If we think the `--tex` flag is still required at a later stage we can introduce it in a separate PR and discuss whether it should be off/on by default. ok?

neural_lam/utils.py
lorenzo30salgado: Thank you for your review! You’re right, it only needs to be checked once. I’m committing your suggestion.

neural_lam/utils.py
lorenzo30salgado: Thank you for your review! To be honest, I set the timeout to 20 seconds arbitrarily because I didn't want it to take any longer. However, you're right that if a user wants to use latex and it takes more than 20 seconds to check, they can expect to wait for longer than the current timeout. By having a flag --tex as you suggested #161, we can maybe also set the timeout manually.

neural_lam/utils.py
joeloskarsson: Great! I think you might have to import it as well (`from functools import cache` at the top of the file) for it to work. Please try it out.

lorenzo30salgado: You’re right. If you check my last commit I added that line.

sadamov (approved):
looks good now. please add an entry to the CHANGELOG and then I will assign the PR to myself and merge.

neural_lam/utils.py
sadamov: I tested this on my cluster, the check always took <<1s for both `false` or `true` outcome. So we can leave 20s for now

PR #178, opened 2025-12-28
[maint] run full test suite only with uv package manager

### LINKED ISSUE #179: Reduce CI test times
Currently, the CI runs the whole test suite in four different configurations:
- pip, cpu
- pip, gpu
- pdm, cpu
- pdm, gpu

These tests are quite expensive in time and resources. I therefore, propose to reduce the CI test times by:

- [x] using uv instead of pdm
- ~~[ ] run test with pip only when a PR is marked as ready (not in draft mode)~~
- [x] run full test suite only with uv and only test successful env creation with pip

I drafted a PR for the first point in #178 and the numbers are convincing:

Image

`pdm` spends as much time setting up the environment as the actual tests lasts. `uv` starts with the actual tests after 20s 🚀 (cpu) and 70s on  gpu. With smart caching this could probably be reduced even further.

observingClouds:
## Describe your changes

This PR enhances the CI test time by:
- using uv instead of pdm
- reduce testing with pip to only test resolution of environment
- update README install instructions for `uv` instead of `pdm`

## Issue Link

closes #179 

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,7 +8,7 @@
 
 ## Issue Link
 
-< Link to the relevant issue or task. > (e.g. `closes #00` or `solves #00`)
+< Link to the relevant issue or task, if applicable > (e.g. `closes #00` or `solves #00`)
 
 ## Type of change
 
@@ -43,10 +43,12 @@ Each PR comes with its own improvements and flaws. The reviewer should check the
   - *added*: when you have added new functionality
   - *changed*: when default behaviour of the code has been changed
   - *fixes*: when your contribution fixes a bug
+  - *maintenance*: when your contribution is relates to repo maintenance, e.g. CI/CD or documentation
 
 ## Checklist for assignee
 
 - [ ] PR is up to date with the base branch
 - [ ] the tests pass
-- [ ] author has added an entry to the changelog (and designated the change as *added*, *changed* or *fixed*)
+- [ ] (if the PR is not just maintenance/bugfix) the PR is assigned to the next milestone. If it is not, propose it for a future milestone.
+- [ ] author has added an entry to the changelog (and designated the change as *added*, *changed*, *fixed* or *maintenance*)
 - Once the PR is ready to be merged, squash commits and merge the PR.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Maintenance
 
-- Update CI/CD to use python 3.13 for testing and full range of current python versions for linting (3.10 - 3.14) [\#173](https://github.com/mllam/neural-lam/pull/173)
+- Update PR template to clarify milestone/roadmap requirement and maintenance changes [\#186](https://github.com/mllam/neural-lam/pull/186) @joeloskarsson
+
+- Update CI/CD to use python 3.13 for testing and full range of current python versions for linting (3.10 - 3.14) [\#173](https://github.com/mllam/neural-lam/pull/173) @observingClouds
 
 - Move development dependencies to dependency-group [#\174](https://github.com/mllam/neural-lam/pull/174) @observingClouds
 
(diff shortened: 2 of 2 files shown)


.github/workflows/install-and-test.yml
```diff
@@ -81,48 +66,62 @@ jobs:
...
-        run: |
-          TORCH_VERSION=$(pdm install --dry-run | grep -Eo 'torch [0-9]+\.[0-9]+\.[0-9]+' | awk '{print $2}')
-          echo "Torch version detected: $TORCH_VERSION"
-          echo "TORCH_VERSION=$TORCH_VERSION" >> $GITHUB_ENV
-
       - name: Install PyTorch
+        if : matrix.package_manager != 'uv'
         run: |
-          ${{ matrix.package_manager == 'pdm' && 'pdm run python -m pip' || 'python -m pip' }} install "torch==$TORCH_VERSION" --index-url ${{ matrix.torch_index }}
+          python -m pip install "torch==$TORCH_VERSION" --index-url ${{ matrix.torch_index }}
```
leifdenby, line 72: I generally think that removing the "pdm"-prefix that I had before is a good idea, it makes the steps easier to read. But there is place where I think you might still need it. Or at least:  as I am reading it now there is nothing in place to ensure that uv will install the expected torch variant because uv will default to the cpu variant by resolving pytorch from pyproject.toml. So I think you need to remove the `if matrix.package_manager != uv` part here, and prefix with `uv pip ...` so that we also get the right torch variant when using uv in ci/cd

leifdenby: oh, wait, no I see you've done this below instead.. ok

.github/workflows/install-and-test.yml
```diff
@@ -81,48 +66,62 @@ jobs:
...
+      - name: Setup environment with uv (if applicable)
+        if: matrix.package_manager == 'uv'
         run: |
-          ${{ matrix.package_manager == 'pdm' && 'pdm install --group :all' || 'python -m pip install --group dev .' }}
+          uv venv --no-project
+          source .venv/bin/activate
+          uv pip install torch --index-url ${{ matrix.torch_index }}
+          uv pip install --group dev -e .
+          uv pip list
+          echo "VIRTUAL_ENV=$PWD/.venv" >> $GITHUB_ENV
```
leifdenby: could you include a comment here about why this is required. I assume it is to ensure that uv keeps using the venv we just installed torch etc into

.github/workflows/install-and-test.yml
```diff
@@ -81,48 +66,62 @@ jobs:
...
+        if: matrix.package_manager == 'uv'
         uses: actions/cache/restore@v4
         with:
           path: tests/datastore_examples/npyfilesmeps/meps_example_reduced.tar.gz
           key: ${{ runner.os }}-meps-reduced-example-data-v0.3.0
           restore-keys: |
             ${{ runner.os }}-meps-reduced-example-data-v0.3.0
 
+      - name: Run smoke test
+        if: matrix.package_manager != 'uv'
```
leifdenby, line 113: we could do this with uv too, no? Right now it looks like you're saying, when installing with `pip` we only check we can import neural-lam, and only when installing with uv do we run the tests. That's ok with me if that's the plan (to save time), but maybe could you put a comment in the start of the github actino file just to make that is claer, that this is supposed to happen?

observingClouds: We could, but it takes about 10-20 seconds and we do this anyway in the full test with pytest, so I thought we could spare those seconds :)

.github/workflows/install-and-test.yml
observingClouds: actually, i think we might live without it, this is only important during creation of the environment, but the next cell should be sufficient

```suggestion
```

leifdenby:
Looking good @observingClouds :) I guess we could put this on the roadmap?

observingClouds:
Can you approve 😀 we can then merge when agreed

joeloskarsson:
I think it could be good to explicitly propose this in the roadmap first, before merging

leifdenby:
Since `v0.5.0` has been released should move this be proposed for `v0.6.0`?

leifdenby (approved):

leifdenby:
Since the next release we are working on is `v0.6.0` and this PR is on the roadmap now, and it has been approved, this PR can be merged in, right? :) You're the assignee @observingClouds 🚀

PR #184, opened 2026-01-21
Allow for graph handling of the research branch

zweihuehner:
## Describe Your Changes

This PR is part of the effort to align the main branch graph handling with the implementation used in the research branch.

To my understanding at a high level, the required changes fall into four categories:

1. Graph loading
2. Model-side handling of the graph
3. Graph visualization
4. Graph creation (coupled to the boundary datastore)

This PR only includes changes that are NOT coupled to the boundary datastore (1., 2. and 3.). 
Changes that are required to completely comply with the research branch but are coupled to the boundary datastore are listed at the bottom (4.). 
It should be noted that the graph visualization (3.) of the research branch also requires a coupling to the datastore. In this PR the existing graph visualization from the main branch has merely be adapted to comply with the new graph loading.

### 1. Graph Loading

`utils.py`

- Add load_graph() from the research_branch
  - Trigraphs and all related functionality are excluded.
  - Moved zero-indexing of m2g and g2m to dedicated functions.
  - Note: The name mesh_features.pt is kept to comply with the main branch.

- Extend BufferList:
  - Allow multiplication and division that are now required for the load_graph() implementation.

- Add zero_index_edge_index():
  - Ensures sender and receiver indices in edge_index start at 0
  - Previously handled in interaction_net.py, but now applied directly in load_graph()

### 2.  Model-side handling of the graph

`interaction_net.py`

- Remove above described zero-indexing logic that is now handled in load_graph() to avoid redundancy.

### 3.  Graph visualization

`plot_graph.py`

- Add functionality to revert the zero-indexing of the new graph loading for plotting purposes.

## Remaining changes to fully comply with the research branch
(Coupled to Boundary Datastore — Not included in this PR))

### 4. Graph Creation

`build_rectangular_graph.py`

- Replace create_graph.py with build_rectangular_graph.py
  - Can largely be copied from the research_branch
  - Change default --output-dir from "graphs" to "graph" for consistency
  - Coupling to the boundary datastore comes from using load_config_and_datastores() instead of load_config_and_datastore()

`base_graph_model.py`

- Scaling of mesh_static_features that was previously directly done in create_graph()
  - Having it in base_graph_model.py is probably not ideal, alternatives could be:
    - Scale features in build_rectangular_graph.py before saving (aligns with old approach)
    - Scale during graph loading (Would require passing the datastore to load_graph())

### 3. Graph Visualization

`plot_graph.py`

- The current plot_graph.py cannot correctly visualize graphs created with build_rectangular_graph.py because of the boundary datastore.
- Replace it with plot_graph.py from the research_branch
  - Rename "graphs" to "graph"
  - Add support for "mesh_lat_lon"
  - Coupling to the boundary datastore comes from using load_config_and_datastores() instead of load_config_and_datastore()

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased](https://github.com/mllam/neural-lam/compare/v0.5.0...HEAD)
 
-### Changed
+### Added
 
+- Enable `pin_memory` in DataLoaders when GPU is available for faster async CPU-to-GPU data transfers [\#236](https://github.com/mllam/neural-lam/pull/236) @abhaygoudannavar
+
+### Changed
 - Refactor graph loading: move zero-indexing out of the model and update plotting to prepare using the research-branch graph I/O [\#184](https://github.com/mllam/neural-lam/pull/184) @zweihuehner
 
+- Replace `print()`-based `rank_zero_print` with `loguru` `logger.info()` for structured log-level control ([#33](https://github.com/mllam/neural-lam/issues/33))
+
 ### Fixed
+
+- Replace `shell=True` subprocess call in `compute_standardization_stats.py` with a safe argument list and Python-side hostname parsing to prevent command injection via `SLURM_JOB_NODELIST` [\#264](https://github.com/mllam/neural-lam/pull/264) @ashum9
+- Avoid NaN when standardizing fields with zero std [#189](https://github.com/mllam/neural-lam/pull/189) @varunsiravuri
 
 - Fix README image paths to use absolute GitHub URLs so images display correctly on PyPI [\#188](https://github.com/mllam/neural-lam/pull/188) @bk-simon
 
@@ -19,8 +27,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed the hardcoded True to a conditional check "persistent_workers=self.num_workers > 0" [\#235](https://github.com/mllam/neural-lam/pull/235) @santhil-cyber
 
-- `fractional_plot_bundle` now correctly multiplies by fraction instead of dividing [
-\#222](https://github.com/mllam/neural-lam/pull/222) @santhil-cyber
+- `fractional_plot_bundle` now correctly multiplies by fraction instead of dividing
+[\#222](https://github.com/mllam/neural-lam/pull/222) @santhil-cyber
+
+### Added
+
+- Expose `--wandb_id` CLI argument to allow resuming an existing W&B run by
+  ID. When provided, `resume="allow"` is set automatically so the same job
+  script works for both the initial submission and all resubmissions, making
+  it suitable for HPC systems with limited job runtimes or that may crash.
+  [\#197](https://github.com/mllam/neural-lam/pull/197) @Mani212005
+
+
 
 - Fix Slack domain link [\#288](https://github.com/mllam/neural-lam/pull/288) @sadamov
 
@@ -39,6 +57,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update CI/CD to use only uv for full test suite and drop pdm [\#178](https://github.com/mllam/neural-lam/pull/178) @observingClouds
 
 - Fix caching of MEPS example data in CI/CD [\#181](https://github.com/mllam/neural-lam/pull/181) @observingClouds
+
+- Migrated build backend from PDM to Hatchling with hatch-vcs and added uv build in deploy CI
 
 ## [v0.5.0](https://github.com/mllam/neural-lam/releases/tag/v0.5.0)
 
--- a/neural_lam/models/base_graph_model.py
+++ b/neural_lam/models/base_graph_model.py
@@ -34,7 +34,7 @@ class BaseGraphModel(ARModel):
 
         # Specify dimensions of data
         self.num_mesh_nodes, _ = self.get_num_mesh()
-        utils.rank_zero_print(
+        utils.log_on_rank_zero(
             f"Loaded graph with {self.num_grid_nodes + self.num_mesh_nodes} "
             f"nodes ({self.num_grid_nodes} grid, {self.num_mesh_nodes} mesh)"
         )
--- a/neural_lam/utils.py
+++ b/neural_lam/utils.py
@@ -10,6 +10,7 @@ from pathlib import Path
 # Third-party
 import pytorch_lightning as pl
 import torch
+from loguru import logger
 from pytorch_lightning.loggers import MLFlowLogger, WandbLogger
 from pytorch_lightning.utilities import rank_zero_only
 from torch import nn
@@ -441,9 +442,19 @@ def fractional_plot_bundle(fraction):
 
 
 @rank_zero_only
-def rank_zero_print(*args, **kwargs):
-    """Print only from rank 0 process"""
-    print(*args, **kwargs)
+def log_on_rank_zero(msg: str, level: str = "info", *args, **kwargs):
+    """Log a message only on rank zero using loguru logger.
+
+    Parameters
+    ----------
+    msg : str
+        The message to log.
+    level : str, optional
+        The logging level (e.g. "info", "warning", "error"). Default is "info".
+    """
+    if rank_zero_only.rank == 0:
+        log_fn = getattr(logger, level, logger.info)
+        log_fn(msg, *args, **kwargs)
 
 
 def init_training_logger_metrics(training_logger, val_steps):
@@ -466,7 +477,7 @@ def init_training_logger_metrics(training_logger, val_steps):
 
 @rank_zero_only
 def setup_training_logger(datastore, args, run_name):
-    """
+    """Set up the training logger (WandB or MLFlow).
 
     Parameters
     ----------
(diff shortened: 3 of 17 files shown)


neural_lam/interaction_net.py
```diff
@@ -52,13 +53,18 @@ def __init__(
...
             hidden_dim = input_dim
 
-        # Make both sender and receiver indices of edge_index start at 0
-        edge_index = edge_index - edge_index.min(dim=1, keepdim=True)[0]
-        # Store number of receiver nodes according to edge_index
         self.num_rec = edge_index[1].max() + 1
-        edge_index[0] = (
-            edge_index[0] + self.num_rec
-        )  # Make sender indices after rec
+        # edge_index is expected to be zero-based and local:
```
zweihuehner, line 57: if to detailed here is an alternative:

"edge_index is zero-based and local; sender indices are offset by num_rec
to obtain global node indices (receivers first, then senders)."

neural_lam/models/base_graph_model.py
```diff
@@ -26,6 +26,13 @@ def __init__(self, args, config: NeuralLAMConfig, datastore: BaseDatastore):
             graph_dir_path=graph_dir_path
         )
         for name, attr_value in graph_ldict.items():
+            # NOTE: It would be good to rescale mesh node position features in
```
zweihuehner: I think this part can eventually savely be moved to build_rectangular_graph.py, which keeps base_graph_model a bit less "hardcoded"

neural_lam/interaction_net.py
joeloskarsson: I think detailed comments are good here

neural_lam/interaction_net.py
```diff
@@ -52,13 +53,18 @@ def __init__(
...
+        #   edge_index[0]: sender indices in [0 .. num_snd-1]
+        #   edge_index[1]: receiver indices in [0 .. num_rec-1]
+        # Global node indexing is defined as:
+        #   receivers → [0 .. num_rec-1]
+        #   senders   → [num_rec .. num_rec+num_snd-1]
+        # Hence, sender indices are offset by num_rec to obtain global indices.
```
joeloskarsson: The terminology "global" confused me, here is a suggestion:
```suggestion
        # The edge indices used in this GNN layer are defined as:
        #   receivers → [0 .. num_rec-1]
        #   senders   → [num_rec .. num_rec+num_snd-1]
        # Hence, sender indices from the input edge_index are offset by num_rec to obtain the indices used in this layer..
```

neural_lam/models/base_graph_model.py
joeloskarsson: Yes I agree this is not the best way to handle this. Actually, I even know that this approach can cause problems, when the coordinates in the datastore do not match with those used for static features in `mesh_static_features`.

joeloskarsson: For now the `mesh_static_features` are already rescaled in `create_graph.py`, so we should not add in this yet. What I propose is to instead add in this rescaling later, when switching to `build_rectangular_graph.py`, and then do it in a better place than here.

neural_lam/utils.py
```diff
@@ -82,7 +100,7 @@ def load_graph(graph_dir_path, device="cpu"):
         - mesh_down_edge_index
         - g2m_features
         - m2g_features
-        - m2m_features
+        - m2m_node_features
         - mesh_up_features
         - mesh_down_features
         - mesh_static_features
```
joeloskarsson, line 211: Could you double-check that the names here (and used further down in the function) match those in `create_graph.py`, so this is compatible?
In particular, what used to be called `m2m_features` are a set of edge features associated with the intra-level graph edges. On the research branch we had renamed `mesh_static_features` to `m2m_node_features` as that is more descriptive (these are static features associated with the nodes).

neural_lam/utils.py
```diff
@@ -96,13 +114,74 @@ def loads_file(fn):
             weights_only=True,
         )
 
+    # Need to reindex some edge index to start from 0
+    reindex_func = zero_index_edge_index
```
joeloskarsson: Maybe we can drop this until triangular graphs are introduced? And just use `zero_index_edge_index` directly below.

neural_lam/utils.py
```diff
@@ -112,22 +191,21 @@ def loads_file(fn):
...
-    )  # Col. 0 is length
-    m2m_features = BufferList(
-        [level_features / longest_edge for level_features in m2m_features],
-        persistent=False,
-    )
+    if three_dim_features:
+        # For trigraphs the edge features are already rescaled when created
+        # Set to 1. to not rescale here
+        longest_edge = 1.0
+    else:
```
joeloskarsson: At the moment we don't have this implemented, so could leave out this check.

joeloskarsson: And then we don't need the to define the `three_dim_features` boolean

neural_lam/utils.py
zweihuehner: As I understand it, the difference is mainly a naming change at the graph-creation stage. On the main branch, what is called `m2m_features` is on the research branch referred to as `m2m_node_features`. In both cases, these features are mapped to `mesh_static_features` during graph loading, and the models consistently operate on `mesh_static_features` afterward.

I am a bit confused by the statement “on the research branch we had renamed `mesh_static_features `to `m2m_node_features`”, as it seems that `mesh_static_features` remains the model-level name in both branches, and only the pre-graph loading input (`m2m_features `→ `m2m_node_features`) was renamed.

Please let me know if I am missing something here, but I assume we are talking about the same thing.

joeloskarsson: Oh I see, sorry I misread this and thought it was about the saved files, not the return values. 

As the code is now, this all looks like it should work. But maybe just to clarify what I meant (which is really about the saved files, and not about the return values of this function): 
`m2m_features` is not node features. That is a file/tensor with edge features, that matches the `m2m_edge_index`. The file `m2m_node_features` is what is mapped to `mesh_static_features` on the research branch (https://github.com/joeloskarsson/neural-lam-dev/blob/51ced3627b7a5cb89f07213f30c43a4fbb5766a0/neural_lam/utils.py#L123-L125).

neural_lam/utils.py
```diff
@@ -43,6 +43,93 @@ def __len__(self):
...
+):
+    """
+    Zero-index the m2g (mesh-to-grid) edge index, or reverse the operation.
+
+    This needs special handling as not all mesh nodes might be indexed,
+    so we can't simply subtract the minimum. If the mesh indices come first
+
+    """
+
+    sign = 1 if reverse else -1
```
joeloskarsson: Could you add some explanation of what the `reverse` argument does here? From context I guess this is to inverse the reindexing operation?

joeloskarsson:
In general I think this is looking great, but I can try out also building some graphs and running the code (later).

neural_lam/utils.py
zweihuehner: Sure! I renamed to "restore" and added a full numpy docstring for both functions.

neural_lam/plot_graph.py
```diff
@@ -78,6 +78,16 @@ def main():
...
 
+    # The plotting requires the edges to be non-zero-indexed
+    # with grid indices following mesh indices
+    m2g_edge_index = utils.zero_index_m2g(
+        m2g_edge_index, [mesh_static_features], mesh_first=True, restore=True
+    )
+
+    g2m_edge_index = utils.zero_index_g2m(
+        g2m_edge_index, [mesh_static_features], mesh_first=True, restore=True
+    )
```
joeloskarsson, line 81: Testing this locally I ran into issues here. For a hierarchical graph `mesh_static_features` is a `BufferList`, but the functions expects a tensor.
Removing the list brackets seemed to fix that, but the hierarchical graph did not look correct when I then got the plot up 🤔 Here is what I ran:
```
python -m neural_lam.create_graph --config_path tests/datastore_examples/mdp/danra_100m_winds/config.yaml --hierarchical --name test_184_hi
python -m neural_lam.plot_graph --datastore_config_path tests/datastore_examples/mdp/danra_100m_winds/config.yaml --graph test_184_hi
```

joeloskarsson: Noticing this, it also makes me realize that we lack tests that `plot_graph.py` actually works. To make sure this code works we could maybe add a test that just runs that plotting for a flat and a hierarchical graph?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased](https://github.com/mllam/neural-lam/compare/v0.5.0...HEAD)
 
+### Changed
+
+- Refactor graph loading: move zero-indexing out of the model and update plotting to prepare using the research-branch graph I/O [\#184](https://github.com/mllam/neural-lam/pull/184)
```
joeloskarsson:
```suggestion
- Refactor graph loading: move zero-indexing out of the model and update plotting to prepare using the research-branch graph I/O [\#184](https://github.com/mllam/neural-lam/pull/184) @zweihuehner 
```
:)

neural_lam/plot_graph.py
zweihuehner: Regarding the test, really good idea!

For this I needed to expose a `plot_graph()` function for pytest, so it is now explicitly called by `main()` within plot_graph.py.

leifdenby:
I am just posting a TODO here to remember that we should 
- [ ] change the spec text on https://github.com/mllam/neural-lam/pull/184 once this PR has been merged in.

neural_lam/plot_graph.py
joeloskarsson: Great, seems to be fixed now!

neural_lam/plot_graph.py
```diff
@@ -103,22 +105,37 @@ def main():
...
+            )
+
+        for level, level_down_ei in enumerate(mesh_down_edge_index):
+            ei_down = level_down_ei.numpy().copy()
+            # Row 0: source in level l+1, Row 1: target in level l
+            ei_down[0] += level_offsets[level + 1]
+            ei_down[1] += level_offsets[level]
+            edge_plot_list.append(
+                (ei_down, "green", 1, f"Mesh down {level + 1}->{level}")
+            )
```
joeloskarsson, line 138: It's quite awkward having to do all this arithmetic with edge indices, but given how this script is structured I think that it is really the best solution for now.
At some point later we should change this so that the set of position each edge set starts and ends from is different, and not always this stacked matrix node_pos.

tests/test_plot_graph.py
```diff
@@ -0,0 +1,84 @@
...
+from neural_lam import utils
+from neural_lam.create_graph import create_graph_from_datastore
+from neural_lam.plot_graph import (
+    plot_graph,
+)
+from tests.dummy_datastore import DummyDatastore
+
+
+@pytest.fixture(params=["1level", "multiscale", "hierarchical"])
+def graph_fixture(request, tmp_path):
```
joeloskarsson: This test is really great! Thanks for this. 
Trying to keep test times down, do you see a way to use this great fixture also the graph creation tests? So that we do not need to build the same graphs twice?
If that becomes tricky we can just skip it for now.

zweihuehner: But because of your comment I have noticed that the default scope of the fixture is function-wide. So within the `test_plot_graph.py` module the graph creation was not shared between `test_returns_figure` and `test_save_html`. I have adapted it to a module-wide scope now, so thanks for the hint!

PR #189, opened 2026-02-19
fix: avoid NaN when standardizing fields with zero std #136 

### LINKED ISSUE #136: State standardization causes `NaN` if field's std is 0.
In the case state values are all the same and the standard deviation is 0, the standardization returns a `NaN` field, due to division by zero. While this does not cause an immediate error, the predictions are as a consequence all `NaN`.

Honestly, this case will probably rarely happen for normal runs, but for test runs this is not unlikely, e.g. rain rates might be all 0 if a time-period without rain has been chosen.

**Relevant code section:**
https://github.com/mllam/neural-lam/blob/ada9c1644beb56de832127cba46971b1be1bdd98/neural_lam/weather_dataset.py#L399-L414

One can deactivate the standardization, but that would currently affects all input features.

Shall we maybe do the standardization only if `da_state_std`/`da_forcing_std` is not 0?

varunsiravuri:
Closes #136

## Describe your changes

When all values of a state or forcing field are identical (e.g. rain
rates that are 0 throughout a dry period), the standard deviation is 0.
The current standardization code divides by `da_state_std` /
`da_forcing_std` unconditionally, producing `NaN` values via
division-by-zero. All subsequent predictions are then `NaN`.

Fix by guarding both the state and forcing standardization with
`xr.where(std > 0, (x - mean) / std, 0.0)` inside
`WeatherDataset._build_item_dataarrays`.

When `std == 0` the field is constant, so subtracting the mean already
yields `0` — we preserve that `0` instead of producing `NaN`.

No new dependencies are required; `xarray` is already imported.

## Issue Link

Closes #136

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass

cc @joeloskarsson @observingClouds — this addresses the bug you reported in #136.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased](https://github.com/mllam/neural-lam/compare/v0.5.0...HEAD)
 
+### Fixed
+
+- Avoid NaN when standardizing fields with zero std [#189](https://github.com/mllam/neural-lam/pull/189) @varunsiravuri
+
+- Fix README image paths to use absolute GitHub URLs so images display correctly on PyPI [\#188](https://github.com/mllam/neural-lam/pull/188) @bk-simon
+
+- Fix typo in `ar_model.py` that causes `AttributeError` during evaluation [\#204](https://github.com/mllam/neural-lam/pull/204) @ritinikhil
+
+- Changed the hardcoded True to a conditional check "persistent_workers=self.num_workers > 0" [\#235](https://github.com/mllam/neural-lam/pull/235) @santhil-cyber
+
+- `fractional_plot_bundle` now correctly multiplies by fraction instead of dividing [
+\#222](https://github.com/mllam/neural-lam/pull/222) @santhil-cyber
+
+- Fix Slack domain link [\#288](https://github.com/mllam/neural-lam/pull/288) @sadamov
+
+### Fixed
+
+- Infer spatial coordinate names for MDPDatastore (rather than assuming names `x` and `y`), allows for e.g. lat/lon regular grids [\#169](https://github.com/mllam/neural-lam/pull/169) @leifdenby
+
 ### Maintenance
 
 - Update PR template to clarify milestone/roadmap requirement and maintenance changes [\#186](https://github.com/mllam/neural-lam/pull/186) @joeloskarsson
@@ -18,10 +37,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update CI/CD to use only uv for full test suite and drop pdm [\#178](https://github.com/mllam/neural-lam/pull/178) @observingClouds
 
 - Fix caching of MEPS example data in CI/CD [\#181](https://github.com/mllam/neural-lam/pull/181) @observingClouds
-
-### Fixed
-
-- Avoid NaN when standardizing fields with zero std by clamping to machine epsilon [\#189](https://github.com/mllam/neural-lam/pull/189) @varunsiravuri
 
 ## [v0.5.0](https://github.com/mllam/neural-lam/releases/tag/v0.5.0)
 


varunsiravuri:
Test added in the latest commit that reproduces the original bug (naive division producing NaN) and confirms the fix (xr.where guard returning 0.0). Test passes locally.

joeloskarsson:
Hi! Some of the checklists from the PR template seems to be missing, could you add these back?

varunsiravuri:
Thanks for the review! I've added back the missing checklist sections. Also noting that the CHANGELOG entry was already added in the second commit — it's under `### Fixes` in the `[unreleased]` section.

AswaniSahoo:
Thanks for the fix @varunsiravuri! I've been looking into this issue as well, having worked on weather data normalization in my own ERA5 pipeline ([weather-transformer-scratch](https://github.com/AswaniSahoo/weather-transformer-scratch)), I ran into the same zero-std edge case with constant fields.

A few observations after reviewing the code:

1. The `xr.where` approach looks solid, setting the standardized value to `0.0` when `std == 0` makes physical sense since a constant field carries no information after mean subtraction.
2. One thing to consider: should the same guard also be applied in `datastore/base.py` → `_standardize_datarray()` (line ~213)? That helper also does `(da - mean) / std` and could hit the same issue depending on how it's called by different datastore backends.
3. A `loguru.logger.warning()` when zero-std is detected might help users debug unexpected behavior e.g., "State feature 'X' has std=0, setting standardized values to 0.0".


Happy to test this locally and report back. Great work on the fix! 👍

shauryam2807:
Hi @varunsiravuri, I tested this locally and the fix works as expected — no more NaN values when a field has zero variance. 👍

I had independently worked on a similar fix in #201 using a slightly different approach (`xr.where(std == 0, 1.0, std)` to replace the denominator, rather than wrapping the full expression). Both are mathematically equivalent since `(x - mean)` is guaranteed to be `0` when `std == 0`, but I think your approach of directly outputting `0.0` is arguably cleaner since it makes the intent more explicit.

One small note: the inline comment explaining *why* setting the result to `0.0` is correct (i.e., constant field → zero after mean subtraction) is very helpful for future contributors. Nice work!

sadamov:
Hi :wave: , thanks for spotting this edge case and for the discussion.
-> @varunsiravuri your current implementation breaks most dataset and plotting tests, please run `pytest` locally and fix these issues. Some tests might need to be updated with the same xr.where safeguard for 0 Std.

>One thing to consider: should the same guard also be applied in datastore/base.py → _standardize_datarray() (line ~213)? 

Yes, I agree

>A loguru.logger.warning() when zero-std is detected might help users debug unexpected behavior e.g., "State feature 'X' has std=0, setting standardized values to 0.0".

Sounds good!

- Replace std > 0 with std > 1e-8 to also catch near-constant fields. I think we should also catch numerical instabilities in our safeguard

Since this can be considered a bugfix we can merge this is as soon as its ready. I have assigned @varunsiravuri who opened the PR now. But it's of course great that multiple people are involved in the discussion. Thank you :pray:

varunsiravuri:
Hi @sadamov, I ran the full test suite on this branch. The `test_dataset_item_shapes` failures exist on `main` as well (pre-existing — confirmed by running `git stash` to remove my changes and re-running the tests). So those are unrelated to this PR.

I'm now working on:
1. Applying the same `xr.where(std > 1e-8, ...)` guard to `datastore/base.py` → `_standardize_datarray()`
2. Changing `std > 0` to `std > 1e-8` everywhere
3. Adding a `loguru.logger.warning()` when near-zero std is detected

Will push the updated commit shortly.

varunsiravuri:
Done! Pushed the updated commit (35dd533). 
changes are :

1.(std > 1e-8) everywhere-replaced `std > 0` with `std > 1e-8` in `weather_dataset.py` to also catch near-constant fields with numerical instabilities
2. datastore/base.py - applied the same `xr.where(std > 1e-8, ...)` guard to `_standardize_datarray()`, plus `.transpose(*da.dims)` to preserve dimension order
3. loguru warning — added `logger.warning()` in `weather_dataset.py` when any state or forcing feature has near-zero std

All 61 relevant tests pass locally.

sadamov:
@varunsiravuri thanks for the update.
- Could you please add the comment back about the forcing feature location (last dim), which holds still true and should be adressed in another PR
- There are a total of 111 tests and many of them still fail. I have a strong suspicion that there are two main issues with your implementation:
**Bug 1**: xr.where can silently reorder output dimensions, causing shape mismatches downstream; fixed by applying `.where(std > 1e-8, other=1.0)` on the 1-D std array alone, similar to @shauryam2807 logic above.
**Bug 2**: .item() is not supported on dask-backed arrays (used by the mdp/zarr datastore) and raises NotImplementedError; -> suggested fix: `if bool((self.da_state_std <= 1e-8).any()):`

@leifdenby do you agree with my suggested approach here that whenever we introduce new logging, we do it with loguru? Even if the script is using `import warnings` already?

varunsiravuri:
Hey @sadamov, I've addressed all three points from your review:

1. Restored the comment about forcing feature dimension assumption in _build_item_dataarrays
2. Fixed the (xr.where) dimension reordering bug by applying the threshold on the 1-D "std" array first  (std.where(std > 1e-8, other=1.0))rather than on the full multi-dimensional array — same approach used in "base.py".
3. Replaced .item() with bool(...) to support dask-backed arrays

All 113 tests pass locally including the full `test_datasets.py` suite.

sadamov:
@varunsiravuri please comment in #136 so that I can assign the issue to you

leifdenby:
> @leifdenby do you agree with my suggested approach here that whenever we introduce new logging, we do it with loguru? Even if the script is using `import warnings` already?

Yes, good idea 👍

varunsiravuri:
I’ve pushed the latest changes addressing all review points and updated the logging to use loguru for new messages. Please let me know if anything else would be helpful to adjust.

sadamov:
@varunsiravuri I fixed the CHANGELOG, please link the PR and not the Issue. And don't forget to add your **name** :wink:, next time.

@joeloskarsson I categorized this as a bugfix, but I think NaN-handling is always something that should be discussed. Could you please go over the implementation here that should prevent NaNs during standardization.

varunsiravuri:
Thanks @sadamov  Noted , will link the PR and include my name in the CHANGELOG from next time.

neural_lam/weather_dataset.py
```diff
@@ -398,21 +399,41 @@ def _build_item_dataarrays(self, idx):
...
+            if self.da_forcing_std is not None and bool(
+                (self.da_forcing_std <= 1e-8).any()
+            ):
+                logger.warning(
+                    "Some forcing features have near-zero std (<=1e-8) "
+                    "and will be set to 0.0 after standardization."
+                )
+            state_std_safe = self.da_state_std.where(
+                self.da_state_std > 1e-8, other=1.0
+            )
```
joeloskarsson: Now this code runs for each loaded sample. I think we should be very careful with what computations we do per-sample. So in particular:
1. It seems excessive to warn about this for each sample.
2. While these computations are light, it is still a check of all values in an array, so brings some overhead. Since neither `self.da_state_std` nor `self.da_forcing_std` change during runtime it should be sufficient to check this on dataset initialization.

neural_lam/weather_dataset.py
```diff
@@ -398,21 +399,41 @@ def _build_item_dataarrays(self, idx):
...
+                )
+            if self.da_forcing_std is not None and bool(
+                (self.da_forcing_std <= 1e-8).any()
+            ):
+                logger.warning(
+                    "Some forcing features have near-zero std (<=1e-8) "
+                    "and will be set to 0.0 after standardization."
+                )
+            state_std_safe = self.da_state_std.where(
+                self.da_state_std > 1e-8, other=1.0
```
joeloskarsson: Is there a specific motivation for 1e-8 here? I am a little bit scared that this might end up causing strange behavior for fields that just have a very small magnitude. In particular, would it not make more sense to clamp the standard deviation to a value (i.e. `other=1e-8` here), rather than having it "jump" to 1 once the standard deviation becomes too small. If we are really just trying to catch completely static fields than it seems like we should check for exactly 0 or <= machine epsilon. Happy to discuss these thoughts.

varunsiravuri: The motivation for 1e-8 was to also catch near-zero std fields caused by numerical precision issues, not just exactly-zero ones. That said, your point makes sense using other=1.0 does cause an unintended "jump" in behaviour for fields that are legitimately small in magnitude. Clamping with other=1e-8 is cleaner and more consistent. I'll update both base.py and weather_dataset.py to use other=1e-8. I'm also open to switching to <= machine epsilon if that's the preferred approach , happy to go with whatever you think fits best here!

joeloskarsson: Since @observingClouds encountered the original issue maybe you have some idea what would be best

neural_lam/weather_dataset.py
varunsiravuri: That's a valid point , since da_state_std and da_forcing_std don't change during runtime, there's no need to check them per sample. I'll move both the warning and the std_safe computation to __init__() so they only run once at dataset initialization.

neural_lam/weather_dataset.py
observingClouds: Nice discussion. My initial issue would be fixed with a simple check for exact `0` values, but I can see that including a machine epsilon would be a nice addition. Using 1e-8 could maybe indeed cause an issue for e.g. atmospheric trace gas concentrations, or in case of very local limited area models, where some forcing feels might vary very little.

varunsiravuri: Thanks for the context @observingClouds! That's a really useful point fields like atmospheric trace gas concentrations or subtle forcing variations in local area models could legitimately have std well below 1e-8, so using a fixed threshold there could incorrectly suppress real signal.
Given this, I think the safest approach is to keep the threshold at or below machine epsilon (np.finfo(float).eps ≈ 2.2e-16) rather than 1e-8.This would only catch fields that are numerically indistinguishable from zero, and avoid false positives on physically small but meaningful fields.

An alternative would be to make the threshold configurable (e.g., a zero_std_threshold parameter on the dataset), defaulting to machine epsilon, so users working with trace gas data can explicitly tune it for their use case.
@observingClouds @joeloskarsson @sadamov happy to go with whichever direction you prefer. Should I update the PR to use np.finfo(float).eps as the threshold?

observingClouds: I think we do not need a configurable threshold for now. The machine epsilon should be good enough. We should just make sure that we use the correct epsilon depending on the dtype.

neural_lam/weather_dataset.py
sadamov: yes, sorry I missed that after we switched to the fast 1D approach. but of course no need to recalculate this!

neural_lam/weather_dataset.py
sadamov: the `1e-8` was a rather arbitrary choice I think I have seen in single precision NWP-models. But the machine epsilon is the way to go here. If you could implement something along these lines @varunsiravuri that would be lovely:
`return (da - mean) / std.where(std > np.finfo(std.dtype).eps, other=np.finfo(std.dtype).eps)`

varunsiravuri: now using np.finfo(std.dtype).eps as the threshold and clamping with other=np.finfo(std.dtype).eps, making it dtype-aware. Thanks to everyone for the discussion!

neural_lam/weather_dataset.py
varunsiravuri: Done, moved both the warning and std_safe computation to __init__() they now only run once at dataset initialization.

CHANGELOG.md
```diff
@@ -9,16 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- Fix typo in `ar_model.py` that causes `AttributeError` during evaluation [\#204](https://github.com/mllam/neural-lam/pull/204) @ritinikhil
-
-- Changed the hardcoded True to a conditional check "persistent_workers=self.num_workers > 0" [\#235](https://github.com/mllam/neural-lam/pull/235) @santhil-cyber
-
-- `fractional_plot_bundle` now correctly multiplies by fraction instead of dividing [
-\#222](https://github.com/mllam/neural-lam/pull/222) @santhil-cyber
+### Fixed
 
-- Fix Slack domain link [\#288](https://github.com/mllam/neural-lam/pull/288) @sadamov
+- Avoid NaN when standardizing fields with zero std by clamping std to machine epsilon ([#189](https://github.com/mllam/neural-lam/pull/189)) @varunsiravuri
```
sadamov: make sure to keep all the latest changelogs from main branch

sadamov (changes requested):
this looks ready, except for the removed changelogs. please add those back and then we are good to merge.

CHANGELOG.md
varunsiravuri: Done! Synced the CHANGELOG with all the latest entries from main.

sadamov (changes requested):
In the changelog you are not supposed to change the existing entries but only add your new one. It looks like an agent took too many liberties here

neural_lam/datastore/base.py
```diff
@@ -213,7 +213,11 @@ def _standardize_datarray(
         mean = standard_da[f"{category}_mean"]
         std = standard_da[f"{category}_std"]
 
-        return (da - mean) / std
+        # Clamp std to machine epsilon (dtype-aware) to avoid NaN from
+        # division-by-zero for constant fields, without affecting fields
+        # with small but meaningful variance.
```
observingClouds: Comment not needed, definitely not in this detail. One can always do git blame if things are unclear.

neural_lam/weather_dataset.py
```diff
@@ -115,6 +116,31 @@ def __init__(
                 self.da_forcing_mean = self.ds_forcing_stats.forcing_mean
                 self.da_forcing_std = self.ds_forcing_stats.forcing_std
 
+            # Compute std_safe ONCE at init std arrays never
+            # change during runtime
```
observingClouds: This comment is unclear.

neural_lam/weather_dataset.py
```diff
@@ -115,6 +116,31 @@ def __init__(
...
+            if self.da_forcing_std is not None:
+                forcing_eps = np.finfo(self.da_forcing_std.dtype).eps
+                if bool((self.da_forcing_std <= forcing_eps).any()):
+                    logger.warning(
+                        "Some forcing features have near-zero std and will be "
+                        "standardized using machine epsilon to avoid NaN."
+                    )
+                self.forcing_std_safe = self.da_forcing_std.where(
+                    self.da_forcing_std > forcing_eps, other=forcing_eps
+                )
```
observingClouds: This is a lot of shared code between forcing and state. Please move to a function. This will also automatically lead to clearer code and the comment won't be needed as the code documents itself.

tests/test_datasets.py
observingClouds, line 1: Thanks for implementing a test. As I see this test is independent of the actual code base so it is not ensured that the fix within the core neural-lam is actually effective. Please adjust.

varunsiravuri:
Hi @observingClouds and @sadamov, I've addressed all the review feedback. Here's a summary of what was changed:
@observingClouds:

Removed the verbose comment in datastore/base.py

Removed the unclear std_safe comment in weather_dataset.py

Extracted a _compute_std_safe() helper method to eliminate the duplicated state/forcing logic

Rewrote the regression test in test_datasets.py to directly exercise WeatherDataset._compute_std_safe() from the actual codebase

@sadamov:

Reset the CHANGELOG.md to only add the new ### Fixed entry under [unreleased] with a link to this PR , no existing entries were modified

Please let me know if anything needs further adjustment. Thanks for the thorough review!

neural_lam/weather_dataset.py
```diff
@@ -115,6 +116,26 @@ def __init__(
...
+            )
+
+            if self.da_forcing_std is not None:
+                self.forcing_std_safe = self._compute_std_safe(
+                    self.da_forcing_std, "forcing"
+                )
+            else:
+                self.forcing_std_safe = None
+
+    def _compute_std_safe(self, std_da, label: str):
```
observingClouds: Can we rename the `label` attribute to `feature` to better reflect what it is?
```suggestion
    def _compute_std_safe(self, std_da, feature: str):
```

Could you also please add the type annotation for std_da, I'm not sure right now if it is e.g. `xr.DataArray` or if it can be (also) something else. Please check and then rename to e.g. `std` as the type will then already tell us what it is.

sadamov:
@varunsiravuri thanks for the improvements! code is looking good. you are however still removing 17 lines from the changelog. Make sure to use the latest version of the changelog from the main branch before adding yours.

varunsiravuri:
sorry my bad @sadamov , just added mine to the main changelog!

sadamov (approved):
LGTM

observingClouds:
Handing over to you @sadamov as assignee

varunsiravuri:
> LGTM too! Thanks @varunsiravuri to keep working on it 🚀

thank you @observingClouds and @sadamov  would love to contribute more !

sadamov:
@varunsiravuri great work, thanks for this contribution and happy to have you here! merging now...

PR #197, opened 2026-02-20
feat: expose wandb.init resume and id arguments in train_model

### LINKED ISSUE #129: Expose `wandb.init` argument `resume` to `neural-lam.train_model`
Wandb.init has a `resume` flag where you can continue existing experiments based on id: https://docs.wandb.ai/ref/python/init/. I find that quite useful, especially on HPC systems that crash often or have limited runtime for jobs.

This should require changes only in `train_model` and in `utils`.

Mani212005:
### Summary

This PR exposes the `resume` and `id` arguments from `wandb.init` through the `train_model` CLI.
This is particularly useful on HPC systems where jobs have limited runtimes or may crash
mid-training, allowing users to seamlessly continue an existing W&B run rather than starting a new one.

#### `train_model.py`

- Added `--wandb_resume` argument (`choices: allow | must | never | auto`) — passed directly to `wandb.init(resume=...)`.
- Added `--wandb_id` argument — passed directly to `wandb.init(id=...)` to target a specific existing run.
- Both arguments are grouped under a `# Wandb-specific settings` comment, consistent with the existing `# Logger Settings` grouping convention.

#### `utils.py`

- Forwarded `args.wandb_resume` and `args.wandb_id` to `pl.loggers.WandbLogger`.
- When `--wandb_id` is provided, `name` is set to `None` to preserve the existing run's display name in the W&B UI instead of overwriting it with a freshly generated one.

### Issue Link

Closes #129 

### Type of change

- [ ]    Bug fix (non-breaking change that fixes an issue)
- [x]   New feature (non-breaking change that adds functionality)
- [ ]   Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ]    Documentation (Addition or improvements to documentation)

### Checklist before requesting a review

- [ ] I have updated the **README** to cover introduced code changes

### Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

### Author checklist after completed review

- [x] I have added a line to the CHANGELOG describing this change, in a section reflecting type of change (add section where missing):
  - `added` — when you have added new functionality
  - `changed` — when default behaviour of the code has been changed
  - `fixes` — when your contribution fixes a bug
  - `maintenance` — when your contribution relates to repo maintenance, e.g. CI/CD or documentation

### Checklist for assignee

- [x] the tests pass
- [x] author has added an entry to the changelog (and designated the change as `added`, `changed`, `fixed` or `maintenance`)
- [x] Once the PR is ready to be merged, squash commits and merge the PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed the hardcoded True to a conditional check "persistent_workers=self.num_workers > 0" [\#235](https://github.com/mllam/neural-lam/pull/235) @santhil-cyber
 
-- `fractional_plot_bundle` now correctly multiplies by fraction instead of dividing 
+- `fractional_plot_bundle` now correctly multiplies by fraction instead of dividing
 [\#222](https://github.com/mllam/neural-lam/pull/222) @santhil-cyber
 
 ### Added


Mani212005:
Hello @joeloskarsson
 kindly please review the code. Let me know if there is any feedback. Thank you!

observingClouds:
Thank you for the PR @Mani212005. I think we do not need to expose the `--wandb_resume` to the CLI and can set `resume` always to `must` or `auto` whenever the `id` is given.

I am curious what @sadamov thinks about this

Mani212005:
@observingClouds Thank you for the Feed back. I agree with you that --wandb_resume should not be exposed to CLI. I can remove --wandb_resume from the CLI to keep things simple.  

I propose hardcoding [resume="allow"] whenever [--wandb_id] is given, What do you think about this?

sadamov:
@Mani212005 thanks for implementing this! I agree with @observingClouds that we don't need two extra flags, but the wandb_id is sufficient with mode "allow" for resume, which will:

If the user provides a run id:
- If that id already exists in the project, W&B resumes that run.
- If that id does not exist, W&B creates a new run with that id (no hard failure).
- If no id is provided, there is nothing specific to resume; a normal new run is created.

It would be good if you could also add a statement about the wandb resume behavior to the logger, so that the behavior is clearly visible.

I ran pre-commits and pytest on your branch and they look OK. Just rerun them after your changes.
Later, please update the CHANGELOG.md with the new feature you added.

This issue/PR introduces a new feature, but it has not been assigned to a milestone. So before we can merge the PR we need to discuss the appropriate milestone in the next dev-meeting (occuring monthly).

Mani212005:
@sadamov 
I have implemented all the requested changes : 
-->  Removed `--wandb_resume` flag --- Simplified to only `--wandb_id` argument.
-->  Auto-inferred `resume="allow"` --- When `--wandb_id` is provided
- If run ID exists → resumes that run
- If run ID doesn't exist → creates new run with that ID
- If no ID → normal new run (unchanged behavior)
--> Added logger statement --- Resume mode and run ID are now logged.
--> CHANGELOG.md updated --- Added feature entry with PR reference.
-->  Tests added
- 6 unit tests covering all cases
- Passing in ~4 seconds with `--noconftest`
 
Let me know if you think something should be done differently. Thank you!

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased](https://github.com/mllam/neural-lam/compare/v0.5.0...HEAD)
 
+### Added
+
+- Expose `--wandb_id` CLI argument to allow resuming an existing W&B run by
+  ID. When provided, `resume="allow"` is set automatically so the same job
+  script works for both the initial submission and all resubmissions, making
+  it suitable for HPC systems with limited job runtimes or that may crash.
+  [\#197](https://github.com/mllam/neural-lam/pull/197) @Mani212005
```
sadamov, line 27:
```suggestion
  script works for both the initial submission and all resubmissions,.
  [\#197](https://github.com/mllam/neural-lam/pull/197) @Mani212005
```

neural_lam/train_model.py
```diff
@@ -192,6 +192,18 @@ def main(input_args=None):
...
+    # Wandb-specific settings
+    parser.add_argument(
+        "--wandb_id",
+        type=str,
+        default=None,
+        help="Wandb run ID to use. If the run ID already exists in the "
+        "project, W&B resumes that run. If it does not exist, W&B creates "
+        "a new run with that ID. Useful on HPC systems with limited job "
+        "runtimes or that may crash, allowing training to be continued "
+        "across multiple job submissions.",
```
sadamov, line 205:
```suggestion
        "a new run with that ID. If the run ID is None a new job with random ID is submitted. This argument is ignored for any --logger that is not wandb.",
```

neural_lam/utils.py
```diff
@@ -351,10 +352,21 @@ def setup_training_logger(datastore, args, run_name):
     """
 
     if args.logger == "wandb":
-        logger = pl.loggers.WandbLogger(
+        # If a run id is given, use resume="allow": resumes the run if it
+        # already exists, otherwise creates a new run with that id.
+        # This makes it safe to reuse the same job script across the initial
+        # submission and all resubmissions on.
```
sadamov: please put these comments into the docstring of the function. there is also a typo the last "on" is dangling.

neural_lam/utils.py
```diff
@@ -351,10 +352,21 @@ def setup_training_logger(datastore, args, run_name):
...
+        wandb_resume = "allow" if args.wandb_id else None
+        loguru_logger.info(
+            f"Wandb resume mode: {wandb_resume!r} (id: {args.wandb_id!r})"
+        )
+        return pl.loggers.WandbLogger(
             project=args.logger_project,
-            name=run_name,
+            # When resuming by id, don't override the run's existing name
```
sadamov: put this into the docstring as well

neural_lam/utils.py
```diff
@@ -351,10 +352,21 @@ def setup_training_logger(datastore, args, run_name):
...
     if args.logger == "wandb":
-        logger = pl.loggers.WandbLogger(
+        # If a run id is given, use resume="allow": resumes the run if it
+        # already exists, otherwise creates a new run with that id.
+        # This makes it safe to reuse the same job script across the initial
+        # submission and all resubmissions on.
+        wandb_resume = "allow" if args.wandb_id else None
+        loguru_logger.info(
+            f"Wandb resume mode: {wandb_resume!r} (id: {args.wandb_id!r})"
+        )
```
sadamov, line 371: also add a warning when the user sets wandb_id but choses mlflow as logger

```python
    if args.wandb_id and args.logger != "wandb":
        loguru_logger.warning(
            f"--wandb_id is set but logger is {args.logger!r}; "
            "the wandb_id will have no effect."
        )
```

tests/test_wandb_resume.py
```diff
@@ -0,0 +1,95 @@
...
+    """With --wandb_id, name must be None to preserve the existing run name."""
+    from neural_lam.utils import setup_training_logger
+
+    args = _make_args(wandb_id="abc123")
+    datastore = MagicMock()
+    datastore._config = {}
+
+    setup_training_logger(datastore, args, run_name="my-run")
+
+    _, kwargs = mock_wandb.call_args
```
sadamov: These two tests can be combined as they share most of their setup.

tests/test_wandb_resume.py
```diff
@@ -0,0 +1,95 @@
...
+
+def _make_parser():
+    """Reconstruct only the logger-related args from train_model's parser,
+    so this test has no heavy imports and runs instantly."""
+    parser = ArgumentParser()
+    parser.add_argument("--logger", type=str, default="wandb")
+    parser.add_argument("--logger-project", type=str, default="neural_lam")
+    parser.add_argument("--logger_run_name", type=str, default=None)
+    parser.add_argument("--wandb_id", type=str, default=None)
+    return parser
```
sadamov: use the real parser here instead of a new mock one, like:

```python
def _parse_real(args_list):
    """Parse args through the real train_model parser."""
    from neural_lam.train_model import _build_arg_parser

    return _build_arg_parser().parse_args(args_list)
```

tests/test_wandb_resume.py
sadamov, line 1: add a new test that asserts wandb_id is ignored when logger is mlflow and the warning is emitted

Mani212005:
@sadamov 
I have implemented the suggestions and changed the code accordingly! Let me know what do you think about this

sadamov:
@Mani212005 please comment in #129 so that I can assign the issue to you

sadamov:
@Mani212005 I think this is looking good, but the PR is now touching two more general design choices where we need input from some people:

@observingClouds: should this feature be expanded to mlflow? and if so, I am usure if mlflow has a "resume=allow" logic. I did see however that their ids are much longer.

@joeloskarsson: I like the idea of splitting main from args creation and parsing. But not sure if we should hide such a big change inside this PR. @Mani212005 can you explain why you think the split is needed. I assume its related to the decorator you introduced for main?

Mani212005:
@sadamov yes, The split was needed specifically because of the @logger.catch decorator on  main. That decorator catches all exceptions including AssertionError and SystemExit  which makes it impossible to test argument parsing behaviour directly by calling  in a test .
 By extracting [_build_arg_parser()], the tests can import and call just the parser without going through [@logger.catch], so we can write clean unit tests like:
``` 
args = _build_arg_parser().parse_args(["--wandb_id", "abc123"])
assert args.wandb_id == "abc123" 
```

The split was just for making testing better and more convenient. Without the split, the only way to test would be to mock the decorator away or call the real CLI as a subprocess both are more fragile.  This was my thought process.

joeloskarsson:
I agree that it is a good change, and I understand your logic @Mani212005 . But I also agree with @sadamov that it feels like a large unrelated change in this PR. Even though I understand that it is introduced here so that a proper test can actually be written for this PR.

I think my preference, just to keep things structured, would be to not do this change here, but for now just mock the decorator away so that this change can be tested and merged. Then open a separate PR that is about improving argparse testing that makes this improvement.

Mani212005:
@sadamov I have done the required changes. have removed the decorator part. Will open a separate PR for the decorator as @joeloskarsson suggested. 
Thank you

Mani212005:
@sadamov Have implemented the suggested changes also fixed the rank state bug properly. Added the ` ensure_rank_zero ` fixture to `conftest.py` with `autouse=True ` as you suggested. Thanks! :)

sadamov:
@Mani212005 thanks again for this contribution! great work :rocket:

Issue #49, opened 2024-06-02
Refactor model class hierarchy


joeloskarsson:
## Background
The different models that can be trained in Neural-LAM are currently all sub-classes of `pytorch_lightning.LightningModule`. In particular, much of the functionality  sits in the first subclass, `ARModel`. The current hierarchy looks like this:
![classes_current](https://github.com/mllam/neural-lam/assets/16475490/0cfdead6-2971-4a3b-b42f-8d80fe850c0d)
(I am making these rather than some more fancy UML-diagrams since I think this should be enough for the level of detail we need to discuss here).

## The problem
In the current hierarchy everything is a subclass of `ARModel`. This has a number of drawbacks:

* All functionality ends up in one class. There is no clear division of responsibility, but rather we end up with one class that does everything from logging, device-handling, unrolling forecasts and the actual forward and backward pass through GNNs. 
* The subclasses do not utilize the standard torch `forward` calls, but rather must resorts to our own similar construction (e.g. `predict_step`)
* This limits us to deterministic, auto-regressive models.
* This is hard to extend upon for ensemble-models.

## Proposed new hierarchy
I propose to split up the current class hierarchy into subclasses that have clear responsibilities. These should not just all inherit `ARModel`, but rather be members of each other as suitable. A first idea for this is shown below, including also potential future classes for new models (to show how this is more extendible):

![classes_proposal](https://github.com/mllam/neural-lam/assets/16475490/53bc4e7d-abf0-44b8-9819-ed59d4e3ca89)

The core components are (I here count static features as part of forcing):

* `ForecasterModule`: Takes over much of the responsibility of the old `ARModel`. Handles things not directly related to the nerual network components such as plotting, logging, moving batches to the right device. This inherits `pytorch_lightning.LightningModule` and have the different train/val/test steps. In each step (train/val/test), unpacks the batch of tensors and uses a `Forecaster` to produce a full forecast. Also responsible for computing the loss based in a produced forecast (could also be in `Forecaster`, not entirely sure about this).
* `Forecaster`: A generic forecaster capable of mapping from a set of initial states, forcing and boundary forcing into a full forecast of the requested length.
  * `ARForecaster`: Subclass of `Forecaster` that uses an auto-regressive strategy to unroll a forecast. Makes use of a `StepPredictor` at each AR step.
[...]

PR #208, opened 2026-02-21
Refactor model class hierarchy into composable Forecaster/StepPredictor layers

Sir-Sloth-The-Lazy:
## Describe your changes

Refactors the monolithic `ARModel` class into a composable hierarchy of smaller, focused components:

- **`ForecasterModule`** (`pl.LightningModule`): Training loop, metrics, plotting, optimizer config
- **`ARForecaster`** (`nn.Module`): Auto-regressive unrolling with boundary masking  
- **`StepPredictor`** (`nn.Module`): Single-step prediction, normalization, clamping

This separation makes it straightforward to add non-autoregressive forecasters, new step predictor architectures (e.g. Vision Transformers), or ensemble strategies without modifying the training infrastructure.

Also fixes two pre-existing bugs:
- `interior_mask_bool` shape `(1,)` → `(N,)` for correct loss masking
- `all_gather_cat` dimension collapse on single-device runs

## Issue Link

Closes #49

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form
- [x] I have requested a reviewer and an assignee

PR #209, opened 2026-02-21
Extract plot_examples from ARmodel.py into stateless function in vis.py

### LINKED ISSUE #223: Refactor: Decouple WeatherDataset from ARModel._create_dataarray_from_tensor
## Summary
`ARModel._create_dataarray_from_tensor` currently instantiates a full [WeatherDataset](Desktop/neurallam/neural_lam/weather_dataset.py:15:0-600:17) 
on every call just to access coordinate metadata for constructing `xr.DataArray` objects.

https://github.com/mllam/neural-lam/blob/main/neural_lam/models/ar_model.py#L191-L198

## Problem
Each call triggers `WeatherDataset.__init__()`, which:
- Loads all state/forcing dataarrays from the datastore
- Validates dimension ordering
- Computes standardization statistics

...all just to extract `grid_index`, feature names, and `x`/`y` coordinates for plotting.

This is flagged as a TODO in the code itself:
> "creating an instance of WeatherDataset here on every call is not how this should be done"

## Proposed Solution
Call `self._datastore.get_dataarray(category, split)` directly inside [ARModel](Desktop/neural-lam/neural_lam/models/ar_model.py:23:0-823:63) to get 
coordinate metadata, and construct the `xr.DataArray` inline. This removes the 
[WeatherDataset](Desktop/neural-lam/neural_lam/weather_dataset.py:15:0-600:17) import from [ar_model.py](Desktop/neural-lam/neural_lam/models/ar_model.py:0:0-0:0) entirely, cleaning up the dependency graph.

I have implemented this solution locally and if this sounds good, I can raise a PR for it to be merged.

oleeviyababu:
## Describe your changes

Summary of the changes.

I extracted the "plot_examples" method from the "ARModel" class which lies in the file ar_model.py and moved it into "vis.py" as an independant , stateless function.

Please also include relevant motivation and context. 
As discussed in the Issue #49, the `ARModel` class is currently the main class that handles neural network logic, loss calculations, and complex plotting. As a part of the preparation for the upcoming GSoC Probabilistic Refactoring, this PR keep down the `ARModel` by isolating the plotting logic. 

List any dependencies that are required for this change. 
None 

## Issue Link

 https://github.com/mllam/neural-lam/issues/49

## Type of change

- [x ]   Refactoring the code to improve the structire

- [x ] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x ] I have performed a self-review of my code
- [ x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [ x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [x ] I have added tests that prove my fix is effective or that my feature works
- [ x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -192,6 +192,10 @@ This release introduces new features including GIF animation support, wandb run 
 
 - Fix Slack domain link [\#288](https://github.com/mllam/neural-lam/pull/288) @sadamov
 
+### Refactored
+
+- Refactored `ARModel`: Extracted plotting logic into stateless functions in `vis.py` [\#209](https://github.com/mllam/neural-lam/pull/209) @oleeviyababu
+
 ### Maintenance
 
 - Update PR template to clarify milestone/roadmap requirement and maintenance changes [\#186](https://github.com/mllam/neural-lam/pull/186) @joeloskarsson
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -194,7 +194,9 @@ This release introduces new features including GIF animation support, wandb run 
 
 ### Changed
 
-- Refactored `ARModel`: Extracted plotting logic into stateless functions in `vis.py` [\#209](https://github.com/mllam/neural-lam/pull/209) @oleeviyababu
+- Refactored `ForecasterModule`: extracted example-plotting logic into a
+  standalone `plot_examples` function in `vis.py`, with the model now
+  delegating to it [\#209](https://github.com/mllam/neural-lam/pull/209) @oleeviyababu
 
 ### Maintenance
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -192,7 +192,7 @@ This release introduces new features including GIF animation support, wandb run 
 
 - Fix Slack domain link [\#288](https://github.com/mllam/neural-lam/pull/288) @sadamov
 
-### Refactored
+### Changed
 
 - Refactored `ARModel`: Extracted plotting logic into stateless functions in `vis.py` [\#209](https://github.com/mllam/neural-lam/pull/209) @oleeviyababu
 
--- a/neural_lam/datastore/base.py
+++ b/neural_lam/datastore/base.py
@@ -462,17 +462,41 @@ class BaseDatastore(abc.ABC):
     def create_dataarray_from_tensor(
         self,
         tensor: torch.Tensor,
-        time: Union[datetime.datetime, list[datetime.datetime]],
+        time: datetime.datetime | list[datetime.datetime],
         category: str,
     ) -> xr.DataArray:
         """
-        Construct a xarray.DataArray from a `pytorch.Tensor` with coordinates
-        for `grid_index`, `time` and `{category}_feature` matching the shape
-        and number of times provided and add the x/y coordinates from the
-        datastore.
+        Construct an xarray.DataArray from a torch.Tensor.
+
+        The tensor shape determines how ``time`` is interpreted:
+
+        - 2-D ``(grid_index, {category}_feature)``: ``time`` must be a
+          single timestamp, stored as a scalar coordinate.
+        - 3-D ``(n_timesteps, grid_index, {category}_feature)``: ``time``
+          must be a list of ``n_timesteps`` timestamps, stored as a
+          dimension.
+
+        The x/y coordinates from the datastore are attached to
+        ``grid_index`` so the DataArray can be unstacked into a 2-D grid.
+
+        Parameters
+        ----------
+        tensor : torch.Tensor
+            2-D or 3-D tensor to convert.
+        time : datetime.datetime or list of datetime.datetime
+            Timestamp(s) aligned with the tensor's time dimension.
+        category : str
+            Data category (``"state"``, ``"forcing"``, or ``"static"``).
+
+        Returns
+        -------
+        xarray.DataArray
+            DataArray with coordinates for grid_index, time, and
+            {category}_feature, plus attached x/y coordinates.
         """
 
         def _is_listlike(obj):
+            """Return ``True`` for list/tuple/ndarray-like containers."""
             return hasattr(obj, "__iter__") and not isinstance(obj, str)
 
         add_time_as_dim = False
--- a/neural_lam/datastore/base.py
+++ b/neural_lam/datastore/base.py
@@ -4,6 +4,7 @@
 import abc
 import collections
 import dataclasses
+import datetime
 import functools
 from datetime import timedelta
 from functools import cached_property
@@ -12,6 +13,7 @@ from pathlib import Path
 # Third-party
 import cartopy.crs as ccrs
 import numpy as np
+import torch
 import xarray as xr
 from pandas.core.indexes.multi import MultiIndex
 
@@ -456,6 +458,85 @@ class BaseDatastore(abc.ABC):
             dim_order.append(f"{category}_feature")
 
         return tuple(dim_order)
+
+    def create_dataarray_from_tensor(
+        self,
+        tensor: torch.Tensor,
+        time: Union[datetime.datetime, list[datetime.datetime]],
+        category: str,
+    ):
+        """
+        Construct a xarray.DataArray from a `pytorch.Tensor` with coordinates
+        for `grid_index`, `time` and `{category}_feature` matching the shape
+        and number of times provided and add the x/y coordinates from the
+        datastore.
+        """
+
+        def _is_listlike(obj):
+            return hasattr(obj, "__iter__") and not isinstance(obj, str)
+
+        add_time_as_dim = False
+        if len(tensor.shape) == 2:
+            dims = ["grid_index", f"{category}_feature"]
+            if _is_listlike(time):
+                raise ValueError(
+                    "Expected a single time for a 2D tensor with assumed "
+                    f"dimensions (grid_index, {category}_feature), but got "
+                    f"{len(time)} times"  # type: ignore
+                )
+        elif len(tensor.shape) == 3:
+            add_time_as_dim = True
+            dims = ["time", "grid_index", f"{category}_feature"]
+            if not _is_listlike(time):
+                raise ValueError(
+                    "Expected a list of times for a 3D tensor with assumed "
+                    f"dimensions (time, grid_index, {category}_feature), but "
+                    "got a single time"
+                )
+        else:
+            raise ValueError(
+                "Expected tensor to have 2 or 3 dimensions, but got "
+                f"{len(tensor.shape)}"
+            )
+
+        # Get original dataarray to extract proper coordinates
+        # Get original dataarray to extract proper coordinates
+        da_datastore_state = self.get_dataarray(category=category, split="test")
+
+        if da_datastore_state is None:
+            raise ValueError(
+                f"Could not load dataarray for category '{category}'"
+            )
+
+        da_grid_index = da_datastore_state.grid_index
+
+        # Safely fetch the correct feature category
+        da_category_feature = getattr(da_datastore_state, f"{category}_feature")
+
+        coords = {
+            f"{category}_feature": da_category_feature,
+            "grid_index": da_grid_index,
+        }
+        if add_time_as_dim:
+            coords["time"] = time
+
+        da = xr.DataArray(
+            tensor.cpu().numpy(),
+            dims=dims,
+            coords=coords,
+        )
+
+        for grid_coord in ["x", "y"]:
+            if (
+                grid_coord in da_datastore_state.coords
+                and grid_coord not in da.coords
+            ):
+                da.coords[grid_coord] = da_datastore_state[grid_coord]
+
+        if not add_time_as_dim:
+            da.coords["time"] = time
+
+        return da
 
 
 @dataclasses.dataclass
--- a/neural_lam/datastore/base.py
+++ b/neural_lam/datastore/base.py
@@ -464,7 +464,7 @@ class BaseDatastore(abc.ABC):
         tensor: torch.Tensor,
         time: Union[datetime.datetime, list[datetime.datetime]],
         category: str,
-    ):
+    ) -> xr.DataArray:
         """
         Construct a xarray.DataArray from a `pytorch.Tensor` with coordinates
         for `grid_index`, `time` and `{category}_feature` matching the shape
@@ -499,43 +499,27 @@ class BaseDatastore(abc.ABC):
                 f"{len(tensor.shape)}"
             )
 
-        # Get original dataarray to extract proper coordinates
-        # Get original dataarray to extract proper coordinates
-        da_datastore_state = self.get_dataarray(category=category, split="test")
-
-        if da_datastore_state is None:
-            raise ValueError(
-                f"Could not load dataarray for category '{category}'"
-            )
-
-        da_grid_index = da_datastore_state.grid_index
-
         # Safely fetch the correct feature category
-        da_category_feature = getattr(da_datastore_state, f"{category}_feature")
-
-        coords = {
-            f"{category}_feature": da_category_feature,
-            "grid_index": da_grid_index,
+        feature_names = self.get_vars_names(category)
+        grid_index = np.arange(self.num_grid_points)
+        coords: dict = {
+            f"{category}_feature": feature_names,
+            "grid_index": grid_index,
         }
         if add_time_as_dim:
             coords["time"] = time
-
         da = xr.DataArray(
             tensor.cpu().numpy(),
             dims=dims,
             coords=coords,
         )
-
-        for grid_coord in ["x", "y"]:
-            if (
-                grid_coord in da_datastore_state.coords
-                and grid_coord not in da.coords
-            ):
-                da.coords[grid_coord] = da_datastore_state[grid_coord]
-
+        # Attach x, y coordinates so the DataArray can later be unstacked
+        # into a 2D spatial grid via unstack("grid_index")
+        xy = self.get_xy(category, stacked=True)
+        da.coords["x"] = xr.DataArray(xy[:, 0], dims=["grid_index"])
+        da.coords["y"] = xr.DataArray(xy[:, 1], dims=["grid_index"])
         if not add_time_as_dim:
             da.coords["time"] = time
-
         return da
 
 
--- a/neural_lam/vis.py
+++ b/neural_lam/vis.py
@@ -1,6 +1,7 @@
 """Visualization helpers for analysing Neural-LAM predictions and errors."""
 
 # Standard library
+import os
 import warnings
 from typing import Optional
 
@@ -14,9 +15,10 @@ import matplotlib.colors
 import matplotlib.figure
 import matplotlib.pyplot as plt
 import numpy as np
-
+import pytorch_lightning as pl
 import torch
 import xarray as xr
+from PIL import Image
 
 # Local
 from . import utils
@@ -26,11 +28,6 @@ from .datastore.base import BaseRegularGridDatastore
 _TITLE_SIZE = 13  # suptitle and per-axes titles
 _LABEL_SIZE = 11  # axis / colorbar labels
 _TICK_SIZE = 11  # tick labels
-
-import os
-import warnings
-import torch
-import pytorch_lightning as pl
 
 # Annotations become unreadable when cells are smaller than this (in points)
 # or when the total number of cells exceeds a readable count.
@@ -790,17 +787,28 @@ def plot_examples(
     target: torch.Tensor,
     time_batch: torch.Tensor,
     first_example_idx: int = 0,
+    create_gif: bool = False,
 ) -> None:
     """
     Plot example forecasts from provided tensors.
 
-    Args:
-        datastore: The object containing dataset metadata.
-        logger: The logger instance used to save the images.
-        prediction: Output tensors predicted from the model.
-        target: Ground truth tensors.
-        time_batch: Time timestamps corresponding to the data.
-        first_example_idx: Starting index for naming saved files/logs.
+    Parameters
+    ----------
+    datastore : BaseRegularGridDatastore
+        The object containing dataset metadata.
+    logger : pl.loggers.Logger
+        The logger instance used to save the images.
+    prediction : torch.Tensor
+        Output tensors predicted from the model.
+    target : torch.Tensor
+        Ground truth tensors.
+    time_batch : torch.Tensor
+        Time timestamps corresponding to the data.
+    first_example_idx : int, optional
+        Starting index for naming saved files/logs.
+    create_gif : bool, optional
+        If True, save per-variable animated GIFs (one frame per prediction
+        step) alongside the logged figures.
     """
     time_step_int, time_step_unit = utils.get_integer_time(
         datastore.step_length
@@ -843,6 +851,15 @@ def plot_examples(
         )
         var_vranges = list(zip(var_vmin, var_vmax))
 
+        if create_gif:
+            plot_dir_path = os.path.join(
+                logger.save_dir, f"example_plots_{example_i}"
+            )
+            os.makedirs(plot_dir_path, exist_ok=True)
+            png_frames: dict[str, list[str]] = {
+                var_name: [] for var_name in datastore.get_vars_names("state")
+            }
+
         for t_i, _ in enumerate(zip(pred_slice, target_slice), start=1):
             var_figs = [
                 plot_prediction(
@@ -880,7 +897,36 @@ def plot_examples(
                 else:
                     warnings.warn(f"{logger} does not support image logging.")
 
+                if create_gif:
+                    png_path = os.path.join(
+                        plot_dir_path,
+                        f"{var_name}_example_{example_i}"
+                        f"_prediction_t_{t_i:02d}.png",
+                    )
+                    fig.savefig(png_path, dpi=100, bbox_inches="tight")
+                    png_frames[var_name].append(png_path)
+
             plt.close("all")
+
+        if create_gif:
+            for var_name, frames_for_var in png_frames.items():
+                if frames_for_var:
+                    gif_path = os.path.join(
+                        plot_dir_path,
+                        f"{var_name}_example_{example_i}_prediction.gif",
+                    )
+                    frames = [Image.open(f) for f in frames_for_var]
+                    try:
+                        frames[0].save(
+                            gif_path,
+                            save_all=True,
+                            append_images=frames[1:],
+                            loop=0,
+                            duration=1000,
+                        )
+                    finally:
+                        for frame in frames:
+                            frame.close()
 
         pred_filename = f"example_pred_{example_i}.pt"
         torch.save(
--- a/neural_lam/vis.py
+++ b/neural_lam/vis.py
@@ -21,6 +21,7 @@ import xarray as xr
 # Local
 from . import utils
 from .datastore.base import BaseRegularGridDatastore
+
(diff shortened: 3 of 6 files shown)


sadamov:
Thanks @oleeviyababu, I just had an initial look at your suggested changes. The general direction of relieving ARModel of some of its duties is great and in line with https://github.com/mllam/neural-lam/issues/49. I also really like the fact that you managed to get rid of the `_create_datarray_from_tensor` helper function. At least so far I couldn't see any downside of doing that.

Here are some more detailed comments from my side:
- The creation of the `weather_dataset` object should happen outside of the for loop
- The new `plot_examples` function is missing a proper docstring
- Please make sure that all objects you introduce are typed (we have legacy code that doesn't enforce static typing, but new additions must -> see PR template)
- Don't forget to run pre-commits you have some unused imports left over

nitpick: i wouldn't say the new function is strictly stateless as its main job is to generate, save and log artifacts. but that's just semantics :)

sadamov:
There is one issue that I ran into while testing some scenarios, and I am not completely sure how to handle it best. So I added @joeloskarsson for his input as well. It's about **detaching** tensors:

## Current code in PR-branch

- `neural_lam/vis.py` (`plot_examples`) currently passes `pred_slice`/`target_slice` directly to `create_dataarray_from_tensor`.
- `neural_lam/weather_dataset.py` (`create_dataarray_from_tensor`) currently uses `tensor.cpu().numpy()` (no detach).

> This can fail if a tensor still has `requires_grad=True`

## Recommended adjustment

Use detach at the plotting boundary,  plotting is explicitly non-grad work.
`WeatherDataset` stays a neutral utility and does not silently alter autograd behavior.

### Update in `neural_lam/vis.py`

- Detach slices before conversion:
  - `pred_slice = pred_slice.detach()`
  - `target_slice = target_slice.detach()`
- Then pass detached tensors to `create_dataarray_from_tensor`.

oleeviyababu:
> There is one issue that I ran into while testing some scenarios, and I am not completely sure how to handle it best. So I added @joeloskarsson for his input as well. It's about **detaching** tensors:
> 
> ## Current code in PR-branch
> * `neural_lam/vis.py` (`plot_examples`) currently passes `pred_slice`/`target_slice` directly to `create_dataarray_from_tensor`.
> * `neural_lam/weather_dataset.py` (`create_dataarray_from_tensor`) currently uses `tensor.cpu().numpy()` (no detach).
> 
> > This can fail if a tensor still has `requires_grad=True`
> 
> ## Recommended adjustment
> Use detach at the plotting boundary, plotting is explicitly non-grad work. `WeatherDataset` stays a neutral utility and does not silently alter autograd behavior.
> 
> ### Update in `neural_lam/vis.py`
> * Detach slices before conversion:
>   
>   * `pred_slice = pred_slice.detach()`
>   * `target_slice = target_slice.detach()`
> * Then pass detached tensors to `create_dataarray_from_tensor`.

Thanks @sadamov for the review and guidance, I've gone ahead and pushed the requested changes:

Let me know if there's anything else that needs adjusting!

joeloskarsson:
In general we just plot things during inference (validation / test batches), so there are no gradients tracked and detach does nothing. But it does not hurt to do the detach, to be future-safe.

neural_lam/vis.py
```diff
@@ -182,3 +189,139 @@ def plot_spatial_error(
...
+        datastore.step_length
+    )
+
+    # Instantiating dataset outside the loop
+    weather_dataset = WeatherDataset(datastore=datastore, split=split)
+
+    for pred_slice, target_slice, time_slice in zip(
+        prediction_rescaled[:n_examples],
+        target_rescaled[:n_examples],
+        time_batch[:n_examples],
```
joeloskarsson: Rather than passing in `n_examples` and doing this, I think it would be better to just pass this function tensors that just contain the examples that should be plotted. So slicing things before calling this function.

neural_lam/vis.py
```diff
@@ -182,3 +189,139 @@ def plot_spatial_error(
...
+            os.path.join(logger.save_dir, pred_filename),
+        )
+
+        target_filename = f"example_target_{plotted_examples}.pt"
+        torch.save(
+            target_slice.cpu(),
+            os.path.join(logger.save_dir, target_filename),
+        )
+
+    return plotted_examples
```
joeloskarsson: I understand the idea of returning how many examples were plotted back, but I think that ARModel still needs to keep track of how many examples it has plotted itself, and keeping its own counter. Otherwise the `plot_examples` function becomes very specific and can really only work with the exact setup in ARModel, in which case the value of breaking out this function is lost.

neural_lam/vis.py
```diff
@@ -182,3 +189,139 @@ def plot_spatial_error(
...
+def plot_examples(
+    datastore: BaseRegularGridDatastore,
+    state_std: torch.Tensor,
+    state_mean: torch.Tensor,
+    logger: pl.loggers.Logger,
+    batch: tuple,
+    n_examples: int,
+    split: str,
+    prediction: torch.Tensor,
+    plotted_examples: int = 0,
```
joeloskarsson: Rather than `plotted_examples`, that relies on the fact that there is some counter somewhere, I think one could just pass in a number to be used as starting index. (Fills the same purpose, but makes the function less confusing when reading it without the context of ARModel).

neural_lam/vis.py
joeloskarsson: Perhaps one could feed the target and time separately, rather than the whole batch, if it is annoying to slice.

joeloskarsson (changes requested):
I think this is an improvement in the right direction. I do think it is important that the new `plot_examples` does not become something with such a narrow use-case that it really only makes sense to be called from a specific point in `ARModel`, as we then lost the point of the refactoring. I had some ideas on how to separate concerns to better avoid this, in particular to let `ARModel` still keep track of how many examples it wants to plot, and then just have `plot_examples` make plots for whatever data is thrown at it.

neural_lam/vis.py
oleeviyababu: Thanks @joeloskarsson , this is a great point regarding separation of concerns!

oleeviyababu: Hello @joeloskarsson 
I applied the mentioned changes.

neural_lam/vis.py
```diff
@@ -182,3 +189,127 @@ def plot_spatial_error(
...
+        state_mean: Mean of state variables for rescaling.
+        logger: The logger instance used to save the images.
+        split: The dataset split .
+        prediction: Output tensors predicted from the model.
+        target: Ground truth tensors.
+        time_batch: Time timestamps corresponding to the data.
+        start_index: The starting index number for naming saved files/logs.
+    """
+    prediction_rescaled = prediction * state_std + state_mean
+    target_rescaled = target * state_std + state_mean
```
sadamov: plot_examples should accept already-rescaled tensors so it stays agnostic to the normalisation scheme. rescaling with state_std/state_mean is a model concern and rather belongs at the call site in ARModel.

neural_lam/vis.py
```diff
@@ -182,3 +189,127 @@ def plot_spatial_error(
...
 
     return fig
+
+
+def plot_examples(
+    datastore: BaseRegularGridDatastore,
+    state_std: torch.Tensor,
+    state_mean: torch.Tensor,
+    logger: pl.loggers.Logger,
+    split: str,
```
sadamov: Instead of constructing a WeatherDataset internally from a split string, plot_examples should receive a pre-built instance as a parameter. WeatherDataset could be cached per test epoch instead of re-created per batch. constructed once in on_test_epoch_start, reused across batches, and released in on_test_epoch_end.

sadamov (changes requested):
I finally found time to look into this in more detail. Following Joel's comments I suggest to make the new function even more modular (see comments).
If you see better ways than I did to improve modularity even better!

The tests in test_plotting are now mostly redundant. Would be great to implement some new ones for plot_examples.

And finally don't forget to add your contribution to the changelog.

neural_lam/vis.py
joeloskarsson: Thanks for the changes, I think this looks a lot cleaner now :D

neural_lam/vis.py
joeloskarsson: Smart! that seems reasonable

neural_lam/vis.py
joeloskarsson: https://github.com/mllam/neural-lam/pull/224 is highly related to this. Maybe we can just address that issue directly here, and then don't need #224? The current solution in there needs some work (it is a bit tricky how to solve this).

In general, I don't think we should instantiate more than one dataset. The only reason we do it is to get access to `create_dataarray_from_tensor`, but this just points to the fact that that method sits in the wrong place.

neural_lam/vis.py
```diff
@@ -182,3 +189,127 @@ def plot_spatial_error(
...
+    prediction: torch.Tensor,
+    target: torch.Tensor,
+    time_batch: torch.Tensor,
+    start_index: int = 0,
```
joeloskarsson: Could we give this argument a more descriptive name?

oleeviyababu:
Thanks @sadamov and @joeloskarsson for the review! I've pushed a commit addressing the feedback:

Rescaling: I moved the * state_std + state_mean logic back to ARModel. plot_exampless now strictly accepts prescaled tensors 

Naming: Renamed start_index to first_example_idx for good clarity.

Changelog: Added an entry for this refactoring.

Regarding the WeatherDataset instantiation & Issue #224:
@joeloskarsson, I completely agree that the create_dataarray_from_tensor method is currently sitting in the wrong place. To solve this cleanly and address #224, I propose we extract create_dataarray_from_tensor from WeatherDataset and move it directly into BaseDatastore (or BaseRegularGridDatastore). This would allow us to remove the WeatherDataset import from vis.py entirely and we can just use the datastore object we are already passing in.

looking forward for your opinion

joeloskarsson:
I think moving create_dataarray_from_tensor directly to the datastore makes a lot of sense. I would push for that change. I also think it would be best to fix that within the scope of this PR, and then close #224.

@leifdenby do you know why we did not put this method directly in the datastore in the first place? You might have some good reason that I am missing here.

kshirajahere:
Hi @oleeviyababu @sadamov @joeloskarsson, I have been following #209.
Since the PR is still open with requested changes, and I do not see any follow-up activity after March 9, 2026 BY @oleeviyababu, I wanted to ask whether it would be helpful for me to take this over.

If @oleeviyababu is no longer planning to continue it, I would be happy to open a fresh PR that supersedes #209 and addresses the remaining review points, in particular:
- keeping `plot_examples` more modular and less ARModel-specific
- moving `create_dataarray_from_tensor` to the datastore as discussed here
- adding focused tests for the new plotting path
- cleaning up typing, docstrings, changelog, and pre-commit issues

If the original author is still planning to continue, I am also happy to stay off it.

oleeviyababu:
Hi team, I have just pushed the requested refactoring updates........

Here is a quick summary of what was fixed in this latest push:

Moved create_dataarray_from_tensor: Relocated this entirely to BaseDatastore, making the datastore the rightful owner of coordinate mapping and spatial grid logic.

Cleaned WeatherDataset: Removed the dataarray creation logic and the datetime import so the dataset is strictly responsible for PyTorch tensor batching.

Fixed Cyclic Dependency: Removed the WeatherDataset import from vis.py and updated plot_examples to call datastore.create_dataarray_from_tensor(...) directly.

Checks Passed: I ran pre-commit run --all-files and all formatting and type checks (Black, mypy, flake8) are passing cleanly.

Let me know if everything looks good or if you need any further changes

neural_lam/vis.py
joeloskarsson: Better now

neural_lam/vis.py
```diff
@@ -182,3 +188,116 @@ def plot_spatial_error(
...
+    ):
+        example_i = first_example_idx + i
+
+        # Detach tensors to safely separate from autograd graph
+        pred_slice = pred_slice.detach()
+        target_slice = target_slice.detach()
+
+        time_arr = np.array(time_slice.cpu(), dtype="datetime64[ns]")
+
+        # Call directly on datastore now!
```
joeloskarsson:
```suggestion
```
This feels like something that only makes sense within this refactoring, and not good to keep in the codebase.

joeloskarsson (changes requested):
To me this looks pretty much good to merge now. Just had one though on a comment.

We should plan the order of merging this and #208, as there will be merge conflicts.

oleeviyababu:
Thanks @joeloskarsson ,that comment definitely doesn't age well. I have removed it in the latest commit.

Regarding the merge order: I just saw your update on #208 that the dev team decided to merge #208 first for v0.7.0. That makes total sense to me!

Once it is merged, I will immediately pull the new main branch, rebase this PR (#209) to resolve the merge conflicts, and comment  for the final green light.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased](https://github.com/mllam/neural-lam/compare/v0.5.0...HEAD)
 
+### Refactored
```
sadamov: `### Refactored` is not a valid [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) 

```suggestion
### Changed
```

sadamov (changes requested):
This is looking really good overall and the refactoring has improved significantly over the review rounds. A few remaining issues before we can merge:

1. `create_dataarray_from_tensor` in `base.py` was loading the full dataset via `get_dataarray(split="test")` on every call just to read coordinate metadata. Fixed by using lightweight metadata methods instead (`get_vars_names`, `num_grid_points`, `get_xy`).
2. Missing return type annotation and duplicate comment in `base.py`. 
3. CHANGELOG section `### Refactored` is not a valid Keep a Changelog heading should be `### Changed`.
4. **Tests for `vis.plot_examples` are still missing**  plese consider adding tests like the following

```python
@pytest.mark.parametrize("n_examples", [1, 2])
@pytest.mark.parametrize("n_pred_steps", [1, 3])
def test_plot_examples_calls_logger_and_saves_pt_files(
    tmp_path, n_examples, n_pred_steps
):
    """Unit test for vis.plot_examples."""
    from unittest.mock import MagicMock
    import isodate

    datastore = DummyDatastore()
    n_state_features = datastore.get_num_data_vars("state")
    n_grid = datastore.num_grid_points

    prediction = torch.randn(n_examples, n_pred_steps, n_grid, n_state_features)
    target = torch.randn(n_examples, n_pred_steps, n_grid, n_state_features)

    t0_ns = int(isodate.parse_datetime("2021-01-01T00:00:00").timestamp() * 1e9)
    step_ns = int(datastore.step_length.total_seconds() * 1e9)
    time_batch = torch.tensor(
        [[t0_ns + i * step_ns + j * step_ns for j in range(n_pred_steps)]
         for i in range(n_examples)],
        dtype=torch.int64,
    )
    mock_logger = MagicMock()
    mock_logger.save_dir = str(tmp_path)

    vis.plot_examples(
        datastore=datastore, logger=mock_logger,
        prediction=prediction, target=target,
        time_batch=time_batch, first_example_idx=0,
    )

    expected_log_calls = n_examples * n_pred_steps * n_state_features
    assert mock_logger.log_image.call_count == expected_log_calls
    for i in range(n_examples):
        assert (tmp_path / f"example_pred_{i}.pt").exists()
        assert (tmp_path / f"example_target_{i}.pt").exists()
    plt.close("all")


def test_plot_examples_first_example_idx_offset(tmp_path):
    """Verify first_example_idx offsets saved filenames correctly."""
    from unittest.mock import MagicMock
    import isodate

    datastore = DummyDatastore()
    n_state_features = datastore.get_num_data_vars("state")
    n_grid = datastore.num_grid_points

    prediction = torch.randn(1, 1, n_grid, n_state_features)
    target = torch.randn(1, 1, n_grid, n_state_features)

    t0_ns = int(isodate.parse_datetime("2021-01-01T00:00:00").timestamp() * 1e9)
    step_ns = int(datastore.step_length.total_seconds() * 1e9)
    time_batch = torch.tensor([[t0_ns + step_ns]], dtype=torch.int64)

    mock_logger = MagicMock()
    mock_logger.save_dir = str(tmp_path)

    vis.plot_examples(
        datastore=datastore, logger=mock_logger,
        prediction=prediction, target=target,
        time_batch=time_batch, first_example_idx=3,
    )

    assert (tmp_path / "example_pred_3.pt").exists()
    assert not (tmp_path / "example_pred_0.pt").exists()
    plt.close("all")
```

Once these are addressed and the branch is rebased on `main` (post #208 merge), this should be good to go.

neural_lam/datastore/base.py
```diff
@@ -429,6 +431,85 @@ def expected_dim_order(
 
         return tuple(dim_order)
 
+    def create_dataarray_from_tensor(
+        self,
+        tensor: torch.Tensor,
+        time: Union[datetime.datetime, list[datetime.datetime]],
+        category: str,
+    ):
```
sadamov: Missing return type annotation

```suggestion
    ) -> xr.DataArray:
```

neural_lam/datastore/base.py
```diff
@@ -429,6 +431,85 @@ def expected_dim_order(
...
+                    "got a single time"
+                )
+        else:
+            raise ValueError(
+                "Expected tensor to have 2 or 3 dimensions, but got "
+                f"{len(tensor.shape)}"
+            )
+
+        # Get original dataarray to extract proper coordinates
+        # Get original dataarray to extract proper coordinates
```
sadamov: Duplicate comment — copy-paste artifact.

```suggestion
        # Get original dataarray to extract proper coordinates
```

neural_lam/datastore/base.py
```diff
@@ -429,6 +431,85 @@ def expected_dim_order(
...
+            if (
+                grid_coord in da_datastore_state.coords
+                and grid_coord not in da.coords
+            ):
+                da.coords[grid_coord] = da_datastore_state[grid_coord]
+
+        if not add_time_as_dim:
+            da.coords["time"] = time
+
+        return da
```
sadamov: **Performance regression** `get_dataarray(category=category, split="test")` loads the full dataset on every call just to extract coordinate metadata. The previous `WeatherDataset` implementation used a pre-cached `self.da_{category}`. Coordinates (feature names, grid indices, x/y) are split-independent and can be derived from lightweight methods already on `BaseDatastore`: `get_vars_names()`, `num_grid_points`, and `get_xy()`.

```suggestion
        feature_names = self.get_vars_names(category)
        grid_index = np.arange(self.num_grid_points)

        coords: dict = {
            f"{category}_feature": feature_names,
            "grid_index": grid_index,
        }
        if add_time_as_dim:
            coords["time"] = time

        da = xr.DataArray(
            tensor.cpu().numpy(),
            dims=dims,
            coords=coords,
        )

        # Attach x, y coordinates so the DataArray can later be unstacked
        # into a 2D spatial grid via unstack("grid_index")
        xy = self.get_xy(category, stacked=True)
        da.coords["x"] = xr.DataArray(xy[:, 0], dims=["grid_index"])
        da.coords["y"] = xr.DataArray(xy[:, 1], dims=["grid_index"])

        if not add_time_as_dim:
            da.coords["time"] = time

        return da
```

oleeviyababu:
Thanks for the detailed review @sadamov ,I have pushed a new commit that addresses all your points:

Updated CHANGELOG.md to use the valid ### Changed heading.

Fixed the missing return type annotation and removed the duplicate comment in base.py.

Implemented the performance fix in create_dataarray_from_tensor to use the lightweight metadata methods instead of loading the full test split.

Added the requested unit tests for vis.plot_examples with the correct indentation.

Let me know if everything looks good right now.

joeloskarsson:
#208 is merged now, so the conflicts can be resolved here now and we can see the final form of this :)

oleeviyababu:
> #208 is merged now, so the conflicts can be resolved here now and we can see the final form of this :)

Hi @joeloskarsson 
I have rebased the branch on top of the new main and resolved all the merge conflicts from #208. Everything is up to date. Let me know if everything looks good for the final merge,

Thankyou

neural_lam/datastore/base.py
```diff
@@ -460,6 +462,69 @@ def expected_dim_order(
...
+        tensor: torch.Tensor,
+        time: Union[datetime.datetime, list[datetime.datetime]],
+        category: str,
+    ) -> xr.DataArray:
+        """
+        Construct a xarray.DataArray from a `pytorch.Tensor` with coordinates
+        for `grid_index`, `time` and `{category}_feature` matching the shape
+        and number of times provided and add the x/y coordinates from the
+        datastore.
+        """
```
joeloskarsson, line 496: Could we change this to a full numpy-style docstring? Trying to be more consistent with this from now on, as much of the codebase now nicely follows this.

neural_lam/vis.py
```diff
@@ -639,3 +645,115 @@ def plot_spatial_error(
...
+    Plot example forecasts from provided tensors.
+
+    Args:
+        datastore: The object containing dataset metadata.
+        logger: The logger instance used to save the images.
+        prediction: Output tensors predicted from the model.
+        target: Ground truth tensors.
+        time_batch: Time timestamps corresponding to the data.
+        first_example_idx: Starting index for naming saved files/logs.
+    """
```
joeloskarsson, line 815: Similar comment about numpy-style docstring. Here I think it is mostly about formatting.

neural_lam/vis.py
```diff
@@ -20,6 +21,11 @@
 _LABEL_SIZE = 11  # axis / colorbar labels
 _TICK_SIZE = 11  # tick labels
 
+import os
+import warnings
+import torch
+import pytorch_lightning as pl
+
```
joeloskarsson: Place imports at top

neural_lam/vis.py
```diff
@@ -639,3 +645,115 @@ def plot_spatial_error(
...
+
+def plot_examples(
+    datastore: BaseRegularGridDatastore,
+    logger: pl.loggers.Logger,
+    prediction: torch.Tensor,
+    target: torch.Tensor,
+    time_batch: torch.Tensor,
+    first_example_idx: int = 0,
+) -> None:
```
joeloskarsson, line 791: Hmm, I think something might have got lost in resolving the conflicts here now. Nothing calls this function now right? There should be some change on the model size I suppose?

CHANGELOG.md
```diff
@@ -124,6 +124,10 @@ This release introduces new features including GIF animation support, wandb run
 
 - Fix Slack domain link [\#288](https://github.com/mllam/neural-lam/pull/288) @sadamov
 
+### Changed
+
+- Refactored `ARModel`: Extracted plotting logic into stateless functions in `vis.py` [\#209](https://github.com/mllam/neural-lam/pull/209) @oleeviyababu
```
joeloskarsson: I think this is out of date? The actual model-side refactoring seems to be missing atm.

neural_lam/datastore/base.py
joeloskarsson: I think also some additional explanantion is needed here about the relationship between the assumed tensor dimensions and the time argument. I think the old docstring had this.

joeloskarsson (changes requested):
Thanks for updating this. Had a look over the current code.

joeloskarsson:
Is there still interest to continue work on this @oleeviyababu ? I still find this to be a big improvement, and the community seems to agree.

oleeviyababu:
> Is there still interest to continue work on this @oleeviyababu ? I still find this to be a big improvement, and the community seems to agree.

Hello Joel, Sorry for the late response, I was in my home country for the last 2 months, I will complete the work soon, Thankyou for the message and the understanding

oleeviyababu:
Hi @joeloskarsson , thanks for the review. I've pushed a commit addressing all the comments:

 -Rewrote the `create_dataarray_from_tensor` docstring in NumPy style and restored the tensor/time dimension explanation.
 -Converted the `plot_examples` docstring to NumPy style and cleaned up imports.
 -Restored the model-side refactoring so `ForecasterModule.plot_examples` only handles rescaling and delegates plotting to `vis.plot_examples`, with GIF creation moved there.
 -Updated the changelog.

Let me know if there's anything else to get adjust

joeloskarsson:
Thanks, could you also resolve the current conflicts with main before I have another look? Better to review PRs as they are up to date, as we might otherwise discuss things that get changed in conflict resolution.

oleeviyababu:
Hello @joeloskarsson 

Rebased on the latest main and resolved the conflicts in weather_dataset.py and module.py. I also removed the now-dead _create_dataarray_from_tensor helper and its unused imports (WeatherDataset, xarray) from module.py, since plotting is now fully delegated to vis.plot_examples. All prcommit hooks pass locally. 

Thankyou

neural_lam/vis.py
```diff
@@ -775,3 +778,164 @@ def plot_spatial_error(
...
+    Parameters
+    ----------
+    datastore : BaseRegularGridDatastore
+        The object containing dataset metadata.
+    logger : pl.loggers.Logger
+        The logger instance used to save the images.
+    prediction : torch.Tensor
+        Output tensors predicted from the model.
+    target : torch.Tensor
+        Ground truth tensors.
```
joeloskarsson: Write out shape of these. I think there is a specific shape required here.

joeloskarsson (changes requested):
Thanks for updating this, only one small thing from me, then I think this is good to be merged. I'll let @sadamov go over and resolve his comments based on the updated code.

neural_lam/vis.py
oleeviyababu: > Write out shape of these. I think there is a specific shape required here.

I have added the shapes

joeloskarsson:
Code change looks all good to me now, but seems there are also some tests that needs updating w.r.t. this change.

KumarShivam1908:
if this Issue is still active , i would like to take this up.
Thanks

sadamov:
@oleeviyababu if you have nothing against it @KumarShivam1908 can take over from here

oleeviyababu:
Hi @joeloskarsson @sadamov, sorry for the long silencve. I moved back
to Germany and got with PhD work. I am getting back on top of things
now and pushed the work , the failing tests are fixed and i merged current main in, so the
conflicts are gone.

One thing is  it wasn't only the tests that needed updating.
When create_dataarray_from_tensor moved to BaseDatastore it started building
grid_index with np.arange instead of carrying over the datastore's spatial
multi index.   vis.plot_examples still called .unstack("grid_index") on that,
which blows up with:

ValueError: cannot unstack dimensions that do not have exactly one
    multi-index: ('grid_index',) 

So plotting was actually broken at runtime,. That
was 6 of the 17 CI failures. The other 11 were the two test files still
calling WeatherDataset.create_dataarray_from_tensor.

What i changed:
1. vis.py now unstacks with datastore.unstack_grid_coords() instead of
  .unstack("grid_index")
2. both test files go through the datastore now and the extra WeatherDataset
  instantiation is gone
3. base.py: time also accepts np.ndarray

Two small doubts: could someone approve the workflow run so CI can actually
run on this? And a re-review for  the change?

@KumarShivam1908 thanks for offering to pick this up that was kind. I am
back on it now but also happy to work together on it too.

Issue #63, opened 2024-06-18
Merge global forecasting capabilities from `prob_model_global`


joeloskarsson:
Global forecasting can be viewed as a special case of LAM, with no boundary conditions to consider. Therefore it is not a hard task to allow for global forecasting within the Neural-LAM framework. This has the benefit of being able to use any models in here also for global forecasting.

Global forecasting, as done in [Oskarsson et al. (2024)](https://arxiv.org/abs/2406.04759), is currently implemented on the branch `prob_model_global`. This issue is to keep track of merging of the `prob_model_global` branch into main.

The main changes in this branch include:

- [ ] Add script for creating of global forcing https://github.com/mllam/neural-lam/blob/f68a5132a1d6aafcd5695705762d08ce29cbb777/create_global_forcing.py#L26 This will likely be taken care of through #34, rather than merging `create_global_forcing.py` directly.
- [ ] Add script for creating global static grid features https://github.com/mllam/neural-lam/blob/f68a5132a1d6aafcd5695705762d08ce29cbb777/create_global_grid_features.py#L23 Also likely better handled in #34.
- [ ] Add script for creating global mesh graphs (multiscale and hierarchical) https://github.com/mllam/neural-lam/blob/f68a5132a1d6aafcd5695705762d08ce29cbb777/create_global_mesh.py#L128 Will see if this should sit in this repo or in [weather-model-graphs](https://github.com/mllam/weather-model-graphs), and to what extent we keep the graphcast dependency for graph generation.
- [ ] Allow for 3d graph plotting of global graphs https://github.com/mllam/neural-lam/blob/f68a5132a1d6aafcd5695705762d08ce29cbb777/plot_global_graph.py#L99 It is perhaps better to first move the 3d plotting functionality to [weather-model-graphs](https://github.com/mllam/weather-model-graphs) and do this there. See point above.
- [ ] Enable generating variable weights based on pressure levels https://github.com/mllam/neural-lam/blob/f68a5132a1d6aafcd5695705762d08ce29cbb777/create_parameter_weights.py#L53 This is better handled by defining more general variable weighting setup where one can either set these weights manually per-variable in config files or use some pre-defined strategies.
[...]

PR #211, opened 2026-02-21
Make boundary forcing optional to support global forecasting

nehamanoj1105:
Description

This PR makes boundary forcing optional in ARModel, enabling support for global forecasting where no boundary conditions are present.

As discussed in Issue #63, global forecasting can be viewed as a special case of LAM without boundary overwriting. Previously, boundary masking was always applied during autoregressive rollout, which prevented clean support for global datasets.

Changes

Made boundary_mask optional during initialisation.

Introduced self.use_boundary_forcing flag.

Updated unroll_prediction() to:

Apply boundary overwrite only when the boundary mask exists.

Skip overwrite entirely for global forecasting.

Ensured loss and metrics behaviour remains unchanged:

LAM: loss computed over interior only.

Global: loss computed over the full grid.

### DIFF
--- a/neural_lam/models/ar_model.py
+++ b/neural_lam/models/ar_model.py
@@ -124,17 +124,26 @@ class ARModel(pl.LightningModule):
         # Instantiate loss function
         self.loss = metrics.get_metric(args.loss)
 
-        boundary_mask = torch.tensor(
-            da_boundary_mask.values, dtype=torch.float32
-        ).unsqueeze(
-            1
-        )  # add feature dim
+        if da_boundary_mask is not None:
+            boundary_mask = torch.tensor(
+                da_boundary_mask.values, dtype=torch.float32
+            ).unsqueeze(1)
 
-        self.register_buffer("boundary_mask", boundary_mask, persistent=False)
-        # Pre-compute interior mask for use in loss function
-        self.register_buffer(
-            "interior_mask", 1.0 - self.boundary_mask, persistent=False
-        )  # (num_grid_nodes, 1), 1 for non-border
+            self.register_buffer(
+                "boundary_mask", boundary_mask, persistent=False
+            )
+            self.register_buffer(
+                "interior_mask", 1.0 - boundary_mask, persistent=False
+            )
+            self.use_boundary_forcing = True
+        else:
+            self.boundary_mask = None
+            self.register_buffer(
+                "interior_mask",
+                torch.ones(self.num_grid_nodes, 1),
+                persistent=False,
+            )
+            self.use_boundary_forcing = False
 
         self.val_metrics: Dict[str, List] = {
             "mse": [],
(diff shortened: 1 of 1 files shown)


InFiNiTy0639 (commented):
I noticed you've added handling for when da_boundary_mask is None and introduced the use_boundary_forcing flag.  Could you briefly clarify the downstream use case motivating this change? Is this specifically to support global datasets, or are there other models/datasets we're planning to introduce that don't require boundary forcing?

nehamanoj1105:
The primary motivation is to support global forecasting as discussed in Issue #63. In the global setting, the domain covers the entire sphere and therefore has no physical boundary where true states should be injected during rollout. In that case, boundary overwriting is not applicable, and the model should evolve autoregressively over the full grid.

At the moment, the intended downstream use case is global datasets (e.g., ERA5-style global forecasting). The change is structured generically so that any future dataset without boundary conditions can also use the same logic without modification.

esmacimsit:
One thing I wanted to double-check:
since self.boundary_mask can now be None, are there any other methods in ARModel (or inherited classes) that access self.boundary_mask directly without guarding on self.use_boundary_forcing?

I checked unroll_prediction() which is properly protected, but I just want to ensure there are no downstream assumptions that boundary_mask always exists.

Also, do you think adding a minimal unit test covering the da_boundary_mask=None case would be useful to avoid regressions?

nehamanoj1105:
I reviewed the rest of ARModel, and boundary_mask is only used inside unroll_prediction(), which is now guarded via self.use_boundary_forcing. Other components (loss, metrics, masking) rely on interior_mask, which is always defined, including in the global case where it defaults to all ones.

So there should not be any downstream assumptions that boundary_mask is always present.

Regarding testing, yes, I agree that adding a minimal unit test for the da_boundary_mask=None case would be useful to prevent regressions. I can add a small test that initialises ARModel with boundary_mask=None and verifies that unroll_prediction() runs without applying overwrite logic.

neural_lam/models/ar_model.py
```diff
@@ -229,50 +238,55 @@ def predict_step(self, prev_state, prev_prev_state, forcing):
...
-        )  # (B, pred_steps, num_grid_nodes, d_f)
+        prediction = torch.stack(prediction_list, dim=1)
+
         if self.output_std:
-            pred_std = torch.stack(
-                pred_std_list, dim=1
-            )  # (B, pred_steps, num_grid_nodes, d_f)
+            pred_std = torch.stack(pred_std_list, dim=1)
         else:
-            pred_std = self.per_var_std  # (d_f,)
```
joeloskarsson, line 275: What is the motivation for these changes? They seem unrelated

neural_lam/models/ar_model.py
```diff
@@ -124,17 +124,26 @@ def __init__(
...
-        ).unsqueeze(
-            1
-        )  # add feature dim
-
-        self.register_buffer("boundary_mask", boundary_mask, persistent=False)
-        # Pre-compute interior mask for use in loss function
-        self.register_buffer(
-            "interior_mask", 1.0 - self.boundary_mask, persistent=False
-        )  # (num_grid_nodes, 1), 1 for non-border
+        if da_boundary_mask is not None:
```
joeloskarsson, line 127: This might seem like a reasonable check to determine if we should use boundary forcing or not, however right now there is not DataStore that does not have a boundary mask, so we could not actually run the non-boundary-forcing case here. How is the idea that this would be used?

joeloskarsson (changes requested):
Thanks for starting to look at this. I mostly just want some clarifications here, before diving deeper.

sadamov:
@nehamanoj1105 is there still interest in working on this PR from your side or could someone else take over here?

kshirajahere:
@sadamov can I volunteer to work on this?

sadamov:
@kshirajahere I saw that @joeloskarsson and other gsoc-mentors are currently waiting on the proposal review process before assigning issues that are covered by the gsoc projects. It makes sense that the selected applicants would be involved in solving these issues. So we should probably stand by for a bit here.

And most likely any solution to #90 will introduce optional boundary forcing, thereby solving this PR.

sadamov:
Closing this one. The GSoC project on global forecasting (issue #63) did not materialise. Proper boundary handling is now being introduced via #635 (optional boundary datastore), which makes the `use_boundary_forcing` flag here redundant. Thanks @nehamanoj1105 for the contribution.

PR #220, opened 2026-02-22
fix: add missing bounds check in test_step to prevent IndexError

### LINKED ISSUE #219: [Bug] `IndexError` — Out-of-Range Step in `ARModel.test_step` Log Dict
**File:** `ar_model.py`

The log dict comprehension in `test_step` that records per-step test losses does not check whether `step` is within range of `time_step_loss`. If `val_steps_to_log` contains a step larger than `ar_steps_eval`, this raises an `IndexError`.
```python
# WRONG — no bounds check
{f"test_loss_step_{step}": time_step_loss[step] for step in val_steps_to_log}

# Should be
{f"test_loss_step_{step}": time_step_loss[step] for step in val_steps_to_log if step < len(time_step_loss)}
```

This will crash silently during evaluation if `val_steps_to_log` is configured with steps beyond `ar_steps_eval`.

santhil-cyber:
##  Changes

`test_step` in `ar_model.py` was missing the bounds check that `validation_step` already has. If `val_steps_to_log` contains a step larger than the number of prediction steps, it causes an `IndexError`.

Added the same `if step < len(time_step_loss)` guard that already exists in `validation_step` — one-line change, no new dependencies.

**Change:**
```diff
# Before
- {f"test_loss_step_{step}": time_step_loss[step] for step in val_steps_to_log}
# After
+ {f"test_loss_step_{step}": time_step_loss[step] for step in val_steps_to_log if step < len(time_step_loss)}
```

---

## Issue Link

Closes #219

---

## Type of Change

- [ ] 📖 Documentation (addition or improvements to documentation)

---

## Checklist Before Requesting a Review

---

## Checklist for Reviewers

- [x] The code is readable
- [x] The code is well tested
- [x] The code is documented (including return types and parameters)
- [x] The code is easy to maintain

---

## Author Checklist After Completed Review

- [x] I have added a line to the CHANGELOG describing this change:
  - `fixes: Add missing bounds check in test_step to match validation_step behavior`

---

## Checklist for Assignee

- [x] The tests pass
- [x] Author has added an entry to the changelog
- [x] Once the PR is ready to be merged, squash commits and merge the PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Resolve `xarray` `FacetGrid` `DeprecationWarning` in `plot_example.py` by using a compatibility shim [\#482](https://github.com/mllam/neural-lam/pull/482) @sohampatil01-svg
 
 - Add missing bounds check in `test_step` to prevent `IndexError` when `val_steps_to_log` exceeds prediction steps [\#220](https://github.com/mllam/neural-lam/pull/220) @santhil-cyber
+
 ### Maintenance
 
 - Add comprehensive type hints to all functions and class methods in `utils.py` [\#620](https://github.com/mllam/neural-lam/pull/620) @GiGiKoneti
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,8 @@ This release contains maintenance and fixes, preventing some unexpected crashes 
 
 - Adding a more robust LaTeX availability check function [\#162](https://github.com/mllam/neural-lam/pull/162) @lorenzo30salgado
 
+- Add missing bounds check in `test_step` to prevent `IndexError` when `val_steps_to_log` exceeds prediction steps [\#219](https://github.com/mllam/neural-lam/pull/219) @santhil-cyber
+
 ### Maintenance
 
 - Add link to full MEPS data [\#102](https://github.com/mllam/neural-lam/pull/102) @joeloskarsson
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Infer spatial coordinate names for MDPDatastore (rather than assuming names `x` and `y`), allows for e.g. lat/lon regular grids [\#169](https://github.com/mllam/neural-lam/pull/169) @leifdenby
 
+- Add missing bounds check in `test_step` to prevent `IndexError` when `val_steps_to_log` exceeds prediction steps [\#219](https://github.com/mllam/neural-lam/pull/219) @santhil-cyber
+
 ### Maintenance
 
 - Update PR template to clarify milestone/roadmap requirement and maintenance changes [\#186](https://github.com/mllam/neural-lam/pull/186) @joeloskarsson
@@ -92,8 +94,6 @@ This release contains maintenance and fixes, preventing some unexpected crashes 
 [\#150](https://github.com/mllam/neural-lam/pull/150) @YUTAIPAN
 
 - Adding a more robust LaTeX availability check function [\#162](https://github.com/mllam/neural-lam/pull/162) @lorenzo30salgado
-
-- Add missing bounds check in `test_step` to prevent `IndexError` when `val_steps_to_log` exceeds prediction steps [\#219](https://github.com/mllam/neural-lam/pull/219) @santhil-cyber
 
 ### Maintenance
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,30 +5,62 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [unreleased](https://github.com/mllam/neural-lam/compare/v0.5.0...HEAD)
-
-### Added
-
-- Enable `pin_memory` in DataLoaders when GPU is available for faster async CPU-to-GPU data transfers [\#236](https://github.com/mllam/neural-lam/pull/236) @abhaygoudannavar
-
-### Changed
-- Refactor graph loading: move zero-indexing out of the model and update plotting to prepare using the research-branch graph I/O [\#184](https://github.com/mllam/neural-lam/pull/184) @zweihuehner
-
-- Replace `print()`-based `rank_zero_print` with `loguru` `logger.info()` for structured log-level control ([#33](https://github.com/mllam/neural-lam/issues/33))
+## [unreleased](https://github.com/mllam/neural-lam/compare/v0.6.0...HEAD)
 
 ### Fixed
 
+- Fix `AssertionError` in `aggregate_and_plot_metrics` when using `--metrics_watch` flags by using `isinstance` dispatch for figure vs scalar logging [\#303](https://github.com/mllam/neural-lam/pull/303) @AftAb-25
+
+- Resolve `xarray` `FacetGrid` `DeprecationWarning` in `plot_example.py` by using a compatibility shim [\#482](https://github.com/mllam/neural-lam/pull/482) @sohampatil01-svg
+
 - Add missing bounds check in `test_step` to prevent `IndexError` when `val_steps_to_log` exceeds prediction steps [\#220](https://github.com/mllam/neural-lam/pull/220) @santhil-cyber
+
+## [v0.6.0](https://github.com/mllam/neural-lam/releases/tag/v0.6.0)
+
+This release introduces new features including GIF animation support, wandb run resumption, and improved ensemble loading, alongside a large number of bug fixes and maintenance updates.
+
+### Added
+
+- Add support for GIF animation generation for model predictions [\#218](https://github.com/mllam/neural-lam/pull/218) @kartikangiras
+
+- Add `AGENTS.md` file to the repo to give agents more information about the codebase and the contribution culture.[\#416](https://github.com/mllam/neural-lam/pull/416) @sadamov
+
+- Enable `pin_memory` in DataLoaders when GPU is available for faster async CPU-to-GPU data transfers [\#236](https://github.com/mllam/neural-lam/pull/236) @abhaygoudannavar
+
+- Expose `--wandb_id` CLI argument to allow resuming an existing W&B run by
+  ID. When provided, `resume="allow"` is set automatically so the same job
+  script works for both the initial submission and all resubmissions, making
+  it suitable for HPC systems with limited job runtimes or that may crash.
+  [\#197](https://github.com/mllam/neural-lam/pull/197) @Mani212005
+
+### Changed
+
+- Change the default ensemble-loading behavior in `WeatherDataset` / `WeatherDataModule` to use all ensemble members as independent samples for ensemble datastores (with matching ensemble-member selection for forcing when available); single-member behavior now requires explicitly opting in via `--load_single_member` [\#332](https://github.com/mllam/neural-lam/pull/332) @kshirajahere
+
+- Refactor graph loading: move zero-indexing out of the model and update plotting to prepare using the research-branch graph I/O [\#184](https://github.com/mllam/neural-lam/pull/184) @zweihuehner
+
+- Replace `print()`-based `rank_zero_print` with `loguru` `logger.info()` for structured log-level control [\#33](https://github.com/mllam/neural-lam/issues/33)
+
+- Switch to lat/lon-based plotting with `pcolormesh` and `cartopy` for accurate spatial visualisation regardless of underlying projection. [\#168](https://github.com/mllam/neural-lam/pull/168) @sadamov
+
+- Infer spatial coordinate names for MDPDatastore (rather than assuming names `x` and `y`), allows for e.g. lat/lon regular grids [\#169](https://github.com/mllam/neural-lam/pull/169) @leifdenby
+
+### Fixed
+
+- Fix validation crash in `plot_error_map` and resolve DDP NCCL initialization error on single-device setups [\#193](https://github.com/mllam/neural-lam/pull/193) @AdityaKumarSethia
+
+- Fix `--metrics_watch` handling to avoid AttributeError when unset and improve warning behavior during evaluation [#420](https://github.com/mllam/neural-lam/pull/420) @archit7-beep
+
+- Standardize all script references to use `create_graph` instead of the legacy `create_mesh` name in README and `pyproject.toml`, and fix minor README typos [\#426](https://github.com/mllam/neural-lam/pull/426) @GiGiKoneti
 
 - Initialize `da_forcing_mean` and `da_forcing_std` to `None` when forcing data is absent, fixing `AttributeError` in `WeatherDataset` with `standardize=True` [\#369](https://github.com/mllam/neural-lam/issues/369) @Sir-Sloth-The-Lazy
 
 - Ensure proper sorting of `analysis_time` in `NpyFilesDatastoreMEPS._get_analysis_times` independent of the order in which files are processed with glob [\#386](https://github.com/mllam/neural-lam/pull/386) @Gopisokk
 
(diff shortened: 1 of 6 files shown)


sadamov:
thanks @santhil-cyber that is true and this silent runtime error has annoyed me on multiple occasions. Certainly something we should fix and I agree that there is no reason why val and test should be treated differently. Please add the following:
- A [fixed] entry in the CHANGELOG.md
- A warning emitted by the logger to notify the user about too many steps in `val_steps_to_log`. For now you can use the already imported `warning` module for that

santhil-cyber:
@sadamov made these changes...have a look once.

sadamov:
Okay I tested this again and found some inconsistencies that we should address:

- Fix unguarded spatial loss indexing at `ar_model.py#L455`, which suffers from the same problem and would still crash `eval test`
  - Update both `zip(self.args.val_steps_to_log, ...)` calls in `on_test_epoch_end` at `L727` and `L750` to use the filtered step list, otherwise plot titles and filenames mismatch the spatial loss maps
- Add the same `UserWarning` to `validation_step` for consistency (currently only added to `test_step`)
- The warning would currently be logged every batch which would flood the logs. Please only emit them once for each val/test per training/eval.

santhil-cyber:
@sadamov Thanks for flagging these,  I had actually already identified these inconsistencies during my recent analysis and was about to push the updated code .... now have pushed the fixes in this latest commit. Here's a summary of what's covered:

1. **Spatial loss indexing (L454)** — Guarded with the same `step <= spatial_loss.shape[1]` filter so eval won't crash on out-of-range steps.

2. **`on_test_epoch_end` zip calls (L726, L750)** — Both now use a filtered step list so plot titles and PDF filenames stay aligned with the actual spatial loss maps.

3. **`validation_step` warning** — Added the same `UserWarning` for consistency with `test_step`, so users get notified regardless of whether they're running validation or testing.

4. **Warn-once pattern** — Introduced a `self._warned_val_steps` flag in `__init__` that gets set on the first occurrence. This ensures the warning fires exactly once per training/eval run instead of flooding logs every batch.

I have Tested these changes you may verify once too. Let me know if there's anything else you'd like adjusted!

sadamov:
@santhil-cyber when working with the CHANGELOG.md, please add your entries to the **unreleased** section and not the already released version 0.5.0

neural_lam/models/ar_model.py
```diff
@@ -346,6 +350,21 @@ def validation_step(self, batch, batch_idx):
...
+            ]
+            if skipped_steps:
+                warnings.warn(
+                    f"val_steps_to_log contains steps {skipped_steps} "
+                    f"that exceed the number of prediction steps "
+                    f"({len(time_step_loss)}). "
+                    "These steps will be skipped.",
+                    UserWarning,
+                )
+                self._warned_val_steps = True
```
sadamov:
```suggestion
            self._warned_val_steps = True
```
Remove one indent-level, to prevent unnecessary repeat calculation

neural_lam/models/ar_model.py
```diff
@@ -400,9 +419,25 @@ def test_step(self, batch, batch_idx):
...
```
sadamov:
```suggestion
            self._warned_val_steps = True
```

neural_lam/models/ar_model.py
```diff
@@ -712,7 +749,12 @@ def on_test_epoch_end(self):
                     f"({(self.time_step_int * t_i)} {self.time_step_int_unit})",
                 )
                 for t_i, loss_map in zip(
-                    self.args.val_steps_to_log, mean_spatial_loss
+                    [
+                        s
+                        for s in self.args.val_steps_to_log
+                        if s <= mean_spatial_loss.shape[0]
+                    ],
+                    mean_spatial_loss,
```
sadamov: I was actually wrong here, the internal logic of zip works with the reduced,  ordered step-sample. 
```suggestion
                    self.args.val_steps_to_log, mean_spatial_loss
```

neural_lam/models/ar_model.py
```diff
@@ -735,7 +777,12 @@ def on_test_epoch_end(self):
                 self.logger.save_dir, "spatial_loss_maps"
             )
             os.makedirs(pdf_loss_maps_dir, exist_ok=True)
-            for t_i, fig in zip(self.args.val_steps_to_log, pdf_loss_map_figs):
+            filtered_steps = [
+                s
+                for s in self.args.val_steps_to_log
+                if s <= len(pdf_loss_map_figs)
+            ]
+            for t_i, fig in zip(filtered_steps, pdf_loss_map_figs):
```
sadamov:
```suggestion
            for t_i, fig in zip(self.args.val_steps_to_log, pdf_loss_map_figs):
```
also here the original code is still working fine with zip and the ordered object.

CHANGELOG.md
```diff
@@ -52,6 +52,8 @@ This release contains maintenance and fixes, preventing some unexpected crashes
 
 - Adding a more robust LaTeX availability check function [\#162](https://github.com/mllam/neural-lam/pull/162) @lorenzo30salgado
 
+- Add missing bounds check in `test_step` to prevent `IndexError` when `val_steps_to_log` exceeds prediction steps [\#219](https://github.com/mllam/neural-lam/pull/219) @santhil-cyber
+
```
sadamov: move this to the unreleased section

sadamov (changes requested):
It is worth noting that when you do **instantiate the model programmatically** (not via CLI), val_steps_to_log and ..._watch are already properly checked: https://github.com/santhil-cyber/neural-lam/blob/e83874acf49a53c1519304c25cf533b777019153/neural_lam/train_model.py#L232-L247

This PR is adding a missing back-up check that is affecting direct instatiation of ARModel for example, as done in some tests of the test-suite.

If you agree you can accept tmy small changes and then we merge.

santhil-cyber:
@sadamov Thanks so much! Applied all your suggestions:

- Moved `_warned_val_steps = True` outside the `if skipped_steps:` block to avoid redundant recalculation
- Reverted the `zip()` filtering, you're right, `zip` already handles the shorter iterable naturally
- Moved CHANGELOG entry to the unreleased section

All 110 tests pass. Ready for re-review! 

not related but can you share the active slack group link

joeloskarsson:
Look like there are som linting errors @santhil-cyber . Pls have a look

santhil-cyber:
@joeloskarsson Hey!! can you try merging it now

sadamov:
@santhil-cyber apologies, I did not see this PR in my list of open PRs because I forgot to re-request a review. I will have a look soon and merge!

santhil-cyber:
No issue @sadamov ...lemme know if ant changes shall need to be implemented

santhil-cyber:
Hey any updates @sadamov

sadamov:
@santhil-cyber I'll try to look at this before the dev meeting on Monday. There is just so much going on these days

santhil-cyber:
no issue @sadamov ...take your time, I can just imagine number of interactions you might be going through in these ending stage of GSoC application period.

CHANGELOG.md
```diff
@@ -49,6 +49,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 
+- Add missing bounds check in `test_step` to prevent `IndexError` when `val_steps_to_log` exceeds prediction steps [\#219](https://github.com/mllam/neural-lam/pull/219) @santhil-cyber
+
```
sadamov: put this under ### Fixed as it is not a new feature but a bugfix. please link #220 instead of 219

neural_lam/models/ar_model.py
```diff
@@ -346,6 +350,22 @@ def validation_step(self, batch, batch_idx):
...
+            if skipped_steps:
+                warnings.warn(
+                    f"val_steps_to_log contains steps {skipped_steps} "
+                    f"that exceed the number of prediction steps "
+                    f"({len(time_step_loss)}). "
+                    "These steps will be skipped.",
+                    UserWarning,
+                )
+            self._warned_val_steps = True
+
```
sadamov: extract this into a help function in the ar_models __init__ to avoid code duplication between the test and val warnings.

sadamov (commented):
@santhil-cyber so I finally got time to review again, and I think we are almost there. I keep changing my mind about the ordering of val steps (user input) and how zip() would work with reduced number of steps. sorry about that flipflopping, I'm aware it's possibly annoying. feel free to also let me know if you disagree with my curret proposal. here are my suggestions:

neural_lam/models/ar_model.py
```diff
@@ -156,6 +156,10 @@ def __init__(
         # For storing spatial loss maps during evaluation
         self.spatial_loss_maps: List[Any] = []
 
+        # Track whether we've already warned about out-of-range
+        # val_steps_to_log to avoid flooding logs every batch
+        self._warned_val_steps = False
+
```
sadamov: Now with the duplicated warning I actually realized that we should not warn in per batch at all but rather warn once in `__init__`, not per-batch with a flag

`len(time_step_loss)` always equals `args.ar_steps_eval` — it is constant for the model’s lifetime, not batch-dependent. The warn-once flag is working around the real fix: check at construction time and store the filtered list for reuse everywhere (in `validation_step`, `test_step`, and `on_test_epoch_end`).

```suggestion
        # Filter val_steps_to_log to valid steps and emit one warning at
        # construction time
        invalid_steps = [
            s for s in args.val_steps_to_log if s > args.ar_steps_eval
        ]
        if invalid_steps:
            warnings.warn(
                f"val_steps_to_log contains steps {invalid_steps} that exceed "
                f"ar_steps_eval ({args.ar_steps_eval}). These steps will be "
                "skipped.",
                UserWarning,
            )
        self.valid_steps_to_log = [
            s for s in args.val_steps_to_log if s <= args.ar_steps_eval
        ]
```

With this, the `_warned_val_steps` flag and the duplicate warning blocks in `validation_step` and `test_step` can be removed, and every use of `self.args.val_steps_to_log` in the model replaced with `self.valid_steps_to_log`.

neural_lam/models/ar_model.py
sadamov: With the warning moved to `__init__`, remove this block. Also replace `self.args.val_steps_to_log` → `self.valid_steps_to_log` on line 371, and drop the `if step <= len(time_step_loss)` filter (already handled at construction time).

```suggestion

```

neural_lam/models/ar_model.py
```diff
@@ -400,9 +420,26 @@ def test_step(self, batch, batch_idx):
...
+            if skipped_steps:
+                warnings.warn(
+                    f"val_steps_to_log contains steps {skipped_steps} "
+                    f"that exceed the number of prediction steps "
+                    f"({len(time_step_loss)}). "
+                    "These steps will be skipped.",
+                    UserWarning,
+                )
+            self._warned_val_steps = True
+
```
sadamov: Same — remove this block. Replace `self.args.val_steps_to_log` → `self.valid_steps_to_log` on lines 441 and 481, and drop the `if` filters on those lines (filtering is already done in `__init__`).

```suggestion

```

sadamov (commented):
I am changing my mind again about the zip here. I actually think we need to make it more robust, still. 

The “zip truncates naturally” argument only holds when out-of-range steps are always sorted to the tail. It breaks for any other ordering.

Once `self.valid_steps_to_log` is introduced in `__init__` (see inline suggestion), fix both zip calls and simplify `valid_step_indices` in `test_step`:

sadamov:
Follow-up to the inline suggestions above — the remaining five places that need `self.args.val_steps_to_log` → `self.valid_steps_to_log` (with `if` guards dropped, since filtering is done once in `__init__`):

**`validation_step` — `val_log_dict` (~line 369)**
```python
val_log_dict = {
    f"val_loss_unroll{step}": time_step_loss[step - 1]
    for step in self.valid_steps_to_log
}
```

**`test_step` — `test_log_dict` (~line 439)**
```python
test_log_dict = {
    f"test_loss_unroll{step}": time_step_loss[step - 1]
    for step in self.valid_steps_to_log
}
```

**`test_step` — `valid_step_indices` (~line 479)**
```python
valid_step_indices = [s - 1 for s in self.valid_steps_to_log]
```

**`on_test_epoch_end` — title zip (~line 755)**
```python
for t_i, loss_map in zip(self.valid_steps_to_log, mean_spatial_loss):
```

**`on_test_epoch_end` — PDF filename zip (~line 779)**
```python
for t_i, fig in zip(self.valid_steps_to_log, pdf_loss_map_figs):
```

After all these changes, there should be zero remaining references to `self.args.val_steps_to_log` in the model (only `args.val_steps_to_log` in `__init__` itself where the filtered list is built).

santhil-cyber:
> @santhil-cyber so I finally got time to review again, and I think we are almost there. I keep changing my mind about the ordering of val steps (user input) and how zip() would work with reduced number of steps. sorry about that flipflopping, I'm aware it's possibly annoying. feel free to also let me know if you disagree with my curret proposal. here are my suggestions:

Hey @sadamov, no worries at all about the iterations... I genuinely think the code got better with each round. The `init`-time filtering is cleaner than the per-batch flag approach, and having self.valid_steps_to_log as a single source of truth eliminates an entire class of ordering/zip edge cases. Completely agree with the direction.

I've pushed all the suggested changes

CHANGELOG.md
```diff
@@ -72,6 +72,8 @@ This release introduces new features including GIF animation support, wandb run
 
 - Fix Slack domain link [\#288](https://github.com/mllam/neural-lam/pull/288) @sadamov
 
+- Add missing bounds check in `test_step` to prevent `IndexError` when `val_steps_to_log` exceeds prediction steps [\#220](https://github.com/mllam/neural-lam/pull/220) @santhil-cyber
```
sadamov: This landed under `## [v0.6.0]` → `### Fixed`. Move it up to the `[unreleased]` section (line 10), which already has a `### Fixed` heading.

```suggestion
```

(Delete this line from here, and add it under `## [unreleased]` → `### Fixed` instead.)

neural_lam/models/ar_model.py
```diff
@@ -157,6 +157,22 @@ def __init__(
...
+        if invalid_steps:
+            warnings.warn(
+                f"val_steps_to_log contains steps {invalid_steps} that exceed "
+                f"ar_steps_eval ({args.ar_steps_eval}). These steps will be "
+                "skipped.",
+                UserWarning,
+            )
+        self.valid_steps_to_log = [
+            s for s in args.val_steps_to_log if s <= args.ar_steps_eval
+        ]
```
sadamov: This block crashes in CI with `AttributeError` on 13+ tests. The minimal `ModelArgs` fixtures in `test_datasets.py`, `test_training.py`, and `test_clamping.py` do not have `val_steps_to_log` or `ar_steps_eval` — those are either absent or passed to `WeatherDataModule` separately, not to the model.

```suggestion
        val_steps = getattr(args, "val_steps_to_log", [])
        ar_steps_eval = getattr(args, "ar_steps_eval", None)
        if ar_steps_eval is not None:
            invalid_steps = [s for s in val_steps if s > ar_steps_eval]
            if invalid_steps:
                warnings.warn(
                    f"val_steps_to_log contains steps {invalid_steps} that exceed "
                    f"ar_steps_eval ({ar_steps_eval}). These steps will be "
                    "skipped.",
                    UserWarning,
                )
            self.valid_steps_to_log = [
                s for s in val_steps if s <= ar_steps_eval
            ]
        else:
            self.valid_steps_to_log = val_steps
```

sadamov (changes requested):
Thanks for your patience during this busy period — almost ready to merge, just two things left.

santhil-cyber:
Hey @sadamov made the changes ...Can you check now?

sadamov:
@santhil-cyber please fix the tests first

santhil-cyber:
Please Verify @sadamov

sadamov (approved):
thanks @santhil-cyber you proposed a good fix for a small bug and over 2 months and 33 comments later we can merge. I am really sorry for this tedious road. Made some very small final adjustments and will merge right away. Thanks again!

santhil-cyber:
@sadamov You have been so supportive and polite all this time... its took a while but can understand that Gsoc application phase it was hard for you to manage all those chaos!! ...But Thank you learned and iterated a lot because of you ... Looking forward for more contri to the org!!

PR #224, opened 2026-02-22
refactor: use datastore directly in ARModel instead of WeatherDataset

### LINKED ISSUE #223: Refactor: Decouple WeatherDataset from ARModel._create_dataarray_from_tensor

Ayushhgit:
Refactor `ARModel._create_dataarray_from_tensor` to call `self._datastore.get_dataarray()` 
directly instead of instantiating a full `WeatherDataset` on every call. This removes 
unnecessary overhead (loading dataarrays, validating dimensions, computing standardization 
stats) when the method only needs coordinate metadata for constructing `xr.DataArray` objects.

Also removes the `WeatherDataset` import from `ar_model.py`, cleaning up the dependency graph.

Resolves the existing TODO comment at line ~191 of `ar_model.py`.

## Issue Link

closes #223 

## Type of change

  reflecting type of change:
  - *maintenance*: Refactor `ARModel._create_dataarray_from_tensor` to use datastore directly

### DIFF
--- a/neural_lam/models/ar_model.py
+++ b/neural_lam/models/ar_model.py
@@ -18,7 +18,7 @@ from .. import metrics, vis
 from ..config import NeuralLAMConfig
 from ..datastore import BaseDatastore
 from ..loss_weighting import get_state_feature_weighting
-from ..weather_dataset import WeatherDataset
+
 
 
 class ARModel(pl.LightningModule):
@@ -170,7 +170,7 @@ class ARModel(pl.LightningModule):
         """
         Create an `xr.DataArray` from a tensor, with the correct dimensions and
         coordinates to match the datastore used by the model. This function in
-        in effect is the inverse of what is returned by
+        effect is the inverse of what is returned by
         `WeatherDataset.__getitem__`.
 
         Parameters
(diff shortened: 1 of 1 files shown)


joeloskarsson:
Hi, could you make sure that the whole PR template is included?

joeloskarsson:
I think this needs some more thought and discussion. What I see now is that we get rid of the `WeatherDataset` instantiation, which is good, but then just end up with a bunch of duplicate code from the `create_dataarray_from_tensor` method that was used. We should aim for a solution here that does not create duplicate code for the datarrray creation. This might however require to change exactly where this method sits.

santhil-cyber:
@Ayushhgit good call removing the `WeatherDataset` instantiation — that's the right direction. Building on @joeloskarsson's point about duplicate code, two things I noticed:

1. The logic from `WeatherDataset.create_dataarray_from_tensor` is now copied into `ARModel`, so the same code lives in two places. If either changes, they'll drift apart.

2. The new version calls `self._datastore.get_dataarray(category, split)` on every invocation to get coordinate metadata. Depending on the backend, this could load the full dataarray from disk each time — the original `WeatherDataset` avoided this by caching `da_state` at init.

One idea: what if `create_dataarray_from_tensor` lived on `BaseDatastore` itself? It only needs `grid_index`, `{category}_feature`, and `x`/`y` coords — all things the datastore already owns. Then both `ARModel` and `WeatherDataset` just call `self._datastore.create_dataarray_from_tensor(...)`. Single source of truth, no extra instantiation.

Happy to help if useful 🙂

joeloskarsson:
There is some discussion in https://github.com/mllam/neural-lam/pull/209 about this instantiated WeatherDataset. Please also have a look there. It might make more sense to bring in this fix within the scope of that PR, as it is highly related.

neural_lam/models/ar_model.py
Sir-Sloth-The-Lazy, line 765: Just a typo, the changes are great, just to improve user experience !
 the spatial loss map title uses `self.time_step_int_unit`:
```
title=f"Test loss, t={t_i} "
f"({(self.time_step_int * t_i)} {self.time_step_int_unit})",
But the attribute is set in __init__ as:
pythonself.time_step_int, self.time_step_unit = get_integer_time(
    self._datastore.step_length
)
```
The correct attribute name is `self.time_step_unit`. As in, this will raise an AttributeError at test time when plotting spatial loss maps on rank 0. This line appears to have been carried over from the original and was not touched by this PR's changes, but it's a latent bug worth fixing here.
Fix:
```
f"({(self.time_step_int * t_i)} {self.time_step_unit})",
```
elegant work @Ayushhgit  ! 👌🏻 ** **chef's kiss** **

joeloskarsson:
In #209  the plan is to move the functionality for building the dataarray directly into the datastore. This seems like a better solution than what we have here atm, and also highly related to the changes of that PR. So I will close this for now and we can direct efforts around this to #209 .

PR #236, opened 2026-02-23
feat: enable pin_memory in DataLoaders for async GPU transfers

### LINKED ISSUE #233: Optimize GPU data transfer in  WeatherDataModule  by enabling pinned memory and tuning prefetch settings
I was profiling a training run on an A100 and noticed the GPU is intermittently idle between batches during the host-to-device transfer phase. After digging into  WeatherDataModule ([weather_dataset.py#L667-L698](https://github.com/mllam/neurallam/blob/main/neural_lam/weather_dataset.py#L667-L698)), I realized none of the three DataLoaders set pin_memory=True.

## Problem
When pin_memory is False (the default), PyTorch allocates batch tensors in pageable host memory. On every training step, Trainer.fit() has to do a synchronous copy from pageable RAM to GPU VRAM. The GPU sits idle until the copy finishes before it can start the forward pass.

With pin_memory=True, tensors are allocated in page-locked (pinned) memory instead, which lets CUDA do the transfer asynchronously via DMA — the GPU can overlap the next batch's transfer with the current batch's computation.

This matters for neural-lam specifically because batch tensors are fairly large. For a typical MEPS-like setup that is something like 
(4, ar_steps, 63784, 17)
 per sample — a non-trivial chunk of data moving across the PCIe bus every step.

## Current Code
All three dataloaders follow the same pattern:

``` python
def train_dataloader(self):
    """Load train dataset."""
    return torch.utils.data.DataLoader(
        self.train_dataset,
        batch_size=self.batch_size,
        num_workers=self.num_workers,
        shuffle=True,
        multiprocessing_context=self.multiprocessing_context,
        persistent_workers=True,
    )
No pin_memory, and prefetch_factor is left at the default of 2, which could also be worth tuning for our data sizes.
```

## Proposed Fix
1. Add pin_memory=True to all three DataLoaders, gated on CUDA availability (pinned memory on CPU-only machines wastes resources):

```  python
def train_dataloader(self):
    """Load train dataset."""
    return torch.utils.data.DataLoader(
        self.train_dataset,
        batch_size=self.batch_size,
        num_workers=self.num_workers,
        shuffle=True,
        multiprocessing_context=self.multiprocessing_context,
        persistent_workers=True,
        pin_memory=torch.cuda.is_available(),
    )
 ```
2. Optionally expose prefetch_factor as a DataModule parameter so users on memory-constrained machines can tune it, while giving a reasonable default (e.g. 3 for GPU training).

3. Apply the same change to  val_dataloader() and  test_dataloader().

## Why This Matters
This is a well-documented PyTorch best practice ([DataLoader docs — Memory Pinning](https://pytorch.org/docs/stable/data.html#memory-pinning)) that is easy to overlook in research codebases.
For large-grid weather data, the transfer-time fraction of total step time is higher than usual because tensors are spatially dense (every grid node, every feature, every timestep).
It is a one-line fix per dataloader with zero risk of changing model behavior — purely a throughput improvement.
Pairs well with the existing persistent_workers=True that is already set.
Happy to open a PR for this if it sounds reasonable. I can also include a quick benchmark comparing throughput before and after on a single-GPU setup.

### PR DESCRIPTION
**Pull request by abhaygoudannavar**
Enable pin_memory=torch.cuda.is_available() for all three DataLoaders in WeatherDataModule (train, val, test). When training on GPU, this allocates batch tensors in page-locked host memory, enabling async CPU-to-GPU transfer via DMA and reducing GPU idle time between batches.

On CPU-only machines, pin_memory remains False to avoid unnecessary memory overhead.

## Describe your changes

When training on GPU, this allocates batch tensors in page-locked (pinned) host memory, which allows CUDA to perform asynchronous CPU-to-GPU transfers via DMA. This reduces GPU idle time between batches, as the device can overlap the next batch's transfer with the current batch's forward pass.

On CPU-only machines, pin_memory remains False to avoid unnecessary memory overhead.

This is a single-line addition per DataLoader with no change to model behavior — purely a data transfer throughput improvement. It is a well-documented PyTorch best practice ([DataLoader docs — Memory Pinning](https://pytorch.org/docs/stable/data.html#memory-pinning)) and pairs well with the existing persistent_workers=True already set on all three DataLoaders.

No new dependencies are required.

## Issue Link

closes #233 

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [x] the tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix Slack domain link [\#288](https://github.com/mllam/neural-lam/pull/288) @sadamov
 
+### Fixed
+
+- Infer spatial coordinate names for MDPDatastore (rather than assuming names `x` and `y`), allows for e.g. lat/lon regular grids [\#169](https://github.com/mllam/neural-lam/pull/169) @leifdenby
+
 ### Maintenance
 
 - Update PR template to clarify milestone/roadmap requirement and maintenance changes [\#186](https://github.com/mllam/neural-lam/pull/186) @joeloskarsson
--- a/neural_lam/datastore/base.py
+++ b/neural_lam/datastore/base.py
@@ -447,7 +447,9 @@ class BaseRegularGridDatastore(BaseDatastore):
     `BaseDatastore`) for regular-gridded source data each `grid_index`
     coordinate value is assumed to be associated with `x` and `y`-values that
     allow the processed data-arrays can be reshaped back into into 2D
-    xy-gridded arrays.
+    xy-gridded arrays (to change the name of the spatial coordinates the
+    `spatial_coordinates` value should be changed from its default value of
+    `("x", "y")`).
 
     The following methods and attributes must be implemented for datastore that
     represents regular-gridded data:
@@ -464,7 +466,7 @@ class BaseRegularGridDatastore(BaseDatastore):
     `stack_grid_coords` and `unstack_grid_coords` respectively).
     """
 
-    CARTESIAN_COORDS = ["x", "y"]
+    spatial_coordinates = ("x", "y")
 
     @cached_property
     @abc.abstractmethod
@@ -507,8 +509,9 @@ class BaseRegularGridDatastore(BaseDatastore):
     ) -> Union[xr.DataArray, xr.Dataset]:
         """
         Unstack the spatial grid coordinates from `grid_index` into separate `x`
-        and `y` dimensions to create a 2D grid. Only performs unstacking if the
-        data is currently stacked (has grid_index dimension).
+        and `y` dimensions to create a 2D grid (if the spatial coordinates have
+        different names, those are used instead). Only performs unstacking if
+        the data is currently stacked (has grid_index dimension).
 
         Parameters
         ----------
@@ -526,16 +529,33 @@ class BaseRegularGridDatastore(BaseDatastore):
 
         # Check whether `grid_index` is a multi-index
         if not isinstance(da_or_ds.indexes.get("grid_index"), MultiIndex):
-            da_or_ds = da_or_ds.set_index(grid_index=self.CARTESIAN_COORDS)
+            da_or_ds = da_or_ds.set_index(grid_index=self.spatial_coordinates)
 
         da_or_ds_unstacked = da_or_ds.unstack("grid_index")
 
         # Ensure that the x, y dimensions are in the correct order
         dims = da_or_ds_unstacked.dims
-        xy_dim_order = [d for d in dims if d in self.CARTESIAN_COORDS]
+        xy_dim_order = [d for d in dims if d in self.spatial_coordinates]
 
-        if xy_dim_order != self.CARTESIAN_COORDS:
-            da_or_ds_unstacked = da_or_ds_unstacked.transpose("x", "y")
+        if xy_dim_order != self.spatial_coordinates:
+            # work out where the first spatial coordinate is located
+            # so that we can insert the second spatial coordinate next to it in
+            # the correct order. Although this looks verbose, it ensures that
+            # we don't change the order of any other dimensions.
+            first_xy_dim_index = min(
+                dims.index(self.spatial_coordinates[0]),
+                dims.index(self.spatial_coordinates[1]),
+            )
+            new_dim_order = list(dims)
+            new_dim_order.remove(self.spatial_coordinates[0])
+            new_dim_order.remove(self.spatial_coordinates[1])
+            new_dim_order.insert(
+                first_xy_dim_index, self.spatial_coordinates[0]
+            )
+            new_dim_order.insert(
+                first_xy_dim_index + 1, self.spatial_coordinates[1]
+            )
+            da_or_ds_unstacked = da_or_ds_unstacked.transpose(*new_dim_order)
 
         return da_or_ds_unstacked
 
@@ -561,7 +581,7 @@ class BaseRegularGridDatastore(BaseDatastore):
         if "grid_index" in da_or_ds.dims:
             return da_or_ds
 
-        da_or_ds_stacked = da_or_ds.stack(grid_index=self.CARTESIAN_COORDS)
+        da_or_ds_stacked = da_or_ds.stack(grid_index=self.spatial_coordinates)
 
         # infer what category of data the array represents by finding the
         # dimension named in the format `{category}_feature`
--- a/neural_lam/datastore/mdp.py
+++ b/neural_lam/datastore/mdp.py
@@ -107,7 +107,7 @@ class MDPDatastore(BaseRegularGridDatastore):
                     dim_order == dim_order_
                 ), "all inputs must have the same dimension order"
 
-        self.CARTESIAN_COORDS = dim_order
+        self.spatial_coordinates = dim_order
 
     @property
     def root_path(self) -> Path:
@@ -275,7 +275,7 @@ class MDPDatastore(BaseRegularGridDatastore):
                 da_category[coord].attrs["units"] = "m"
 
         # set multi-index for grid-index
-        da_category = da_category.set_index(grid_index=self.CARTESIAN_COORDS)
+        da_category = da_category.set_index(grid_index=self.spatial_coordinates)
 
         if "time" in da_category.dims:
             t_start = (
@@ -441,12 +441,15 @@ class MDPDatastore(BaseRegularGridDatastore):
 
         """
         ds_state = self.unstack_grid_coords(self._ds["state"])
-        da_x, da_y = ds_state.x, ds_state.y
+        xdim, ydim = self.spatial_coordinates
+        da_x, da_y = ds_state[xdim], ds_state[ydim]
         assert da_x.ndim == da_y.ndim == 1
         return CartesianGridShape(x=da_x.size, y=da_y.size)
 
     def get_xy(self, category: str, stacked: bool) -> ndarray:
-        """Return the x, y coordinates of the dataset.
+        """
+        Return the x, y coordinates of the dataset (i.e. the Cartesian
+        coordinates of the regular gridded dataset).
 
         Parameters
         ----------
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by abhaygoudannavar** [comment]
@observingClouds can you review this...

**Comment by observingClouds** [comment]
Thanks for this PR and finding this potential performance improvement in the first place.

At the link you provided it says:
> By default, if the pinning logic sees a batch that is a custom type (which will occur if you have a collate_fn that returns a custom batch type), or if each element of your batch is a custom type, the pinning logic will not recognize them, and it will return that batch (or those elements) without pinning the memory. 

So I am not sure if we currently benefit from setting this explicitly. I have seen `collate_fn` being used in some feature branches, so it is definitely something to keep in mind though. Have you done some test by any chance to see any improvement? @Ishwarpatra might also have some insights maybe?

**Comment by Ishwarpatra** [comment]
@observingClouds
 Thanks so much for tagging me! I'm still getting familiar with all the different feature branches, but this is a really interesting point about the PyTorch docs.
From what I understand, you are totally right that pin_memory=True gets confused by custom Python objects. But looking at the main branch, it seems like WeatherDataModule relies on PyTorch's default collate function, and WeatherDataset.__getitem__ just returns standard tensors (or standard collections of them), right? If that's the case, I think the default collater should hopefully be able to pin them automatically on main.
For the feature branches that do use a custom collate_fn returning custom batch objects, I was reading up on this and it looks like we might just need to add a custom pin_memory(self) method to those specific objects later on so PyTorch knows how to handle them.
To be absolutely sure it's actually helping and not breaking anything, maybe we could just do a quick test run and print batch.is_pinned() (or check the tensors inside) to see if the memory is actually getting pinned? A quick steps/sec benchmark would definitely confirm if it's worth it too! Let me know if I am misunderstanding how those feature branches are structured!

**Comment by abhaygoudannavar** [comment]
Thanks for the review @observingClouds and for the analysis @Ishwarpatra.

On main,WeatherDataset.__getitem__ returns a plain tuple of four standard torch.Tensor objects (init_states, target_states, forcing,target_times), and no custom collate_fn is passed to any of the DataLoaders. So the default collater stacks them into regular batched tensors, which the pinning logic handles without issues.

The custom collate_fn concern is valid for feature branches, but those branches would need to handle it on their end — it should not block this change on 
main.I have not run a formal benchmark yet, but I can put one together — print tensor.is_pinned() on the batch tensors to confirm pinning is working, and compare steps/sec with and without the flag. Let me know if that would be enough or if you would prefer something more detailed.

**Comment by abhaygoudannavar** [comment]
@observingClouds @Ishwarpatra 
Ran a benchmark to address the review feedback. Results below.

**Setup**
- PyTorch `2.6.0+cu124`
- NVIDIA GeForce RTX 4050 Laptop GPU (6 GB)
- Synthetic tensors sized like MEPS batches (`[4, 2, 10000, 17]`, `[4, 3, 10000, 17]`, etc.)
- 4 DataLoader workers, `persistent_workers=True`

### Test 1 — `is_pinned()` verification

All four batch tensors (`init_states`, `target_states`, [forcing](cci:1://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/neural_lam/weather_dataset.py:225:4-331:24), `target_times`) correctly return `is_pinned=True` with `pin_memory=True` and `is_pinned=False` without it. The default collater handles the standard tuple from [__getitem__](cci:1://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/neural_lam/weather_dataset.py:443:4-496:64) without issues.

### Test 2 — DataLoader iteration throughput

| Setting | Avg load time | Throughput |
|---|---|---|
| `pin_memory=False` | 0.02 ms/batch | 46,024 batches/sec |
| `pin_memory=True` | 0.01 ms/batch | 93,411 batches/sec |

The pinned path shows roughly **2x higher iteration throughput** on the DataLoader side. In a real training loop the end-to-end improvement will be smaller since forward/backward pass time dominates, but the async transfer overlap still reduces GPU idle time between batches.

Benchmark script and terminal output attached below.

Image

Image

**Comment by abhaygoudannavar** [comment]
@sadamov conflicts resolved 👍

**Comment by sadamov** [comment]
> @sadamov conflicts resolved 👍

hi, did you ping me on purpose or wanted to ping @observingClouds ?

**Comment by abhaygoudannavar** [comment]
@sadamov  Apologies for any confusion! I mentioned you because of your previous engagement on the thread, but I definitely want to ensure @observingClouds  sees this as well. Looking forward to any feedback you both might have.

**Comment by observingClouds** [comment]
Thanks for running the benchmark. This is very valuable. I might have missed it, but could you attach your benchmark script here again. I like to run it on my hardware as well.

I also re-read the [docs about pinning again](https://pytorch.org/docs/stable/data.html#memory-pinning) and understand now that memory pinning is be default off and can be turned on without modifications for simple tensors (like neural-lam currently uses). If a `collate_fn` is used, then `pin_memory()` needs to be implemented as well.

So, I can definitely see the value of this PR now.

**Review by observingClouds — CHANGES_REQUESTED** [review]
Could you please describe your changes in the CHANGELOG.md?

**Comment by abhaygoudannavar** [comment]
@observingClouds i have updated the CHANGELOG.md and here is the benchmark script:
```
import time

import torch
from torch.utils.data import DataLoader, Dataset


class SyntheticWeatherDataset(Dataset):

    def __init__(
        self,
        n_samples=100,
        num_grid_nodes=10000,
        d_features=17,
        d_forcing=5,
        ar_steps=3,
    ):
        self.n_samples = n_samples
        self.num_grid_nodes = num_grid_nodes
        self.d_features = d_features
        self.d_forcing = d_forcing
        self.ar_steps = ar_steps

    def __len__(self):
        return self.n_samples

    def __getitem__(self, idx):
        init_states = torch.randn(2, self.num_grid_nodes, self.d_features)
        target_states = torch.randn(
            self.ar_steps, self.num_grid_nodes, self.d_features
        )
        forcing = torch.randn(
            self.ar_steps, self.num_grid_nodes, self.d_forcing
        )
        target_times = torch.randn(self.ar_steps)
        return init_states, target_states, forcing, target_times


def check_pinned(batch, label):
    print(f"\n{'='*60}")
    print(f"  {label}")
    print(f"{'='*60}")
    names = ["init_states", "target_states", "forcing", "target_times"]
    all_pinned = True
    for name, tensor in zip(names, batch):
        pinned = tensor.is_pinned()
        if not pinned:
            all_pinned = False
        print(
            f"  {name:>16s}: shape={str(list(tensor.shape)):>25s}  is_pinned={pinned}"
        )
    return all_pinned


def benchmark_iteration(dataloader, label, num_batches=20):
    # warmup
    for i, batch in enumerate(dataloader):
        if i >= 3:
            break

    times = []
    for i, batch in enumerate(dataloader):
        if i >= num_batches:
            break
        start = time.perf_counter()
        _ = [t.shape for t in batch]
        elapsed = time.perf_counter() - start
        times.append(elapsed)

    avg_ms = (sum(times) / len(times)) * 1000
    steps_per_sec = len(times) / sum(times)
    print(f"\n{'='*60}")
    print(f"  {label}")
    print(f"{'='*60}")
    print(f"  Batches timed     : {len(times)}")
    print(f"  Avg load time     : {avg_ms:.2f} ms/batch")
    print(f"  Throughput        : {steps_per_sec:.2f} batches/sec")
    print(f"{'='*60}")
    return avg_ms, steps_per_sec


def main():
    print(f"PyTorch version : {torch.__version__}")
    print(f"CUDA available  : {torch.cuda.is_available()}")
    print(f"CUDA build      : {torch.version.cuda}")
    if torch.cuda.is_available():
        print(f"GPU             : {torch.cuda.get_device_name(0)}")

    dataset = SyntheticWeatherDataset(
        n_samples=100,
        num_grid_nodes=10000,
        d_features=17,
        ar_steps=3,
    )

    batch_size = 4
    num_workers = 4

    print("\n" + "#" * 60)
    print("  TEST 1: Verifying pin_memory behavior")
    print("#" * 60)

    dl_no_pin = DataLoader(
        dataset,
        batch_size=batch_size,
        num_workers=num_workers,
        pin_memory=False,
        persistent_workers=True,
    )
    dl_pin = DataLoader(
        dataset,
        batch_size=batch_size,
        num_workers=num_workers,
        pin_memory=True,
        persistent_workers=True,
    )

    batch_no_pin = next(iter(dl_no_pin))
    no_pin_result = check_pinned(batch_no_pin, "pin_memory=False")

    batch_pin = next(iter(dl_pin))
    pin_result = check_pinned(batch_pin, "pin_memory=True")

    print("\n" + "#" * 60)
    print("  TEST 2: DataLoader iteration throughput")
    print("#" * 60)

    avg_no_pin, sps_no_pin = benchmark_iteration(
        dl_no_pin, "pin_memory=False", num_batches=20
    )
    avg_pin, sps_pin = benchmark_iteration(
        dl_pin, "pin_memory=True", num_batches=20
    )

    print("\n" + "#" * 60)
    print("  SUMMARY")
    print("#" * 60)
    print(f"  PyTorch           : {torch.__version__}")
    print(f"  CUDA available    : {torch.cuda.is_available()}")
    if torch.cuda.is_available():
        print(f"  GPU               : {torch.cuda.get_device_name(0)}")
    print(f"")
    print(f"  pin_memory=False:")
    print(f"    is_pinned       : {no_pin_result}")
    print(f"    avg load time   : {avg_no_pin:.2f} ms/batch")
    print(f"    throughput      : {sps_no_pin:.2f} batches/sec")
    print(f"")
    print(f"  pin_memory=True:")
    print(f"    is_pinned       : {pin_result}")
    print(f"    avg load time   : {avg_pin:.2f} ms/batch")
    print(f"    throughput      : {sps_pin:.2f} batches/sec")
    print(f"")
    if pin_result:
        print(f"  RESULT: Tensors are correctly pinned when pin_memory=True")
    else:
        print(f"  RESULT: pin_memory has no effect (CPU-only PyTorch build)")
    print("#" * 60)


if __name__ == "__main__":
    main()
```

ArnabTechiee:
Hi @abhaygoudannavar, great catch on the pageable memory bottleneck! I pulled this branch locally to verify the changes. Testing the WeatherDataModule with pin_memory=True works perfectly and the dataloaders initialize without any breaking issues. The asynchronous DMA transfer should definitely smooth out the GPU utilization between batches. Code looks solid and ready for merge from my end!

observingClouds (changes requested):
I gave your benchmark a test on some A100 cluster and can confirm that it runs slightly faster with pinning.

If you could revert the changes made to the CHANGELOG.md entries that are not part of this PR, we can propose this work for the next release milestone 🚀

abhaygoudannavar:
Hi @observingClouds, thanks for the review and for testing the benchmark on A100! 

I've reverted all the unrelated CHANGELOG.md formatting changes. The diff now only contains:

- CHANGELOG.md : Added entry for the pin_memory feature (+4 lines)

- neural_lam/weather_dataset.py : pin_memory=torch.cuda.is_available() in train/val/test DataLoaders (+3 lines)

abhaygoudannavar:
@observingClouds will the PR be merged , i have done all the changes you asked for and wanted to know if there are any other changes that need to be done.

joeloskarsson:
This all makes sense to me as well, thanks for adding this! Should help in some settings where data loading overhead is substantial. I think all should be good, process now is just:

1. Waiting for an approve from @observingClouds 
2. We will put this on the roadmap for next version at next dev meeting.
3. Then I'll merge it in.

joeloskarsson:
Now merged. Thanks for this again!

PR #253, opened 2026-02-25
Fix division by zero in math operations

Ayushhgit:
## Describe your changes

**Summary of the changes:**
- Added a `1e-6` epsilon to the output of `torch.nn.functional.softplus` in [base_graph_model.py].
- Added a `1e-8` epsilon to the `self.feature_weights` inverse square root calculation inside `ARModel.__init__`.

**Motivation and context:**
These changes resolve two separate division-by-zero crashes:
1. When computing the `pred_std` using [softplus](neural-lam/neural_lam/utils.py:376:0-396:12), inputs $\le -80$ in `float32` will evaluate exactly to `0.0`. This zero denominator causes cascading `NaN` outputs instantly destroying training across metrics like [wmse](neural-lam/neural_lam/metrics.py:55:0-83:5) or [wmae](neural-lam/neural_lam/metrics.py:110:0-138:5).
2. When users specify a manual feature weight of `0.0` for specific state variables they want the model to ignore, creating `per_var_std` triggers a scalar division by zero, crashing model initialization. The epsilons bring numerical stability against these mathematical singularities.

**Dependencies:**
None

## Issue Link

N/A

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass


neural_lam/models/ar_model.py
```diff
@@ -104,7 +104,7 @@ def __init__(
             # in wMSE/wMAE
             self.register_buffer(
                 "per_var_std",
-                self.diff_std / torch.sqrt(self.feature_weights),
+                self.diff_std / torch.sqrt(self.feature_weights + 1e-8),
```
joeloskarsson: The situation where this would be needed is when a user has set a feature weight of 0, which is very odd and means that the specific field will not be part of the loss. 
If we should even allow this I think we at least need to then give a warning to the user about this.

joeloskarsson (changes requested):
Thanks, these are good changes to make the code more robust. Would it be possible to also add some test that checks that we don't run into numerical issues with the chosen constants?

Ayushhgit:
@joeloskarsson  Thanks for the feedback! I've just pushed a commit with the updates:

Warning: Added a UserWarning in ar_model.py that alerts the user and lists the indices if any variables receive a feature weight of exactly 0.0 (excluding them from the loss). Tests: Added 

test_numerical_stability.py with 5 pure tensor unit tests. These verify that the softplus epsilon (1e-6) and the feature weight epsilon (1e-8) successfully prevent exact zeros and subnormals, keeping downstream loss metrics (wmse, wmae, per_var_std) perfectly finite even with worst-case or zero-weight inputs. Let me know if you need anything else!

sadamov:
please link this issue in your PR description: https://github.com/mllam/neural-lam/issues/526

sadamov (approved):
Thanks @Ayushhgit, rebased onto current main and scoped down. Two changes:

1. Dropped the `softplus(pred_std_raw) + 1e-6` piece - that line overlaps with #523's `* self.diff_std` scaling, and once `pred_std` starts on the empirical scale via #523 the underflow concern is largely moot. Better to land the two ideas separately so they don't fight.
2. Switched to `torch.finfo(torch.float32).eps` for the sqrt eps (matches the existing `state_std` / `forcing_std` epsilon choice a few lines below).

tests requested by @joeloskarsson have been added before. please review

Refs #526. Merging after CI is green.

sadamov (approved):

joeloskarsson:
Thanks for adding the tests, but I am a bit confused here since they don't call any of the actual code. E.g. mirroring the `per_var_std` calculations separately in the test file seems very bad to maintain. Testing the numerical aspect is good, but we should do it running through the actual code used when training models. If this code is not structured such that it can be tested (requiring a mirroring function), then we should restructure it so that is possible.

sadamov:
@joeloskarsson good point, fixed in 37067b6 to keep things moving. Pulled the eps/warning logic out into `ForecasterModule._compute_per_var_std` (same pattern as the existing `_safe_std` static method), so the tests now call the real code instead of mirroring the formula.

Issue #212, opened 2026-02-21
mask_and_reduce_metric does not support area-weighted averaging over grid points


prajwal-tech07:
### **Problem**

All six metrics in metrics.py (mse, mae, wmse, wmae, nll, crps_gauss) reduce over the spatial grid dimension using a simple unweighted mean:

     if average_grid:  # Reduce grid first
          metric_entry_vals = torch.mean(metric_entry_vals, dim=-2)

This implicitly assumes that every grid point represents the same geographic area. That assumption is invalid for:

- **Latitude–longitude grids** — grid cells near the poles cover significantly less surface area than cells near the equator (e.g., a 1°×1° cell at 60°N has   
   roughly half the area of one at the equator).
- **LAM domains spanning wide latitude ranges** — even conformal projections do not perfectly equalize cell areas.
- **Global grids** — distortion becomes extreme, leading to strong polar over-representation.

As a result, reported RMSE/MAE/CRPS values are biased toward high-latitude regions, potentially misleading model evaluation and comparison.

### **Where the Problem Propagates**

Because all metrics call mask_and_reduce_metric, the unweighted averaging affects:

- Training loss via self.loss(...) in training_step
- Validation loss in validation_step
- Test metrics (MSE, MAE, etc.) in test_step

The issue originates from a single reduction point: mask_and_reduce_metric.
Therefore, it can be fixed centrally

### Proposed Solution

Add an optional weights parameter to mask_and_reduce_metric and propagate it through all metric functions.

**Updated Reduction Logic**

    def mask_and_reduce_metric(metric_entry_vals, mask, average_grid, weights=None):

          # Apply mask
          if mask is not None:
                 metric_entry_vals = metric_entry_vals[..., mask]
                 if weights is not None:
                        weights = weights[mask]

          # Reduce grid dimension
          if average_grid:
                 if weights is not None:
                        metric_entry_vals = (
                             (metric_entry_vals * weights.unsqueeze(...)).sum()
                               / weights.sum()
                             )
                else:
                        metric_entry_vals = torch.mean(metric_entry_vals, dim=-2)

### Metric Functions

Each metric (mse, mae, wmse, wmae, nll, crps_gauss) should:

- Add weights=None as an optional argument
- Pass weights through to mask_and_reduce_metric
- Keep the DEFINED_METRICS dictionary unchanged

###Model Integration

[...]

PR #256, opened 2026-02-25
Add optional grid_weights support to metric reduction functions

Joltsy10:
## Describe your changes

Adds an optional `grid_weights` parameter to `mask_and_reduce_metric` and all 
metric functions in `metrics.py`. When provided, grid nodes are averaged using 
a weighted mean rather than a simple mean. Default behavior is unchanged when 
`grid_weights=None`, preserving full backward compatibility with existing 
regional model usage.

This functionality exists in the `prob_model_global` branch. This PR ports a 
minimal, backward-compatible version into main with tests, without changing 
existing behavior or touching any training or wiring code. This is preparatory 
infrastructure for future global forecasting integration.

No new dependencies required.

## Issue Link

Relates to [#63](https://github.com/mllam/neural-lam/issues/63)

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented (including return types and parameters)
- [ ] the code is easy to maintain

## Author checklist after completed review

- [ ] I have added a line to the CHANGELOG describing this change, in a section
  reflecting type of change (add section where missing):
  - *added*: when you have added new functionality
  - *changed*: when default behaviour of the code has been changed
[...]

PR #168, opened 2025-11-20
166 plotting with correct projection config using pcolormesh

sadamov:
## Describe your changes

Switched to latitude/longitude based plotting using `pcolormesh` and `cartopy` for accurate spatial visualization. This method comes directly from the research branch.

- **Lat/Lon Plotting with `pcolormesh`**: Updated `neural_lam/vis.py` to use `ax.pcolormesh` with `ccrs.PlateCarree()` transform, using explicit latitude and longitude coordinates from the datastore. This ensures that data is plotted correctly on the map regardless of the underlying projection, fixing previous misalignment issues.
- **LatLon Retrieval**: 
    - **Base Implementation**: The default `get_lat_lon` in `BaseRegularGridDatastore` derives latitude/longitude from the Cartesian `x`/`y` coordinates using the defined projection (`coords_projection`).
    - **MDP Override**: The new implementation in `MDPDatastore` first checks if explicit latitude and longitude coordinates (e.g., `lat`/`lon` or `latitude`/`longitude`) exist in the dataset. If found, it uses these directly, bypassing the projection transformation.
- **Visualization Enhancements**:
    - Added `boundary_alpha` argument to overlay a semi-transparent mask on boundary regions.
    - Added `crop_to_interior` argument to automatically crop the plot view to the valid interior region.
    Note: The boundary vizualisation is deemed to change soon, with the separate boundary-datastore inclusion.

## Issue Link

solves #166

## Figures

Here are two examples for MEPS and DANRA test datastores with their own projections (`boundary_alpha = 0.7, crop_to_interior = True`).

### These are the spatial loss maps, generated during val and test

Screenshot From 2025-11-20 10-32-16
Screenshot From 2025-11-20 10-14-19

### These are the predictions generated during test

image
image

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
[...]

PR #258, opened 2026-02-25
[Feat] Implement area weights for grid-point averaging

prajwal-tech07:
Implemented area weights for grid-point averaging in the metrics module to ensure spatially weighted averaging instead of uniform averaging. Updated the related metric computations accordingly and added corresponding unit tests to validate correctness.

This improves the physical correctness of spatial averaging, especially for latitude-longitude grids where grid cells represent unequal surface areas.

No additional external dependencies were introduced.

## Issue Link
#212 – mask_and_reduce_metric does not support area-weighted averaging over grid points

## Type of change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📖 Documentation

- [x] I have added tests that prove my feature works
- [x] I have given the PR a clear name

### DIFF
--- a/neural_lam/config.py
+++ b/neural_lam/config.py
@@ -103,6 +103,8 @@ class TrainingConfig:
         default_factory=OutputClamping
     )
 
+    use_area_weights: bool = False
+
 
 @dataclasses.dataclass
 class NeuralLAMConfig(dataclass_wizard.JSONWizard, dataclass_wizard.YAMLWizard):
--- a/neural_lam/metrics.py
+++ b/neural_lam/metrics.py
@@ -97,10 +97,21 @@ def mask_and_reduce_metric(
     if average_grid:  # Reduce grid first
         if grid_weights is not None:
             # Weighted mean over grid dimension
-            # Reshape weights for broadcasting: (..., N', 1)
-            w = grid_weights.unsqueeze(-1)  # (N', 1)
+            # Reshape weights for broadcasting: (N', 1)
+            w = grid_weights.to(
+                device=metric_entry_vals.device,
+                dtype=metric_entry_vals.dtype,
+            ).unsqueeze(
+                -1
+            )  # (N', 1)
             # Normalize weights to sum to 1 over the (possibly masked) grid
-            w = w / w.sum(dim=-2, keepdim=True)
+            w_sum = w.sum(dim=-2, keepdim=True)
+            if w_sum.item() <= 0:
+                raise ValueError(
+                    "Sum of grid weights is non-positive after masking. "
+                    "All weighted grid points may have been masked out."
+                )
+            w = w / w_sum
             metric_entry_vals = torch.sum(
                 metric_entry_vals * w, dim=-2
             )  # (..., d_state)
--- a/neural_lam/metrics.py
+++ b/neural_lam/metrics.py
@@ -1,5 +1,7 @@
 # Third-party
+import numpy as np
 import torch
+from cartopy import crs as ccrs
 
 
 def get_metric(metric_name):
@@ -18,7 +20,51 @@ def get_metric(metric_name):
     return DEFINED_METRICS[metric_name_lower]
 
 
-def mask_and_reduce_metric(metric_entry_vals, mask, average_grid, sum_vars):
+def compute_area_weights(datastore):
+    """
+    Compute area weights for each grid point based on the cosine of the
+    latitude. This accounts for the fact that grid cells at higher latitudes
+    cover less surface area than cells near the equator.
+
+    The weights are normalized so that they sum to the number of grid points
+    (i.e. the mean weight is 1.0), so that using these weights is equivalent
+    to an unweighted mean when all latitudes are equal.
+
+    Parameters
+    ----------
+    datastore : BaseDatastore
+        The datastore providing grid coordinates and projection information.
+
+    Returns
+    -------
+    torch.Tensor
+        Area weights of shape (N,), where N is the number of grid points.
+        Each weight is proportional to cos(latitude) and the weights are
+        normalized to sum to N.
+    """
+    # Get projected x, y coordinates for state grid points
+    xy = datastore.get_xy(category="state", stacked=True)  # (N, 2)
+
+    # Transform projected coordinates to geographic lat/lon
+    lonlat = ccrs.PlateCarree().transform_points(
+        src_crs=datastore.coords_projection,
+        x=xy[:, 0],
+        y=xy[:, 1],
+    )  # (N, 3) -> [lon, lat, z]
+
+    lat_rad = np.deg2rad(lonlat[:, 1])  # latitude in radians
+    cos_lat = np.cos(lat_rad)
+
+    # Normalize weights so they sum to N (mean weight = 1.0)
+    n_points = len(cos_lat)
+    weights = cos_lat * (n_points / cos_lat.sum())
+
+    return torch.tensor(weights, dtype=torch.float32)
+
+
+def mask_and_reduce_metric(
+    metric_entry_vals, mask, average_grid, sum_vars, grid_weights=None
+):
     """
     Masks and (optionally) reduces entry-wise metric values
 
@@ -26,9 +72,13 @@ def mask_and_reduce_metric(metric_entry_vals, mask, average_grid, sum_vars):
         but broadcastable
     metric_entry_vals: (..., N, d_state), prediction
     mask: (N,), boolean mask describing which grid nodes to use in metric
-    average_grid: boolean, if grid dimension -2 should be reduced (mean over N)
+    average_grid: boolean, if grid dimension -2 should be reduced
+        (mean over N). If grid_weights are provided, uses weighted mean.
     sum_vars: boolean, if variable dimension -1 should be reduced (sum
         over d_state)
+    grid_weights: (N,) or None, optional area-based weights for each grid
+        point. When provided and average_grid=True, a weighted mean is used
+        instead of a simple mean. The weights are applied after masking.
 
     Returns:
     metric_val: One of (...,), (..., d_state), (..., N), (..., N, d_state),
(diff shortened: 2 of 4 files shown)


prajwal-tech07 (commented):
reviewed

sadamov:
This supercedes #256. @prajwal-tech07 please use the original PR template.

Joltsy10:
Reviewed the test suite in #258. While the coverage is thorough, there's some overlap worth noting:

test_uniform_weights_equals_no_weights and test_no_weights_gives_same_result_as_before are essentially testing the same invariant. Since both rely on mask_and_reduce_metric, we could test the weighting logic once in TestMaskAndReduceMetric and use a single parametrized check at the metric level to verify the API hook exists.

TestComputeAreaWeights could also be condensed. Testing shape, dtype, positivity, and normalization in four separate tests adds boilerplate; I think one test with multiple assertions covers it just as well.

The tests that cover genuinely distinct behavior are test_metrics_with_batch_and_time_dims, test_weighted_mean_with_mask, and the manual calculation tests. A suite focused on those would be leaner and easier to maintain.

neural_lam/metrics.py
```diff
@@ -18,17 +20,65 @@ def get_metric(metric_name):
     return DEFINED_METRICS[metric_name_lower]
 
 
-def mask_and_reduce_metric(metric_entry_vals, mask, average_grid, sum_vars):
+def compute_area_weights(datastore):
+    """
+    Compute area weights for each grid point based on the cosine of the
+    latitude. This accounts for the fact that grid cells at higher latitudes
+    cover less surface area than cells near the equator.
```
joeloskarsson, line 27: This is true for regular lat-lon grids, but not completely in general. Neural-lam supports all kinds of irregular grids where this might not be the case. While this is great functionality to have, and will be useful and necessary in most global setups, I would really like to make the use of such weights optional, to not cause issues in setups with other grids.

joeloskarsson: Perhaps the use of this (on/off) could be configures in the same config files that we now use to set the weight for each variable in the loss?

Joltsy10: Joel's point makes sense, keeping metrics.py generic with grid_weights as optional is the right direction. One question on where the weight computation itself should live. Should it stay in metrics, computed on the fly when needed, or should it be computed once at ARModel init and registered as a buffer alongside boundary_mask, with the config just controlling whether it's enabled? The buffer approach feels cleaner for consistency with how boundary_mask is handled, but wanted to check if you had a preference before Prajwal iterates on this.

santhil-cyber: Good question @Joltsy10 — I'd strongly lean toward the buffer approach (compute once at `ARModel.__init__`, register alongside `boundary_mask`).

A few reasons:

1. **Consistency** — `boundary_mask` is already a registered buffer. Treating `grid_weights` the same way keeps the pattern predictable.
2. **Performance** — `compute_area_weights()` involves a `cos()` call over all grid nodes. Small cost, but no reason to pay it every forward pass when weights are static.
3. **Config-driven** (addressing Joel's point) — if weights are a buffer, the config just controls whether to populate it or leave it `None`. The metrics code stays generic: `grid_weights=None` → unweighted, `grid_weights=tensor` → weighted.

On the edge cases — in my PR #316 I added two guards that might be worth incorporating here:

- **Device/dtype cast:** `grid_weights.to(metric_entry_vals.device, dtype=metric_entry_vals.dtype)` before any operations — prevents GPU/CPU mismatch during eval
- **Zero-sum guard:** `if w_sum <= 0: raise ValueError(...)` after masking — prevents silent `inf` from divide-by-zero if all weighted nodes get masked out

Happy to help align the implementations if that's useful.

prajwal-tech07:
Thanks for all the feedback! I've addressed everything in the latest push:

@joeloskarsson — Area weights are now fully optional via a `use_area_weights: bool` flag in `TrainingConfig` (defaults to False), so irregular grids aren't affected. The setting lives in the same config alongside `state_feature_weighting` and `output_clamping`.

@Joltsy10 — Went with the buffer approach: `compute_area_weights()` is called once at `ARModel.__init__`( when enabled, registered alongside `boundary_mask`. When disabled, `self.grid_weights = None` and metrics stay unweighted. Also consolidated the redundant tests — merged the two backward-compat tests into one and condensed `TestComputeAreaWeights` into a single test with multiple assertions.

@santhil-cyber — Added `grid_weights.to(device=..., dtype=...)` before operations to prevent GPU/CPU mismatch, and a `if w_sum <= 0: raise ValueError(...)` guard after masking to prevent silent inf.

neural_lam/metrics.py
joeloskarsson: Yea I think the buffer approach does make sense

neural_lam/metrics.py
```diff
@@ -18,17 +20,65 @@ def get_metric(metric_name):
...
+    """
+    # Get projected x, y coordinates for state grid points
+    xy = datastore.get_xy(category="state", stacked=True)  # (N, 2)
+
+    # Transform projected coordinates to geographic lat/lon
+    lonlat = ccrs.PlateCarree().transform_points(
+        src_crs=datastore.coords_projection,
+        x=xy[:, 0],
+        y=xy[:, 1],
+    )  # (N, 3) -> [lon, lat, z]
```
joeloskarsson, line 53: Note that #168 will implement a `get_lat_lon` method for the datastore. So when that is merged that can also be used here.

neural_lam/metrics.py
```diff
@@ -39,12 +89,36 @@ def mask_and_reduce_metric(metric_entry_vals, mask, average_grid, sum_vars):
...
     if average_grid:  # Reduce grid first
-        metric_entry_vals = torch.mean(
-            metric_entry_vals, dim=-2
-        )  # (..., d_state)
+        if grid_weights is not None:
+            # Weighted mean over grid dimension
+            # Reshape weights for broadcasting: (N', 1)
+            w = grid_weights.to(
+                device=metric_entry_vals.device,
+                dtype=metric_entry_vals.dtype,
```
joeloskarsson, line 103: Ensuring correct device and dtype should be responsibility of whoever calls this function.

neural_lam/metrics.py
```diff
@@ -39,12 +89,36 @@ def mask_and_reduce_metric(metric_entry_vals, mask, average_grid, sum_vars):
...
+            w_sum = w.sum(dim=-2, keepdim=True)
+            if w_sum.item() <= 0:
+                raise ValueError(
+                    "Sum of grid weights is non-positive after masking. "
+                    "All weighted grid points may have been masked out."
+                )
+            w = w / w_sum
+            metric_entry_vals = torch.sum(
+                metric_entry_vals * w, dim=-2
+            )  # (..., d_state)
```
joeloskarsson, line 117: I think a better approach here is to keep this as a mean, and not make the weights sum to 1. Instead when created we should rescale the weights with the average, so their mean is 1. See for example weatherbench 2 for this.

joeloskarsson: If this is not a mean (first weighted by given grid weights), then the `average_grid` argument doesn't make sense.

tests/test_metrics.py
```diff
@@ -0,0 +1,395 @@
...
+    # They should generally differ (we use a loose check — if they are
+    # identical bit-for-bit that would be extremely unlikely with random data)
+    # This is a soft assertion; we mainly verify correctness elsewhere.
+    assert result_weighted.shape == result_no_weights.shape
+
+
+# ============================================================
+# Test: mask_and_reduce_metric directly
+# ============================================================
+class TestMaskAndReduceMetric:
```
joeloskarsson, line 111: Why structure this in a class, seems inconsistent with how we typically structure tests.

neural_lam/config.py
```diff
@@ -103,6 +103,8 @@ class TrainingConfig:
         default_factory=OutputClamping
     )
 
+    use_area_weights: bool = False
```
joeloskarsson, line 106: I suggest a different name, as we here assume that we can compute this area from latitude.
```suggestion
    use_lat_area_weights: bool = False
```

joeloskarsson (changes requested):
Starting to look pretty good! Great discussions here. I added some suggestions 😄

neural_lam/metrics.py
```diff
@@ -39,12 +89,36 @@ def mask_and_reduce_metric(metric_entry_vals, mask, average_grid, sum_vars):
...
+        # Also mask the grid weights if provided
+        if grid_weights is not None:
+            grid_weights = grid_weights[mask]  # (N',)
 
     # Optionally reduce last two dimensions
     if average_grid:  # Reduce grid first
-        metric_entry_vals = torch.mean(
-            metric_entry_vals, dim=-2
-        )  # (..., d_state)
+        if grid_weights is not None:
```
Omnagar, line 98: Small design question regarding the weighting logic here.

Since `compute_area_weights()` already normalizes the weights so that
their mean equals 1, would it still be necessary to renormalize them
again inside `mask_and_reduce_metric()` after masking?

For example, if masking removes some grid points, the normalization here
changes the relative scaling of the remaining weights. I’m wondering if
it might be clearer to keep the weights fixed (mean≈1) and simply apply
a weighted mean without renormalizing after masking.

Curious what the preferred behaviour should be in this case.

Omnagar (commented):
Thanks for the detailed implementation of area-weighted metrics.

I left a small comment in `metrics.py` regarding the normalization of grid weights after masking. I was curious whether renormalizing inside `mask_and_reduce_metric()` is necessary given that `compute_area_weights()` already normalizes the weights.

Overall the approach looks clean and the integration with the existing metric pipeline makes sense.

neural_lam/metrics.py
Joltsy10: The renormalization after masking is intentional. compute_area_weights() normalizes over all grid nodes, but after the interior mask is applied the remaining weights no longer sum to 1 since boundary nodes have been removed. Without renormalizing, the weighted mean would not integrate correctly over the interior and the loss scale would vary depending on boundary region size. The second normalization ensures the result is a proper weighted average over interior nodes regardless of how large the boundary region is.

joeloskarsson: related to https://github.com/mllam/neural-lam/pull/258#discussion_r2937039832

Omnagar: > The renormalization after masking is intentional. compute_area_weights() normalizes over all grid nodes, but after the interior mask is applied the remaining weights no longer sum to 1 since boundary nodes have been removed. Without renormalizing, the weighted mean would not integrate correctly over the interior and the loss scale would vary depending on boundary region size. The second normalization ensures the result is a proper weighted average over interior nodes regardless of how large the boundary region is.

That makes sense, thanks for the clarification — especially the point about
the loss scale changing depending on the masked region size.

One follow-up thought: would it be worth documenting this behaviour explicitly
in the docstring of `mask_and_reduce_metric()`? It feels like a subtle detail,
and future contributors might assume the weights are only normalized once at
creation time.

I can help add a short clarification there if that sounds useful.

sadamov:
Closing this in favour of #650, which consolidates the global forecasting / spherical domain track for v0.8.0. The cos-lat area-weighted grid-point averaging here will be folded into the consolidated PR. Thanks for the contribution @prajwal-tech07, you'll be credited.

PR #267, opened 2026-02-25
fix: complete deterministic training with CUBLAS workspace config and…

abhaygoudannavar:
- Set CUBLAS_WORKSPACE_CONFIG=:4096:8 at module level in train_model.py to ensure cuBLAS uses deterministic algorithms on GPU, complementing the existing deterministic=True Trainer flag.

- Add _worker_init_fn to all three DataLoaders in WeatherDataModule that seeds each worker independently using torch.initial_seed(), preventing correlated random states across workers.

## Describe your changes

This PR completes the deterministic training setup in neural-lam. The Trainer already sets `deterministic=True`, but two pieces were missing for runs to be truly bitwise-reproducible with the same seed:
1. **`CUBLAS_WORKSPACE_CONFIG`** — Set to `:4096:8` at module level in [train_model.py](cci:7://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/neural_lam/train_model.py:0:0-0:0) using `os.environ.setdefault()`, so it does not override a user-provided value. Without this, cuBLAS matrix multiplications on GPU can still produce non-deterministic results even with `deterministic=True`.
2. **[worker_init_fn](cci:1://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/neural_lam/weather_dataset.py:16:0-25:28)** — Added a [_worker_init_fn](cci:1://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/neural_lam/weather_dataset.py:16:0-25:28) in [weather_dataset.py](cci:7://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/neural_lam/weather_dataset.py:0:0-0:0) that seeds each DataLoader worker independently using `torch.initial_seed() % 2**32`. Passed to all three DataLoaders (train, val, test). Without this, workers spawned with `num_workers > 0` can end up with correlated random states across runs.
Both changes are additive and do not alter any model logic or existing results.
No new dependencies required.

## Tests

Added [tests/test_deterministic.py](cci:7://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/tests/test_deterministic.py:0:0-0:0) with 4 tests (run with `--noconftest` to skip dataset download):
- [test_cublas_workspace_config_is_set](cci:1://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/tests/test_deterministic.py:12:0-22:65) — Confirms env var is set on import
- [test_cublas_does_not_override_user_value](cci:1://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/tests/test_deterministic.py:25:0-39:51) — Confirms user-provided values are preserved
- [test_worker_init_fn_seeds_deterministically](cci:1://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/tests/test_deterministic.py:42:0-56:64) — Same seed → same random values
- [test_worker_init_fn_different_seeds_differ](cci:1://file://wsl.localhost/Ubuntu/home/abhay/neural-lam/tests/test_deterministic.py:59:0-72:71) — Different seed → different random values
All 4 passed ✅

image

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/neural_lam/weather_dataset.py
+++ b/neural_lam/weather_dataset.py
@@ -1,5 +1,6 @@
 # Standard library
 import datetime
+import random
 import warnings
 from typing import Union
 
@@ -11,6 +12,18 @@ import xarray as xr
 
 # First-party
 from neural_lam.datastore.base import BaseDatastore
+
+
+def _worker_init_fn(worker_id):
+    """Seed each DataLoader worker independently for reproducibility.
+
+    Without this, workers spawned with num_workers > 0 can end up with
+    correlated random states across runs. See:
+    https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading
+    """
+    worker_seed = torch.initial_seed() % 2**32
+    np.random.seed(worker_seed)
+    random.seed(worker_seed)
 
 
 class WeatherDataset(torch.utils.data.Dataset):
@@ -673,6 +686,7 @@ class WeatherDataModule(pl.LightningDataModule):
             shuffle=True,
             multiprocessing_context=self.multiprocessing_context,
             persistent_workers=True,
+            worker_init_fn=_worker_init_fn,
         )
 
     def val_dataloader(self):
@@ -684,6 +698,7 @@ class WeatherDataModule(pl.LightningDataModule):
             shuffle=False,
             multiprocessing_context=self.multiprocessing_context,
             persistent_workers=True,
+            worker_init_fn=_worker_init_fn,
         )
 
     def test_dataloader(self):
@@ -695,4 +710,5 @@ class WeatherDataModule(pl.LightningDataModule):
             shuffle=False,
             multiprocessing_context=self.multiprocessing_context,
             persistent_workers=True,
+            worker_init_fn=_worker_init_fn,
         )
--- a/neural_lam/weather_dataset.py
+++ b/neural_lam/weather_dataset.py
@@ -22,7 +22,7 @@ def _worker_init_fn(worker_id):
     correlated random states across runs. See:
     https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading
     """
-    worker_seed = torch.initial_seed()
+    worker_seed = torch.initial_seed() + worker_id
     np.random.seed(worker_seed)
     random.seed(worker_seed)
 
--- a/neural_lam/weather_dataset.py
+++ b/neural_lam/weather_dataset.py
@@ -21,7 +21,7 @@ def _worker_init_fn(worker_id):
     correlated random states across runs. See:
     https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading
     """
-    worker_seed = torch.initial_seed() % 2**32
+    worker_seed = torch.initial_seed()
     np.random.seed(worker_seed)
     random.seed(worker_seed)
 
--- a/neural_lam/weather_dataset.py
+++ b/neural_lam/weather_dataset.py
@@ -22,7 +22,7 @@ def _worker_init_fn(worker_id):
     correlated random states across runs. See:
     https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading
     """
-    worker_seed = torch.initial_seed() + worker_id
+    worker_seed = (torch.initial_seed() + worker_id) % 2**32
     np.random.seed(worker_seed)
     random.seed(worker_seed)
 
--- a/neural_lam/weather_dataset.py
+++ b/neural_lam/weather_dataset.py
@@ -22,7 +22,7 @@ def _worker_init_fn(worker_id):
     correlated random states across runs. See:
     https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading
     """
-    worker_seed = (torch.initial_seed() + worker_id) % 2**32
+    worker_seed = torch.initial_seed() % 2**32
     np.random.seed(worker_seed)
     random.seed(worker_seed)
 
--- a/tests/test_deterministic.py
+++ b/tests/test_deterministic.py
@@ -13,7 +13,7 @@ def _worker_init_fn(worker_id):
     torch_geometric and other heavy dependencies).  Any change to the
     real function must be reflected here as well.
     """
-    worker_seed = torch.initial_seed() + worker_id
+    worker_seed = (torch.initial_seed() + worker_id) % 2**32
     np.random.seed(worker_seed)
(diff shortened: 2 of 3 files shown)


abhaygoudannavar:
@sadamov can you take a look on this...

sadamov:
@abhaygoudannavar I gotta be honest here, I have very little knowledge of CUBLAS inner workings. Maybe @leifdenby

abhaygoudannavar:
@leifdenby i wanted to know your opinion on this PR.

tests/test_deterministic.py
```diff
@@ -0,0 +1,73 @@
...
+    """Calling _worker_init_fn with different torch seeds should produce
+    different random values."""
+
+    def get_random_values_after_init(seed_val, worker_id):
+        torch.manual_seed(seed_val)
+        _worker_init_fn(worker_id)
+        return np.random.rand(), random.random()
+
+    a1, b1 = get_random_values_after_init(42, worker_id=0)
+    a2, b2 = get_random_values_after_init(99, worker_id=0)
```
joeloskarsson, line 49: Isn't the thing we actually want to test the case where torch initial seed is the same, but worker id different?

joeloskarsson:
I think there is a high-level question here: Do we want neural-lam to be responsible for making users training/eval completely deterministic? This then requires making some opinionated decisions, and as done here start setting environment variables. Even with this, I believe that we are at a point with DL frameworks where we can still not guarantee complete determinism on some different hardware setups. 
My opinion on this is that we should try to make things deterministic in the python code as well as we can, but some of this has to be left to the user. Starting to set the CUBLAS env var is probably something I would view as outside of neural-lam's scope, and I would prefer to leave this up to the user. @sadamov @leifdenby do you have some opinions on this?

Then there is the second part to this PR, about worker seeds. This is not really a reproducability issue (as even with the same seed for different workers, things would still be reproducible). But this is in general a good addition I think. So I would recommend to focus this PR on that, and then the env var change can be up for discussion (this would anyhow be better as two separate PRs, as they are different changes). 

Separate question, do you know if/how the seed set in `worker_init_fn` affects which data indices are sampled for each batch?

neural_lam/weather_dataset.py
```diff
@@ -13,6 +14,18 @@
...
 
 
+def _worker_init_fn(worker_id):
+    """Seed each DataLoader worker independently for reproducibility.
+
+    Without this, workers spawned with num_workers > 0 can end up with
+    correlated random states across runs. See:
+    https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading
+    """
+    worker_seed = torch.initial_seed() % 2**32
```
joeloskarsson, line 25: Why the `% 2**32` here?

sadamov:
> My opinion on this is that we should try to make things deterministic in the python code as well as we can, but some of this has to be left to the user. Starting to set the CUBLAS env var is probably something I would view as outside of neural-lam's scope, and I would prefer to leave this up to the user. @sadamov @leifdenby do you have some opinions on this?

Yes, I agree. What could be nice is having interactive documentation aka notebooks, that guide the user on how to run neural-lam on different common types of hardware and schedulers. In such a notebook one could also impose maximum determinism. But maybe these notebooks would rather live here instead of in neural-lam: https://github.com/mllam/mllam-docs (although that one is a bit dead)

tests/test_deterministic.py
abhaygoudannavar: I've updated the test to use the same seed (42) with different worker IDs (0 and 1). I also updated 
_worker_init_fn to incorporate worker_id into the seed calculation (torch.initial_seed() + worker_id), so different workers are guaranteed independent random states even when the base torch seed is the same. Both tests pass.

abhaygoudannavar:
@joeloskarsson As far as I know, the seed in worker_init_fn doesn't affect which indices get sampled — that's handled by the Sampler in the main process, which is already seeded by seed_everything(). The worker_init_fn is really just there to make sure each worker's numpy/ random state is independent, so you don't get correlated randomness during data loading (like if there are augmentations happening in getitem).

neural_lam/weather_dataset.py
```diff
@@ -14,6 +15,18 @@
...
 
 
+def _worker_init_fn(worker_id):
+    """Seed each DataLoader worker independently for reproducibility.
+
+    Without this, workers spawned with num_workers > 0 can end up with
+    correlated random states across runs. See:
+    https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading
+    """
+    worker_seed = torch.initial_seed() + worker_id
```
kshirajahere: I think this change introduces a real bug.

torch.initial_seed() in a DataLoader worker is already the worker-specific PyTorch seed, so adding worker_id here should not be necessary. More importantly, removing the % 2\**32 breaks NumPy seeding for large valid Torch seeds, because numpy.random.seed() only accepts values in [0, 2\**32 - 1].

I reproduced this locally with torch.manual_seed(2\**40) and a one-worker DataLoader: worker startup fails with ValueError: Seed must be between 0 and 2\**32 - 1.

PyTorch's reproducibility docs still recommend:

```python
worker_seed = torch.initial_seed() % 2**32
np.random.seed(worker_seed)
random.seed(worker_seed)
```
So I think this should go back to the documented pattern rather than torch.initial_seed() + worker_id.

kshirajahere (commented):
Hello my concern is updated seeding path appears to drop the % 2\**32 bound before passing the seed to NumPy / random, which can break worker startup for large but valid Torch seeds. I was able to reproduce that locally with a one-worker DataLoader.

neural_lam/weather_dataset.py
abhaygoudannavar: You're absolutely right — torch.initial_seed() inside a DataLoader worker is already set to base_seed + worker_id by PyTorch, so adding worker_id again is redundant. I've reverted to the documented pattern torch.initial_seed() % 2**32. Good catch on the NumPy seed range issue too.

abhaygoudannavar:
@kshirajahere I've updated the code to follow the PyTorch recommended pattern. The % 2**32 is back to handle the NumPy seed range, and I've dropped the extra + worker_id since torch already handles that per-worker.

kshirajahere (commented):
Looks fine to me, It will be nice if you can add your changes in changelog.md. Rest maintainers will give their insights. 😄

abhaygoudannavar:
@joeloskarsson got approval from @kshirajahere are there any changes still needed on this pr?

PR #273, opened 2026-02-26
Add deterministic regression test for GraphLAM outputs with CI

### LINKED ISSUE #59: Regression tests for model outputs
Something that would be really nice to have is regression testing for model outputs. In short, whenever we refactor something in models we want them to still be able to load checkpoints (or well, see #48 ) and give exactly the same output when being fed with the same data.

One way to achieve this could be to

1. Check out main branch
2. Run some example data through the model and save the predictions (potentially also some internal representation tensors but likely unneccesary and hard to do in practice)
3. Check out PR
4. Run the same example data through the model and compare outputs to saved predictions.

I'm not too familiar with `pytest` and the github workflows to know all the details of how to do this. @SimonKamuk, @leifdenby  do you think something like this is doable? Or are there any better ways to achieve this?

Saptami191:
## Describe your changes

This PR adds deterministic regression testing for `GraphLAM` model outputs.

The following components are introduced:

- A regression test (`tests/test_model_regression.py`) that verifies
  `GraphLAM.process_step` produces identical outputs given a fixed seed
  and frozen reference tensors.
- Reference input/output tensors stored under `tests/regression/`.
- A script (`scripts/update_regression_baseline.py`) to regenerate
  the regression baseline when behavior changes are intentional.
- A GitHub Actions workflow (`.github/workflows/regression.yml`)
  that runs the regression test on pull requests.

### Motivation and context

When refactoring model internals, small implementation changes can
silently alter numerical outputs.

This regression test ensures:

- Model outputs remain stable across refactors
- Behavioral changes are intentional
- CI fails if numerical outputs drift unexpectedly

The regression test:

- Runs on CPU only
- Uses deterministic PyTorch algorithms
- Uses strict floating-point tolerances (`rtol=1e-5`, `atol=1e-6`)
- Avoids GPU-based nondeterminism

### Dependencies

No new dependencies are introduced.

---

## Issue Link

Closes #59

---

## Type of change

---

- [ ] I have requested a reviewer and an assignee (maintainer tagging required)

---

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

---

  reflecting type of change:
  - *added*: Deterministic regression test for GraphLAM model outputs

---

## Checklist for assignee

- [ ] the tests pass


Saptami191:
@joeloskarsson @leifdenby   Would appreciate a review when you have time.  the last commit  [48cc5f5](https://github.com/mllam/neural-lam/pull/273/commits/48cc5f564bb7a6e164044061b57b1d6cf02565a3). Hope this design make it better.  happy to adjust the baseline storage approach or test scope if needed.

Saptami191:
This PR updates the regression test tolerance in test_graph_lam_regression to account for minor cross-platform floating point drift observed between local (Windows/Python 3.11) and CI (Linux/Python 3.13) environments.previous tolerances (rtol=1e-5, atol=1e-6) were too strict and caused failures due to small numerical differences (~3e-6 absolute max diff) despite identical model behavior. The updated tolerances (rtol=1e-4, atol=1e-5) remain strict enough to catch meaningful regressions while allowing expected floating-point variance across platforms.all CPU tests and linting checks are passing. GPU jobs were cancelled due to runner unavailability, not test failures.

Saptami191:
@sadamov Kindly review this pr also . just pinged if the pr gets slipped.

neural_lam/vis.py
```diff
@@ -16,7 +16,13 @@ def plot_error_map(errors, datastore: BaseRegularGridDatastore, title=None):
...
     errors: (pred_steps, d_f)
     """
-    errors_np = errors.T.cpu().numpy()  # (d_f, pred_steps)
+    errors_np = errors.detach().cpu().numpy()
+
+    if errors_np.ndim == 1:
+        # single prediction horizon case
+        errors_np = errors_np[None, :]  # make it (1, d_f)
+
+    errors_np = errors_np.T  # (d_f, pred_steps)
```
joeloskarsson, line 25: Seems unrelated?

scripts/update_regression_baseline.py
```diff
@@ -0,0 +1,93 @@
...
+    config, datastore = load_config_and_datastore(args.config_path)
+
+    ModelClass = MODELS[args.model]
+    model = ModelClass(args, config=config, datastore=datastore)
+    model.eval()
+
+    mesh_nodes = model.mesh_static_features.shape[0]
+    hidden_dim = args.hidden_dim
+
+    batch_input = torch.randn(1, mesh_nodes, hidden_dim)
```
joeloskarsson: See my point 4 here https://github.com/mllam/neural-lam/issues/59#issuecomment-3954507330

tests/test_model_regression.py
```diff
@@ -0,0 +1,96 @@
...
+    parser.add_argument("--num_future_forcing_steps", type=int, default=1)
+
+    args = parser.parse_args(
+        [
+            "--config_path",
+            "tests/datastore_examples/mdp/danra_100m_winds/config.yaml",
+            "--graph",
+            "1level",
+        ]
+    )
```
joeloskarsson: Instead of mocking the whole args setup again, I think it would make more sense to unify this with the approach in https://github.com/mllam/neural-lam/blob/3d5d56b8d0562c7074a0a1ec33036ae9eaafa319/tests/test_training.py#L76-L92. That seems simpler, but in particular we should use the same approach  in all tests.

joeloskarsson:
In my view it would be better to run the same forward pass before and after a change, comparing outputs, rather than storing tensors in the repo. I made this point here https://github.com/mllam/neural-lam/issues/59#issuecomment-3954507330 (points 2 and 3).

Curious to hear what @leifdenby and @observingClouds think though, who have worked more on the cicd setup.

tests/test_model_regression.py
Saptami191: actually sir my initial implementation aimed to create a deterministic regression check for the model forward pass.  idea was to ensure that when the same checkpoint and input are used, the model produces numerically consistent outputs so that internal refactors dontt unintentionally change behavior. This follows the idea of Run baseline model + PR model and compare outputs, where outputs from a known-good version are compared against future runs to detect unintended changes.To keep the test lightweight and reproducible in CI, I used fixed inputs and tolerance-based floating-point comparisons so that small numerical differences do not cause false failures.
 for ci/cd work flow i tested it like this and some changes were unrelated . 
wud it be better to generate the baseline outputs from a trained checkpoint using the same example dataset already used in other tests, rather than synthetic inputs?

If that aligns with the intended approach, I can adjust the test to:

run CPU-only inference
use a small example dataset sample
compare outputs with torch.allclose tolerances

Saptami191:
@joeloskarsson Thanks for the feedback on the testing approach. i updated the PR to remove the stored baseline tensors and the baseline generation script. the regression test now runs the same forward pass twice and checks that the outputs match using torch.testing.assert_close, avoiding storing artifacts in the repository.

PR #281, opened 2026-02-27
feat: add configurable gradient clipping to prevent exploding gradients

### LINKED ISSUE #280: Add configurable gradient clipping to prevent exploding gradients during autoregressive training
## Description
neural-lam's training pipeline has no gradient clipping configured anywhere — neither via PyTorch Lightning's built-in Trainer(gradient_clip_val=...) nor through manual torch.nn.utils.clip_grad_norm_() calls.

This is particularly risky because of how ARModel.unroll_prediction works:

``` python
# neural_lam/models/ar_model.py  L230-277
def unroll_prediction(self, init_states, forcing_features, true_states):
    for i in range(pred_steps):          # <-- up to ar_steps sequential passes
        pred_state, pred_std = self.predict_step(
            prev_state, prev_prev_state, forcing
        )
        new_state = (
            self.boundary_mask * border_state
            + self.interior_mask * pred_state
        )
        prev_prev_state = prev_state
        prev_state = new_state             # <-- gradient chain grows each step
```
Each autoregressive step feeds predictions back as input, creating a computational graph ar_steps layers deep. When --ar_steps_train is set to values like 10–25 (common for multi-day weather forecasting), backpropagation through this chain multiplies gradients at each step — the classic exploding gradient problem from recurrent architectures.

The current Trainer in [train_model.py](https://github.com/mllam/neural-lam/blob/main/neural_lam/train_model.py#L317-L329) has no clipping:

``` python
trainer = pl.Trainer(
    max_epochs=args.epochs,
    deterministic=True,
    strategy="ddp",
    # ...
    precision=args.precision,
    # No gradient_clip_val
    # No gradient_clip_algorithm
)
```
## Why it matters
Training instability with longer rollouts: Users increasing --ar_steps_train beyond the default of 1 (e.g., for curriculum learning or multi-step loss) will encounter NaN losses or sudden loss spikes. This is a silent failure — training continues but the model is destroyed.

Standard practice in weather-ML: All major published models use gradient clipping:

GraphCast (Lam et al., 2023): gradient clipping at max norm = 32
Pangu-Weather (Bi et al., 2023): gradient clipping with max norm = 1.0
GenCast (Price et al., 2024): gradient clipping at max norm = 32
FourCastNet (Pathak et al., 2022): gradient clipping with max norm = 1.0
Mixed precision amplifies the risk: neural-lam supports --precision 16 and bf16. Lower precision reduces the representable dynamic range, making gradient overflow happen sooner and at smaller ar_steps values.

Zero-cost safeguard: Gradient clipping adds negligible computational overhead (a single torch.norm() + conditional scale per parameter group per step).

abhaygoudannavar:
## Describe your changes
Add configurable gradient clipping to neural-lam's training pipeline to prevent exploding gradients during multi-step autoregressive unrolling.

## Changes:
- neural_lam/train_model.py: Add --gradient_clip_val (default 1.0, set to 0 to disable) and --gradient_clip_algorithm (norm/value) CLI arguments. 
Wire both to pl.Trainer(gradient_clip_val=...,gradient_clip_algorithm=...).
- neural_lam/models/ar_model.py: Add on_before_optimizer_step()
Lihtning hook that computes and logs the pre-clipping gradient L2 norm (grad_norm) to
WandB/MLflow for training diagnostics.
- tests/test_gradient_clipping.py (new): 5 tests — 4 unit tests verifying Trainer config forwarding (enabled, disabled, norm algorithm, value algorithm) + 1 end-to-end integration test running 1 epoch with clipping enabled and asserting no NaN loss.

- Motivation: ARModel.unroll_prediction chains up to ar_steps sequential GNN forward passes with full gradient backpropagation. When --ar_steps_train is increased beyond the default (e.g. for curriculum learning), this creates a deep computational graph prone to exploding gradients — the classic problem from recurrent architectures. All major published weather-ML models (GraphCast, Pangu-Weather, GenCast, FourCastNet) use gradient clipping. PyTorch Lightning natively supports this via Trainer arguments, making the change minimal and non-invasive.

- Dependencies: None — uses only built-in PyTorch Lightning functionality.

Issue Link
Closes #280 

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/neural_lam/models/ar_model.py
+++ b/neural_lam/models/ar_model.py
@@ -208,8 +208,7 @@ class ARModel(pl.LightningModule):
         """Log the total gradient L2 norm before clipping for diagnostics.
 
         This uses max_norm=inf so the norm is computed without any actual
-        clipping, giving the *pre-clip* value for monitoring in
-        WandB/MLflow.
+        clipping, giving the *pre-clip* value for monitoring.
         """
         grad_norm = torch.nn.utils.clip_grad_norm_(
             self.parameters(), max_norm=float("inf")
--- a/neural_lam/models/ar_model.py
+++ b/neural_lam/models/ar_model.py
@@ -204,6 +204,25 @@ class ARModel(pl.LightningModule):
         )
         return opt
 
+    def on_before_optimizer_step(self, optimizer):
+        """Log the total gradient L2 norm before clipping for diagnostics.
+
+        This uses max_norm=inf so the norm is computed without any actual
+        clipping, giving the *pre-clip* value for monitoring in
+        WandB/MLflow.
+        """
+        grad_norm = torch.nn.utils.clip_grad_norm_(
+            self.parameters(), max_norm=float("inf")
+        )
+        self.log(
+            "grad_norm",
+            grad_norm,
+            on_step=True,
+            on_epoch=False,
+            prog_bar=False,
+            sync_dist=False,
+        )
+
     @property
     def interior_mask_bool(self):
         """
--- a/neural_lam/train_model.py
+++ b/neural_lam/train_model.py
@@ -146,8 +146,8 @@ def main(input_args=None):
     parser.add_argument(
         "--gradient_clip_val",
         type=float,
-        default=1.0,
-        help="Max norm for gradient clipping (0 = disabled). "
+        default=None,
+        help="Max norm for gradient clipping (default: None = disabled). "
         "Prevents exploding gradients during multi-step "
         "autoregressive unrolling.",
     )
@@ -356,10 +356,12 @@ def main(input_args=None):
         callbacks=[checkpoint_callback],
         check_val_every_n_epoch=args.val_interval,
         precision=args.precision,
-        gradient_clip_val=(
-            args.gradient_clip_val if args.gradient_clip_val > 0 else None
+        gradient_clip_val=args.gradient_clip_val,
+        gradient_clip_algorithm=(
+            args.gradient_clip_algorithm
+            if args.gradient_clip_val is not None
+            else None
         ),
-        gradient_clip_algorithm=args.gradient_clip_algorithm,
     )
 
     # Only init once, on rank 0 only
--- a/neural_lam/train_model.py
+++ b/neural_lam/train_model.py
@@ -144,6 +144,23 @@ def main(input_args=None):
     )
     parser.add_argument("--lr", type=float, default=1e-3, help="learning rate")
     parser.add_argument(
+        "--gradient_clip_val",
+        type=float,
+        default=1.0,
+        help="Max norm for gradient clipping (0 = disabled). "
+        "Prevents exploding gradients during multi-step "
+        "autoregressive unrolling.",
+    )
+    parser.add_argument(
+        "--gradient_clip_algorithm",
+        type=str,
+        default="norm",
+        choices=["norm", "value"],
+        help="Gradient clipping algorithm: 'norm' clips by total gradient "
+        "norm (torch.nn.utils.clip_grad_norm_), 'value' clips each "
+        "gradient element (torch.nn.utils.clip_grad_value_).",
+    )
+    parser.add_argument(
         "--val_interval",
         type=int,
         default=1,
@@ -326,6 +343,10 @@ def main(input_args=None):
         callbacks=[checkpoint_callback],
         check_val_every_n_epoch=args.val_interval,
         precision=args.precision,
+        gradient_clip_val=(
+            args.gradient_clip_val if args.gradient_clip_val > 0 else None
+        ),
+        gradient_clip_algorithm=args.gradient_clip_algorithm,
     )
 
     # Only init once, on rank 0 only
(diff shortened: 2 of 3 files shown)


abhaygoudannavar:
@joeloskarsson are there any changes need to be done in this PR.

neural_lam/models/ar_model.py
```diff
@@ -204,6 +204,25 @@ def configure_optimizers(self):
         )
         return opt
 
+    def on_before_optimizer_step(self, optimizer):
+        """Log the total gradient L2 norm before clipping for diagnostics.
+
+        This uses max_norm=inf so the norm is computed without any actual
+        clipping, giving the *pre-clip* value for monitoring in
+        WandB/MLflow.
```
joeloskarsson:
```suggestion
        clipping, giving the *pre-clip* value for monitoring.
```
Let's be open to arbitrary logging frameworks in the future.

neural_lam/train_model.py
```diff
@@ -143,6 +143,23 @@ def main(input_args=None):
         help="Loss function to use, see metric.py",
     )
     parser.add_argument("--lr", type=float, default=1e-3, help="learning rate")
+    parser.add_argument(
+        "--gradient_clip_val",
+        type=float,
+        default=1.0,
```
joeloskarsson: I really don't think we want this on per default. That is a major change to default behavior. Note that first step of training typically does not use unrolling, where grad clipping is mainly useful

neural_lam/train_model.py
```diff
@@ -143,6 +143,23 @@ def main(input_args=None):
...
+        help="Max norm for gradient clipping (0 = disabled). "
+        "Prevents exploding gradients during multi-step "
+        "autoregressive unrolling.",
+    )
```
joeloskarsson: I think having `None` be the setting for turning off clipping would be more intuitive. 0 to me feels like clipping gradients to be 0.

joeloskarsson: This also makes it match the lightning argument

tests/test_gradient_clipping.py
```diff
@@ -0,0 +1,148 @@
...
+    assert trainer.gradient_clip_algorithm == "value"
+
+
+def test_gradient_clip_algorithm_norm():
+    """Verify that the 'norm' algorithm is properly forwarded."""
+    trainer = _build_trainer(
+        gradient_clip_val=2.5, gradient_clip_algorithm="norm"
+    )
+    assert trainer.gradient_clip_val == 2.5
+    assert trainer.gradient_clip_algorithm == "norm"
```
joeloskarsson: As far as I can see these tests do not actually use any method from neural-lam, so they are not actually testing our codebase?

tests/test_gradient_clipping.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def test_training_with_gradient_clipping():
+    """End-to-end: run 1 epoch with gradient clipping, assert no NaN loss."""
```
joeloskarsson: This seems very expensive to test this, and I am far from convinced that you would get a NaN loss even without grad clipping. It should be sufficient to check that grads do get clipped in one step.

joeloskarsson (changes requested):
I think this option is a nice addition to the codebase. I left some comments with ideas on how to best specify this option. Also I think the testing needs another look.

neural_lam/models/ar_model.py
abhaygoudannavar: updated the docstring to simply say "for monitoring" without referencing any specific logging framework.

neural_lam/train_model.py
abhaygoudannavar: Agreed — I've changed the default to None (disabled). Gradient clipping now has to be explicitly opted-in via --gradient_clip_val <value>. This keeps the existing default behavior unchanged.

neural_lam/train_model.py
abhaygoudannavar: Makes sense — None is now used to disable clipping, matching Lightning's native argument convention. The 0-based disable logic has been removed; gradient_clip_val is passed directly to pl.Trainer.

tests/test_gradient_clipping.py
abhaygoudannavar: You're right — the original unit tests were just testing pl.Trainer constructor behavior. I've replaced them with tests that exercise actual neural-lam code: one uses unittest.mock.patch on train_model.main() to verify CLI args are correctly forwarded to the Trainer, and the other sets up a real GraphLAM model with DummyDatastore data to verify clipping works end-to-end.

tests/test_gradient_clipping.py
abhaygoudannavar: Agreed — I've replaced the full-epoch training test with a lightweight max_steps=1 test that runs a single training step with gradient clipping enabled and verifies the loss is valid. This is much cheaper while still confirming that clipping integrates correctly with our model's training loop.

PR #291, opened 2026-02-28
feat: add NWP-standard evaluation metrics for global forecasting

santhil-cyber:
## Issue Link

Relates to #63

## Type of change

### DIFF
--- a/neural_lam/evaluation/__init__.py
+++ b/neural_lam/evaluation/__init__.py
@@ -1,0 +1,18 @@
+# Local
+from .verification_metrics import (
+    acc,
+    compute_grid_weights,
+    latitude_weighted_rmse,
+    spread_skill_ratio,
+    weighted_mae,
+    weighted_rmse,
+)
+
+__all__ = [
+    "acc",
+    "latitude_weighted_rmse",
+    "weighted_rmse",
+    "weighted_mae",
+    "spread_skill_ratio",
+    "compute_grid_weights",
+]
--- a/neural_lam/evaluation/verification_metrics.py
+++ b/neural_lam/evaluation/verification_metrics.py
@@ -1,0 +1,326 @@
+"""
+verification metrics for global/regional weather model evaluation.
+
+implements standard NWP verification scores (ACC, weighted RMSE, etc.)
+that complement the training-time metrics in neural_lam.metrics.
+
+tensor conventions match neural_lam.metrics:
+    pred/target: (..., N, d_state)
+    grid_weights: (N,)
+    mask: (N,) boolean
+"""
+
+# Standard library
+import math
+
+# Third-party
+import torch
+
+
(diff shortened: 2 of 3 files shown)


santhil-cyber:
Just to add some context on why these metrics matter for the global forecasting work in #63 —

Right now neural-lam only has training-time metrics (`mse`, `mae`, `nll`, `crps_gauss`). These are fine for optimizing the loss, but when you actually want to evaluate whether a global model is producing skillful forecasts, the NWP community uses quite different diagnostics:

- **ACC** is what ECMWF, NCEP etc. use as their primary headline score. The usual benchmark is ACC > 0.6 at day 5 for Z500 — that's roughly where a forecast stops being "useful". WeatherBench2 also reports this.
- **Latitude-weighted RMSE** is important because on a regular lat-lon grid the poles are massively oversampled. Without area weighting, a model can get low RMSE just by nailing the poles while being garbage at the tropics — where it actually matters for weather.
- **Spread-skill ratio** is needed once we use the probabilistic output (`pred_std`). A ratio < 1 means the model is overconfident, > 1 means overdispersive. Useful for calibration tuning.

The module is designed to work with both regional and global setups — `compute_grid_weights` handles regular lat-lon grids (cos-latitude weighting) and equal-area grids like icosahedral meshes where all cells are roughly the same size.

Happy to integrate these into `test_step` / the validation pipeline if that's useful, or keep them as a standalone evaluation module — whatever makes more sense architecturally. @joeloskarsson @sadamov @leifdenby

joeloskarsson:
Hi, I disagree that the metrics in `metrics.py` are only for training-time, and would argue that they are also functional as evaluation metrics (this is indeed what we use now). To make this point further, there is lots of overlaps between the code introduced here and `metrics.py`.

In general, the introduction of a evaluation sub-module is a large enough change that it would have to be anchored within the community before introducing a PR like this. For this reason I will mark this as a draft PR for now and urge you to start a discussion in an issue or on slack if you feel strongly about a need for a verification module.

santhil-cyber:
thanks for the feedback, that's fair. you're right that the current metrics work for evaluation too, i framed that poorly. and yeah the overlap with `mask_and_reduce_metric` is real, should have extended the existing code instead of duplicating it.

thinking about it more, the less invasive approach would be:

1. add an optional `grid_weights` param to `mask_and_reduce_metric` so all existing metrics automatically support area weighting
2. add ACC as a new metric function in `metrics.py` alongside the existing ones

that way there's no new sub-module, just extending what's already there. i'll open an issue to discuss the approach before writing more code. thanks for the guidance!

sadamov:
Closing this in favour of #650, which consolidates the global forecasting / spherical domain track for v0.8.0. The NWP-standard evaluation metrics for global forecasting will be folded into the consolidated PR. Thanks for the contribution @santhil-cyber, you'll be credited.

PR #255, opened 2026-02-25
Fix unhandled metric index crash and malformed clamping error

Ayushhgit:
## Describe your changes

**Summary of the changes:**
- Added a bound check for the `--var_leads_metrics_watch` JSON argument in [train_model.py] to assert that users are asking for variables that physically exist inside the array.
- Fixed an assertion error comparison string in [base_graph_model.py]regarding clamping bounds failing due to multi-line string interpolation causing broken tab indentation logging.

**Motivation and context:**
The `metrics_watch` argument parsing was deferring bounds-checking to heavily inside the validation epoch. Because training massive graphs takes a long time, users passing an invalid JSON argument variable coordinate would train their model for an hour and then hit an `IndexError` completely destroying the run, rather than failing immediately when the configuration was loaded. Furthermore, the [base_graph_model.py]bounds clamping exception incorrectly stated properties were "larger than" each other when the assertion actually triggered from bounds equality (`<=`), and outputted visual garbage due to missing text wrappers.

**Dependencies:**
None

## Issue Link

N/A

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [ ] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [ ] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the README to cover introduced code changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented (including return types and parameters)
- [ ] the code is easy to maintain

[...]

PR #306, opened 2026-03-02
Add bounds checking for var_leads_metrics_watch indices

Ayushhgit:
## Describe your changes
**Summary of the changes:**
Added a bound check for the `--var_leads_metrics_watch` argument in `train_model.py` to assert that users are asking for variables that physically exist inside the array.

**Motivation and context:**
The `metrics_watch` argument parsing was deferring bounds-checking to heavily inside the validation epoch. Because training massive graphs takes a long time, users passing an invalid JSON argument variable coordinate would train their model for an hour and then hit an IndexError completely destroying the run, rather than failing immediately when the configuration was loaded. 

## Issue Link
Splits #255

## Type of change


sadamov:
@Ayushhgit please use the original PR template

sadamov (approved):
okay this one seems abandoned, but the small change implemented is actually very useful. I ran into missing val_lead in datastore multiple times. merging

PR #315, opened 2026-03-02
Add unit tests for utils.get_integer_time

ML-dev-crypto:
Adds unit tests for get_integer_time covering documented examples:
- Week conversion
- Hour conversion
- Millisecond promotion to seconds
- Documented `(1, "unknown")` fallback case

Tests align with the existing docstring examples and verify the unit priority order.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,20 @@
+from datetime import timedelta
+
+from neural_lam.utils import get_integer_time
+
+
+def test_weeks():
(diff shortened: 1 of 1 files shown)


observingClouds:
This is already [covered by the doctests](https://github.com/mllam/neural-lam/blob/3d5d56b8d0562c7074a0a1ec33036ae9eaafa319/neural_lam/utils.py#L425) of the function.

PR #316, opened 2026-03-02
Add optional grid_weights to mask_and_reduce_metric

santhil-cyber:
## Describe your changes

Added an optional `grid_weights` parameter to [mask_and_reduce_metric](cci:1://file:///Users/santhil/neural-lam/neural_lam/metrics.py:20:0-52:28)
in [neural_lam/metrics.py](cci:7://file:///Users/santhil/neural-lam/neural_lam/metrics.py:0:0-0:0). When `average_grid=True` and `grid_weights`
is provided, the function computes a weighted mean over the grid
dimension instead of `torch.mean`. Weights are correctly subsetted when
a boolean [mask](cci:1://file:///Users/santhil/neural-lam/neural_lam/models/ar_model.py:206:4-211:54) is also applied.

**Motivation:** For global weather forecasting on regular lat-lon grids,
grid nodes represent different physical areas — a node at the equator
covers ~5× more area than one at 80°N. Without area weighting,
`torch.mean` over-represents polar regions. This is the standard
approach used by WeatherBench2 and operational NWP centres.

All 6 existing metric functions now accept and forward `grid_weights`
as a keyword argument. No existing function signatures are changed —
`grid_weights` defaults to `None`, preserving exact existing behavior.

Added [tests/test_metrics.py](cci:7://file:///Users/santhil/neural-lam/tests/test_metrics.py:0:0-0:0) with 14 tests covering backward
compatibility, weighted mean correctness, cos-latitude weighting,
mask+weights interaction, and all 6 metric functions.

No new dependencies required.

## Issue Link

Refs #313

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [x] I have added tests that prove my fix is effective or that my feature works
[...]

PR #323, opened 2026-03-03
[Feat] Add graph-on-disk format specification and validator

### LINKED ISSUE #408: Add a small CLI to sanity‑check generated graph files
## Summary

When `create_graph` runs, it writes a bunch of `.pt` files (edge indices, edge features, mesh features, etc.) to disk. At the moment we only really check that these files are self‑consistent inside the tests (for example in `tests/test_graph_creation.py`). As a user there is no easy way to point at an existing graph directory and ask “is this graph actually OK?” without running the test suite.

I’d find it useful to have a small command that can validate a graph directory and print a short report if something looks wrong.

## Idea

Rough sketch of what I had in mind:

- Add a tiny helper, e.g. `neural_lam/validate_graph.py`, and expose it as:

  ```bash
  python -m neural_lam.validate_graph --graph_dir <path/to/graph_dir>

### LINKED ISSUE #594: [Bug] Division by zero in create_graph mesh position normalization
In 'neural_lam/create_graph.py', mesh points are normalized by dividing by the grid's maximum absolute coordinate value ('pos_max'). If a degenerate grid (all zeros) or extremely small spatial values are processed, 'pos_max' evaluates to '0', causing a division-by-zero arithmetic mistake and propagation of 'NaN' or 'Inf' across the model.

### Root Cause

```python
# neural_lam/create_graph.py (line 237)
pos_max = torch.max(torch.abs(grid_xy))

...

# neural_lam/create_graph.py (line 413) - BUGGY
# If pos_max is 0, this divides by zero.
mesh_pos = [pos / pos_max for pos in mesh_pos]

### PR DESCRIPTION
**Pull request by leifdenby**
## Describe your changes

To facilitate bridging the gap between [weather-model-graphs](github.com/mllam/weather-model-graphs) (`wmg`) and `neural-lam` so that it is possible to generate graphs in `wmg`, save them to disk and use them in `neural-lam`, this PR adds two things:

- [A detailed specification](https://github.com/leifdenby/neural-lam/blob/feat/graph-on-disk-spec-and-validator/docs/graph_storage_spec.md) for the `pytorch.Tensor` pickled to disk format that `neural-lam` currently expects, and has been used since `neural-lam` release `v0.1.0`. 
- [A implementation of the spec in form of a validator](https://github.com/leifdenby/neural-lam/blob/feat/graph-on-disk-spec-and-validator/docs/validate_graph.py) that is stored independently from the `neural-lam` codebase with using [PEP 723 compliant inline script dependencies](https://peps.python.org/pep-0723/) so that the validator can be executed directly [with uv](https://docs.astral.sh/uv/guides/scripts/) from the code stored in github like so:
```bash
uv run \
  https://raw.githubusercontent.com/leifdenby/neural-lam/refs/heads/feat/graph-on-disk-spec-and-validator/docs/validate_graph.py \
 tests/datastore_examples/mdp/danra_100m_winds/graph/hier3
```

**NB**: the validator only works for graphs that actually match the specification, which means graphs created with `neural_lam.create_graph` in the branch of this PR (since that now follows the specification version `0.1.0`, e.g. to create the above graph run:

```bash
uv run python -m neural_lam.create_graph --config_path tests/datastore_examples/mdp/danra_100m_winds/config.yaml --name hier3 --levels 3 --hierarchical
```

Screenshot 2026-06-29 at 20 53 25

I also integrated running the validator in a new CI test that runs creation and validation of each of the three archetypes (flat, flat-multiscale and hierarchical).

The key motivation here is that currently it is not possible to create graphs for `neural-lam` in `wmg`, although the functionality is nearly complete ([`wmg.save.to_pyg`](https://github.com/mllam/weather-model-graphs/blob/main/src/weather_model_graphs/save.py#L24)) but I think there may still be issues with it (I will link issues here when I find them)

There are no change of dependencies required.

## Issue Link

No link yet, but I will write up an issue on the state of creating graphs with `wmg` and using them in `neural-lam` and possible approaches.

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


docs/graph_storage_spec.md
```diff
@@ -0,0 +1,188 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in RFC 2119.
+
+## 2. File and Directory Structure
+
+### Directory Structure
+
+Each graph MUST live in a directory (typically
+`<datastore-root>/graph/<name>`).
```
joeloskarsson: It must be the datastore root for us to be able to load them, right?

docs/graph_storage_spec.md
```diff
@@ -0,0 +1,188 @@
...
+architectures follow the encode-process-decode paradigm of sequential message
+passing, where physical variables are represented as features on so-called
+*grid* nodes, are *encoded* to *mesh* nodes, are *processed* on the mesh, and
+then *decoded* back to grid nodes where output tendencies or updated state are
+produced.
+
+The format specified in this document was designed to support the three archetypes of GNN-based MLWP forecasting architectures:
+
+- flat, non-hierarhical graphs where the processing mesh only connects to nearest neighbors (Keisler et al. 2022)
+- flat-multiscale, non-hierarchical graphs where the processing mesh also includes longer-range connections (GraphCast, Lam et al. 2022)
```
joeloskarsson: Note that neural-lam has no idea about flat vs flat-multiscale (these are just different adjacency structures), so might not be a point in discussing here.

docs/graph_storage_spec.md
```diff
@@ -0,0 +1,188 @@
...
+### Nodes
+
+The `neural-lam` graph format on disk does not explicitly store node features for grid nodes, as these are expected to be dynamic and stored separately in the dataset. However, static features for mesh nodes MUST be stored in `mesh_features.pt` files (as described below).
+
+#### Node index space
+
+Node indices in edge tensors MUST be stored in a single global index space:
+
+- Mesh nodes MUST come first.
+- Grid nodes MUST follow after all mesh nodes.
```
joeloskarsson: This is changing in #184

docs/graph_storage_spec.md
```diff
@@ -0,0 +1,188 @@
...
+- For hierarchical graphs, each mesh level MUST occupy a contiguous mesh index
+  range in ascending level order.
```
joeloskarsson: We don't require this. Each mesh level is treated independently, its indices relative to the other levels do not matter.

docs/graph_storage_spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Mesh node features
+
+`mesh_features.pt` files MUST be lists of length `L` (number of mesh levels),
+where each entry is a tensor containing static features for the mesh nodes at
+that level. Each tensor MUST satisfy the following requirements:
+
+- `mesh_features` entries MUST have shape `[N_level, 2]`, where `N_level` is the number of mesh nodes at that level.
+- Columns MUST be x/y coordinates.
```
joeloskarsson: Do we really require these to be specifically coordinates now? If so, we should specify the CRS of these.

docs/graph_storage_spec.md
```diff
@@ -0,0 +1,188 @@
...
+- In graphs created by `neural_lam.create_graph`, coordinates are normalized by
+  dividing by the maximum absolute grid coordinate.
+- Dtype MUST be floating-point.
```
joeloskarsson: More specific

docs/graph_storage_spec.md
```diff
@@ -0,0 +1,188 @@
...
+tensors, i.e. one per inter-level connection, so that entry `i` connects level
+`i` and level `i+1`. Each entry MUST have shape `[2, E_interlevel]`, where
+`E_interlevel` is the number of edges going either up
+(`mesh_up_edge_index.pt`) or down (`mesh_down_edge_index.pt`) between that
+level pair.
+
+For every edge-index tensor above:
+
+- Row `0` MUST be sender node index, row `1` MUST be receiver node index.
+- Dtype MUST be integer.
```
joeloskarsson: Must be long

docs/graph_storage_spec.md
leifdenby: Yes, true, because the current implementation in `neural-lam` derives them from the datastore and places them next to the datastore. I will change this text. I hadn't considered what the path should be relative to what `neural-lam` expects, but you're right.

docs/graph_storage_spec.md
leifdenby: True, but I think it would be good to document what we have now, and then revise the spec as we improve things. Of course if #184 is merged before this PR then I will adapt the formulation here before this PR is merged :)

docs/graph_storage_spec.md
leifdenby: ok, there are two statements here `each mesh level MUST occupy a contiguous mesh index
  range` and `in ascending level order` the former is required no? It would make sense to me based on what you wrote that the latter is not required. But I still expect that each mesh level must cover non-overlapping mesh node index space, so (writing just source nodes for now) `[[1,2,3], [4,5,6]]` would be allowed for a two-level mesh, but `[[1,3,5], [2,4,6]]` would not be allowed. Or is that not true and the node indices for each mesh level are allowed to overlap in their range?

docs/graph_storage_spec.md
leifdenby: I am not sure. I think we do/did some special normalisation of these features, and so would thereby assume what the features represent physically. But maybe that is not the case. But maybe all mesh features are just treated identically and scaled the same way.

docs/graph_storage_spec.md
```diff
@@ -0,0 +1,188 @@
...
+- For hierarchical graphs, each mesh level MUST occupy a contiguous mesh index
+  range in ascending level order.
+
+#### Mesh node features
+
+`mesh_features.pt` files MUST be lists of length `L` (number of mesh levels),
+where each entry is a tensor containing static features for the mesh nodes at
+that level. Each tensor MUST satisfy the following requirements:
+
+- `mesh_features` entries MUST have shape `[N_level, 2]`, where `N_level` is the number of mesh nodes at that level.
```
leifdenby: Re-reading this now I am also not sure that we require what number of mesh features there are and what they physically should be. Maybe we should say what create_graph did by default in neural-lam here in terms of what the physical features were and the order, but maybe we don't need a requirement here.

docs/graph_storage_spec.md
leifdenby: I don't quite understand what you mean here :) Are you saying the spec text needs to be more specific?

docs/graph_storage_spec.md
leifdenby: I am realising it would be better to write the dtype with reference to `torch` itself, so this should be `torch.int64` (which `torch.long` is aliased to, but using `torch.int64` avoids people needing to know that)

leifdenby:
Thanks for the feedback @joeloskarsson! I will address your comments by revising the spec and pushing a new revision once it is done

santhil-cyber:
Great to see this spec formalized @leifdenby — this is exactly the interface my WP0 (global graph integration) needs to consume on the neural-lam side.

Two questions from the integration perspective:

**1. `boundary_mask` for global graphs**

Does the spec require a `boundary_mask` tensor in the graph directory? For regional LAM, this separates interior vs boundary nodes. For global forecasting (icosahedral mesh from WMG), all nodes are interior — there is no boundary. On the neural-lam side, I'm planning to handle this by detecting `boundary_mask.sum() == 0` to skip boundary forcing (per Hauke's guidance). But should the spec:

- **(a)** Require `boundary_mask` to always be present (all-`False` for global), or
- **(b)** Make it optional (absent = global/no boundary)?

Option (a) is simpler for neural-lam's loading code — no `FileNotFoundError` handling needed.

**2. Optional node attributes (e.g. `grid_weights` / `node_area`)**

For area-weighted metrics (#258, #316), neural-lam needs per-node area weights. Should the spec include an optional `grid_weights` or `node_area` tensor? This would let WMG compute exact cell areas (e.g. from Voronoi dual) and pass them through the graph-on-disk format, rather than neural-lam approximating with `cos(lat)`.

Joltsy10:
>Does the spec require a boundary_mask tensor in the graph directory? For regional LAM, this separates interior vs boundary nodes. For global forecasting (icosahedral mesh from WMG), all nodes are interior — there is no boundary. On the neural-lam side, I'm planning to handle this by detecting boundary_mask.sum() == 0 to skip boundary forcing (per Hauke's guidance). But should the spec:

option (a) is the right call. From the neural-lam side, ARModel.__init__ registers boundary_mask as a buffer unconditionally with no optional loading path. A missing file would raise a FileNotFoundError at model construction rather than gracefully handling the global case. Always writing boundary_mask as all-False for global graphs is the safe contract and keeps the loading code simple.

>For area-weighted metrics (https://github.com/mllam/neural-lam/pull/258, https://github.com/mllam/neural-lam/pull/316), neural-lam needs per-node area weights. Should the spec include an optional grid_weights or node_area tensor? This would let WMG compute exact cell areas (e.g. from Voronoi dual) and pass them through the graph-on-disk format, rather than neural-lam approximating with cos(lat).

I'd lean against including it in the graph-on-disk spec. The spec's job is topology and geometry. Weight computation for loss functions is a training policy decision that belongs in ARModel, not in the graph format. Joel's feedback on PR #258 pointed in this direction too, optional cos(lat) weights configurable via the training config, owned at the model level. Coupling weight computation to the graph format means every graph generator needs to know about downstream loss policy, which feels like the wrong separation.

docs/graph_storage_spec.md
Joltsy10: On the mesh_features physical meaning question, as far as I can tell from code inspection, utils.load_graph normalizes m2m edge features internally by dividing by longest_edge taken from column 0, so the spec does carry an implicit physical constraint there: column 0 of edge features must be a raw unnormalized distance. 
```python
# Normalize by dividing with longest edge (found in m2m)
    longest_edge = max(
        torch.max(level_features[:, 0]) for level_features in m2m_features
    )  # Col. 0 is length
```

mesh_features are different though. BaseHiGraphModel passes them directly into a linear embedding layer with no normalization or physical assumption. So the spec probably doesn't need to constrain what mesh_features represent physically, they're treated as arbitrary learned inputs by the model.

```python
mesh_dim = self.mesh_static_features[0].shape[1]
...
self.mesh_embedders = nn.ModuleList(
    [
        utils.make_mlp([mesh_dim] + self.mlp_blueprint_end)
        for _ in range(self.num_levels)
    ]
)
```

Joltsy10: Looking at create_graph.py, the [N_level, 2] in the spec is documenting what create_graph currently happens to produce. It saves pyg_m2m.pos which is normalized x/y mesh node positions, rather than what the model actually requires. load_graph loads mesh_features with no shape check or physical assumption, and BaseHiGraphModel reads mesh_static_features[0].shape[1] dynamically to set the MLP input size. For icosahedral global graphs, 3D Cartesian coordinates would be a natural choice, shape [N_level, 3], and the loading and embedding code would handle it without modification
The pipelines looks like:
In create_graph.py:
```python
mesh_pos = [pyg_m2m.pos.to(torch.float32)]
...
torch.save(mesh_pos, os.path.join(graph_dir_path, "mesh_features.pt"))
```

in utils.load_graph:
```python
mesh_static_features = loads_file("mesh_features.pt")  # List of (N_mesh[l], d_mesh_static)
```

in BaseHiGraphModel:
```python
mesh_dim = self.mesh_static_features[0].shape[1]
```

Together these three form a clean chain, create_graph produces 2, load_graph doesn't care what it is, BaseHiGraphModel adapts to whatever it gets.

docs/validate_graph.py
```diff
@@ -0,0 +1,662 @@
...
+from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser
+from dataclasses import asdict, dataclass, field
+from pathlib import Path
+from typing import Any
+
+# Third-party
+import torch
+
+EDGE_FEATURE_DIM = 3
+MESH_FEATURE_DIM = 2
```
Joltsy10: Two things worth flagging from an icosahedral global graph perspective, curious if these are intentional design choices or worth revisiting.
MESH_FEATURE_DIM = 2 is hardcoded as a hard error in the validator. From code inspection, BaseHiGraphModel reads mesh_static_features[0].shape[1] dynamically to set the MLP input size, so the model itself doesn't require exactly 2 features. For icosahedral meshes, 3D Cartesian coordinates would be natural, giving [N_level, 3]. Is the intent to document the current create_graph convention specifically, or to specify what the model actually requires? If the former, would it make sense to downgrade this to a warning rather than a hard error to leave room for other mesh types?
Same question for EDGE_FEATURE_DIM = 3. The vdiff consistency check assumes [length, dx, dy] which is the 2D rectilinear convention. For icosahedral 3D edges the natural representation would be [length, dx, dy, dz]. Both constants would cause the validator to reject graphs that the model can actually load and train on, wondering if that's the intended scope of the spec or something worth making flexible.

Joltsy10: Following up on this after seeing @leifdenby  response to Himanshu on node coordinates in slack. Happy to concede the MESH_FEATURE_DIM = 2 point. Node positions on the sphere are fully described by 2D spherical coordinates, so 2 is correct for global icosahedral graphs as well, just with lat/lon rather than normalized x/y. That's a fair clarification.

The EDGE_FEATURE_DIM case is different though, and I think worth keeping as an open question for three reasons specific to the edge vdiff.

First, directional information. Weather dynamics are physically directional. Wind, pressure gradients, moisture transport all have direction in 3D space. When the GNN computes messages along edges, edge features tell the network which way information is flowing spatially. Two edges with identical (Δlat, Δlon) = (0°, 5°), one near the equator and one near the pole, represent completely different physical displacements: roughly 550km eastward near the equator versus maybe 50km near the pole. The angular offset is the same but the physical direction and magnitude are not. 3D Cartesian vdiff (Δx, Δy, Δz) encodes a consistent physical direction vector regardless of where on the sphere the edge sits. With (Δlat, Δlon) the network has to learn to compensate for spherical distortion implicitly on top of learning the actual dynamics.

Second, discontinuity. (Δlat, Δlon) breaks at the date line. Two nodes at lon=179° and lon=179° are physically adjacent but their Δlon is 358°. At the poles Δlon loses meaning entirely. 3D Cartesian vdiff is smooth and continuous everywhere on the sphere.

Third, implementation evidence. This was the exact issue identified and fixed in wmg PR 76 <https://github.com/mllam/weather-model-graphs/pull/76> where the original vdiff was computed in 2D and corrected to 3D Cartesian for precisely these reasons.

The proposal would be to keep MESH_FEATURE_DIM = 2 since node positions in 2D spherical is correct, and update EDGE_FEATURE_DIM = 4 for global icosahedral graphs ([length, Δx, Δy, Δz]). The two cases are genuinely different since one is about position completeness and the other is about displacement representation. I don't think the response to Himanshu on node coordinates fully addresses the edge vdiff question.
Happy to be wrong on this if there's a reason the 2D angular convention is preferred for edges in the global case.

docs/graph_storage_spec.md
leifdenby: Addressed in https://github.com/mllam/neural-lam/pull/323/commits/56d9776a005dd3d637f8873e5d830b39d43d37a6

docs/graph_storage_spec.md
leifdenby: Made text more general in https://github.com/mllam/neural-lam/pull/323/commits/4f564ea50a34e2911df40db3eea113483dd19fdf and removed distinction between flat and flat-multiscale graphs (because there is none wrt the message passing that takes place, as you said)

docs/graph_storage_spec.md
leifdenby: As agreed on slack we will leave the text as-is for now, so that it describes what is currently on `main`. #184 has been merged we should update this text (added https://github.com/mllam/neural-lam/pull/184#issuecomment-4012574821 to help us remember this)

docs/graph_storage_spec.md
leifdenby: I miss-read https://github.com/mllam/neural-lam/blob/main/neural_lam/utils.py#L118 where we load the graph and normalization of graph features. I can see that `mesh_static_features` is in fact not scaled when loading, but in the current `create_graph` implementation https://github.com/mllam/neural-lam/blob/main/neural_lam/create_graph.py#L412 we do scale by the `max_pos` (which is the maximum x/y extent of the domain `(0,0)`). Also, in the graph plotting (which is quite hard to reason about though), I think it is assumed that the mesh static features are the `x` and `y` (i.e. projected coordinates): https://github.com/mllam/neural-lam/blob/main/neural_lam/plot_graph.py#L104C9-L104C17. So to some degree those features being `x` and `y` is a requirement of we want the graph plotting code to keep working (this is making me feel more and more like being able to define what each feature actually is would be very helpful).

In summary: as I read this we only assume that the `mesh_static_features` are x/y coordinates for the graph plotting, but we do assume that these features are already normalized when saved to disk. That is different from other features on the mesh part of the graph, say the graph edges, where we normalize the edge lengths when we load the graph: https://github.com/mllam/neural-lam/blob/main/neural_lam/utils.py#L116. I think it would be good to require that either all features are normalized or none of them are.

docs/graph_storage_spec.md
leifdenby: dtypes specified in https://github.com/mllam/neural-lam/pull/323/commits/3278bde19857282e939fb26083af0612afe38b8e

leifdenby:
Summarising on the discussion on the comments you made above @joeloskarsson there are some aspects of the graph-spec that still need discussing:
- normalization of features: `mesh_static_features` [are currently normalised before saving to disk](https://github.com/mllam/neural-lam/blob/main/neural_lam/create_graph.py#L411), but mesh and grid node features [are normalized during graph-loading](https://github.com/mllam/neural-lam/blob/main/neural_lam/utils.py#L155) - what do you think about either normalizing all or none of the graph features? I think it probably makes more sense to not normalize any since we also normalize data at training time, so this approach would mirror how we treat the training data in general
- assumption about mesh features for normalisation: We [currently assume that the 0th feature for `m2m` edges represent the length](https://github.com/mllam/neural-lam/blob/main/neural_lam/utils.py#L117) and use this value to scale all edge features. Would it not make more sense to scale each mesh-feature separately and assume that every each has the same set of features (with different values of course)?
- assumption about static mesh features: the 3D graph plotting currently assumes that the mesh node features represent x and y (spatial coordinates), should we break/remove the 3D graph plotting so that we can relax this requirement? Or should we require what the mesh node features are for now?

docs/validate_graph.py
leifdenby: > 3D Cartesian vdiff (Δx, Δy, Δz) encodes a consistent physical direction vector regardless of where on the sphere the edge sits.

I don't think this is correct unfortunately :) A 3D vertex displacement vector between two nodes on the surface of a unit sphere will be different for every point on the sphere. If you imagine two points along the 0deg longitude, near the equator `(Δx, Δy, Δz) ~ (0, 0, ∆d)` whereas near the north pole `(Δx, Δy, Δz) ~ (0, ∆d, 0)` where `∆d` is the actual distance between the two points.

Instead, I think what we might need here is the `(∆x, ∆y)` displacement in a tangential plane where the y-direction points to the north pole. This way an edge point in the zonal direction (along a latitude) will always have a displacement vector `(∆d, 0)` which makes sense to me physically because this direction points in the direction of rotation of the Earth

In any case: I was aiming to use this PR to document the format that `neural-lam` currently requires. But we should absolutely discuss what would be required to support global icosahedral graphs. Maybe you could create an issue building on this PR describing what is required from your perspective? You could build on #339 by bringing up what you think needs to be considered to handle global icosahdral graphs :) what do you think?

leifdenby:
> Two questions from the integration perspective:
> 
> **1. `boundary_mask` for global graphs**
> 
> Does the spec require a `boundary_mask` tensor in the graph directory? For regional LAM, this separates interior vs boundary nodes. For global forecasting (icosahedral mesh from WMG), all nodes are interior — there is no boundary. On the neural-lam side, I'm planning to handle this by detecting `boundary_mask.sum() == 0` to skip boundary forcing (per Hauke's guidance). But should the spec:
> 
>     * **(a)** Require `boundary_mask` to always be present (all-`False` for global), or
> 
>     * **(b)** Make it optional (absent = global/no boundary)?
> 
> 
> Option (a) is simpler for neural-lam's loading code — no `FileNotFoundError` handling needed.
> 
> **2. Optional node attributes (e.g. `grid_weights` / `node_area`)**
> 
> For area-weighted metrics (#258, #316), neural-lam needs per-node area weights. Should the spec include an optional `grid_weights` or `node_area` tensor? This would let WMG compute exact cell areas (e.g. from Voronoi dual) and pass them through the graph-on-disk format, rather than neural-lam approximating with `cos(lat)`.

@santhil-cyber: Can you explain to me why the `boundary_mask` should be part of this graph-on-disk specification? Currently we have [`boundary_mask` coming from the datastore](https://github.com/mllam/neural-lam/blob/main/neural_lam/datastore/base.py#L267). I think we can keep doing that and still document the format that `neural-lam` currently expects (as in describes what `create_graph` currently creates). If we need to change something wrt `boundary_mask` (which I agree we also need to do when going to a LAM setting, i.e. supporting a datastore for the domain interior and one for the boundary) I don't quite see what that as to do with the current graph format :) Maybe you could create a separate issue for graph-storage requirements when it comes to representing graphs for global forecasting? I think @Joltsy10 you are making the same point here, right?

docs/validate_graph.py
Joltsy10: Thanks for the correction, after looking into it a bit the tangential plane approach makes much more sense physically.

To update the earlier argument on EDGE_FEATURE_DIM: the issue with (Δlat, Δlon) still stands (discontinuity at the date line, distortion away from equator), but the fix is tangential plane displacement rather than raw 3D Cartesian. Edge features for global icosahedral graphs would then be [length, Δx_tangential, Δy_tangential] which keeps EDGE_FEATURE_DIM = 3 but with different semantics than the current rectilinear convention.

Happy to open an issue building on #323 and #339 laying out what global icosahedral graph support would require. On a high level I think it should cover stuff like edge feature convention, mesh node feature convention, boundary_mask requirements, node ordering, and the normalization questions you raised above. Would that be a useful scope for the issue?

Joltsy10:
> Currently we have [`boundary_mask` coming from the datastore](https://github.com/mllam/neural-lam/blob/main/neural_lam/datastore/base.py#L267).



> I don't quite see what that as to do with the current graph format :)


Yes exactly, boundary_mask is determined by the domain topology which the datastore owns, not by how the graph is stored on disk. I'm drafting the issue now and will have it up shortly.

santhil-cyber:
You're right @leifdenby .... I conflated the graph storage format with the datastore's domain topology. `boundary_mask` is a property of the domain (which nodes are interior vs boundary), not of the graph structure itself. The graph spec should stay focused on nodes, edges, and their features.

On the `grid_weights` / `node_area` question same logic applies: if the weights are derived from the domain geometry rather than the graph topology, they belong on the datastore side too. Happy to track that separately.

Thanks for the correction helps me scope the integration work more clearly.

docs/graph_storage_spec.md
joeloskarsson: They are allowed to overlap. There is no relation between the node indices at different levels. So a perfectly ok list of edge_index for the hierarchy is:

```
m2m_edge_index = [
    tensor([
      [0, 1, 2, 2],
      [1, 0, 1, 0]
    ]),
    tensor([
      [0, 1],
      [1, 0]
    ]),
]
```

joeloskarsson: However, they should be contiguous in the sense that at one single level, the indices must lay in the range [start_i, start_i + num_mesh_nodes_at_level], for an arbitrary start_i and with num_mesh_nodes_at_level being the number of nodes on that mesh level.

docs/graph_storage_spec.md
joeloskarsson: Yea I think the longest edge retrieval is the only part that assumes a specific meaning of a feature index here.

docs/graph_storage_spec.md
joeloskarsson: Agreed!

joeloskarsson: And I would opt for none of them being normalized when saved

docs/graph_storage_spec.md
leifdenby: ok, so would it be accurate to say that **the node-indices used in mesh edge index definitions are defined so that for each nodeset (for example "mesh nodes level 0") the indices run from `0` to `N-1` where `N` is the number of nodes in that nodeset**? If this is how it works, then it also clear how the indices of the node features are defined.

Marianadesamadsen:
Thank you for the time on writing the documentation on the graph storage @leifdenby! That is very usefull:)

I was thinking that for future updates, it would be nice to rename some of the saved file such the names are aligned across all files. E.g. `m2m_features.pt` could become: `m2m_edge_features.pt`, and `mesh_nodes_features.pt` instead of `mesh_features.pt`, hope it makes sense.

docs/validate_graph.py
leifdenby: Did you open this issue @Joltsy10? :) If so we should "resolve" this conversation

docs/graph_storage_spec.md
leifdenby: Ok, I will add this to the spec document. We can then revise in future whether this should be a requirement or not

docs/graph_storage_spec.md
leifdenby: ok, I will add this to list of things to address in future when revising the requirements from the `neural-lam` side

docs/validate_graph.py
Joltsy10: Yes, opened it here: #348

docs/graph_storage_spec.md
joeloskarsson: Yes, that is accurate and a good description :D

docs/graph_storage_spec.md
joeloskarsson: I hardly understand what I meant here either (brain cooked from looking at too many PRs in a row 😆). I think my thought was that we should be more specific than "floating-point". You updated some things to specifically the torch dtype (e.g. `torch.float32`), and it would probably be good to do the same here.

Shivampal157:
@joeloskarsson @sadamov ,I had explored a small validate-graph CLI in #408 / #427; happy to align with this PR instead. I can help with: (a) tests for the validator, (b) user-facing error messages, (c) a thin `python -m ...` entry if it matches the spec  whatever maintainers prefer. Point me to the right sub-task on this branch.

joeloskarsson:
Thanks @Shivampal157 ! My understanding from the last dev meeting is that this PR is largely done, so we should avoid adding new things at this point so that it can be merged quickly. But please feel free to review the code changes and comment if anything stands out.
Once this is merged we can always build on top of it with more functionality and improvements.

joeloskarsson:
> normalization of features: mesh_static_features [are currently normalised before saving to disk](https://github.com/mllam/neural-lam/blob/main/neural_lam/create_graph.py#L411), but mesh and grid node features [are normalized during graph-loading](https://github.com/mllam/neural-lam/blob/main/neural_lam/utils.py#L155) - what do you think about either normalizing all or none of the graph features? I think it probably makes more sense to not normalize any since we also normalize data at training time, so this approach would mirror how we treat the training data in general

Agree, better to keep features on disk non-normalized to be consistent with other data.

> assumption about mesh features for normalisation: We [currently assume that the 0th feature for m2m edges represent the length](https://github.com/mllam/neural-lam/blob/main/neural_lam/utils.py#L117) and use this value to scale all edge features. Would it not make more sense to scale each mesh-feature separately and assume that every each has the same set of features (with different values of course)?

Since mesh edge features are lengths and coordinates, rescaling with max length becomes just a rescaling of the coordinate system, so the length feature is still the correct length of the vector diff features. I note that this does actually mean that we make another (very implicit) assumption here, that all edge features are lengths or coordinates in the same CRS. This methodology comes from graphcast, and I think it makes sense with this assumption. However, if we think of edge features more broadly (which we probably want to), then scaling them independently as you describe makes more sense.

> assumption about static mesh features: the 3D graph plotting currently assumes that the mesh node features represent x and y (spatial coordinates), should we break/remove the 3D graph plotting so that we can relax this requirement? Or should we require what the mesh node features are for now?

Yes this is a bit of a problem really, that we don't have access to the mesh node positions without this assumption. To keep with the spirit of this PR (describing current situation), we should assume that the mesh node features are positions. In the future, we should move towards saving the mesh node positions separately (so they are disconnected from the actual node features used by the model).

Prince637-boo:
Hi @leifdenby @joeloskarsson  @Joltsy10 @Marianadesamadsen  @sadamov , I’ve been following this discussion on the graph-on-disk spec with great interest.

I'm currently working on the Global Graph Integration (focusing on the icosahedral meshes from WMG). I’ve already implemented a 3D Plotly visualizer and I’m ready to test the integration.

To align with your new spec:

  1.Should I start generating the boundary_mask as all-False tensors for my global tests?

  2.or the mesh_static_features, I'm moving to 3D Cartesian [N, 3]. Does the current validator you're building already support this shape, or should I help update it?

I’d love to ensure my WMG-to-Neural-LAM pipeline follows the contract you're defining here.

Joltsy10:
Hey @Prince637-boo this pr is largely for the lam side of neural-lam and the discussion for global is currently happening in #348 . The current spec validates according to lam/ rectilinear graph specs.

1. There is more discussion required for this but from what I know for the global validator it would make more sense to not have a boundary mask at all, passing an all false tensor would kind of make it a special case of lam rather than its own thing. Open to see what other options are available. 
2. The current global graph returns a [N, 2] lat/lon features which already matches the lam  spec, this then is planned to undergo sin/cos embedding after passing the spec.

Good points though I would suggest you to bring them up in #348 thanks!

docs/graph_storage_spec.md
leifdenby: Reformulated in https://github.com/leifdenby/neural-lam/commit/60c9906c8144075553249b503f58b8ca3615ee46

docs/graph_storage_spec.md
leifdenby: Change text in https://github.com/leifdenby/neural-lam/commit/4c33d1b5df8fb2d77279109341cc3452d869dbc9 to reflect that edge and node attributes should not be normalized in what is saved to disk, instead any normalization will take place inside of `neural-lam` after loading

docs/graph_storage_spec.md
leifdenby: Great :) yes, I made explicit in https://github.com/leifdenby/neural-lam/commit/3278bde19857282e939fb26083af0612afe38b8e what kind of `torch` dtype everything should be

docs/graph_storage_spec.md
leifdenby: In https://github.com/leifdenby/neural-lam/commit/213ad19115b604c3f927c30c48da4227f092d8dd I have written that first two mesh node features must be xy coordinates. To be able to carry out normalization within `neural-lam` (instead of during graph creation) after the graph is loaded from disk, we need to know which of feature indices define the spatial coordinate of a mesh node. And since we don't communicate in any other way which feature is what, we need to enforce at least which are the spatial coordinate features of the mesh nodes.

leifdenby:
Ok, I've gone through and addressed your comments @joeloskarsson, but I think we have slight problem here in that the current `neural_lam.utils.load_graph(...)` implementation expects that mesh node features (the node coordinates) are already normalized before saving to disk, but here we are saying that going forward these features should be save non-normalized to disk. If we merge in this spec as it is currently written, and people follow it, they will create graphs that, as the `neural-lam` `load_graph` implementation currently stands, will not have the mesh node features normalized. I guess there is no harm in normalizing twice, in that we could introduce the mesh-node feature normalization in this PR, and for the existing graphs that people have these features will actually be unchanged (because they are already normalized), if implement carrying out the exact same normalization to happy after graph load, as we [currently have before writing the graph to disk](https://github.com/mllam/neural-lam/blob/main/neural_lam/create_graph.py#L413).

Wrt to assuming what the mesh node features are, and what the edge features (across all subgraphs are), I think we have to write the spec to restrict that edge features must only be edge distances (x-dist, y-dist and Euclidean distance). That is [because currently scale **all** edge features](https://github.com/mllam/neural-lam/blob/main/neural_lam/utils.py#L265) by "the longest distances". If you agree with this I will change the spec text to say that exactly 3 edge features are required, and they must capture edge lengths (since they are all normalized together).

docs/graph_storage_spec.md
joeloskarsson: Yes, that's good I think. In the future we should probably have the graph coordinates communicated wmg -> neural-lam separately, not specifically part of node features. But that is not how things are now.

joeloskarsson:
Indeed this is a problem, and I pretty much agree with everything you write @leifdenby. However if we are being faithful to this spec really describing the current situation (and not changing it, even if it is bad), then we should just have "mesh node features should be normalized before being saved to disk" in here. And then have a follow-up PR that gets rid of this, and changes it in the spec. 
With that said, I am not against changing this as part of this PR. If you choose to do so, then you can either 1. do the double normalization, as you describe, or 2. remove the normalization from ´create_graph.py´, so it does not need to happen twice.

W.r.t. assumptions of what the edge features are: Agreed, at the moment this is the assumption so that is what we should state. I would specifically put that the first feature is the edge length (this is what we look at for normalization), and the following features are cartesian coordinates. This encompasses both 2d (x,y) and 3d (x,y,z) edges, which requires less specification changes in the future.

leifdenby:
> However if we are being faithful to this spec really describing the current situation (and not changing it, even if it is bad), then we should just have "mesh node features should be normalized before being saved to disk" in here. And then have a follow-up PR that gets rid of this, and changes it in the spec.

True, but in that case we will in implementing this follow PR still have the issue of not being able to support graphs people made with previous revisions of `neural-lam` (with its own `create_graph` CLI). My gut is to go with the double normalization, which although it will be redundant ensures that we can keep supporting the graphs that people already have stored on disk.

> With that said, I am not against changing this as part of this PR. If you choose to do so, then you can either 1. do the double normalization, as you describe, or 2. remove the normalization from ´create_graph.py´, so it does not need to happen twice.

Great! Ok, I will go for option 1. with the double normalization.

> W.r.t. assumptions of what the edge features are: Agreed, at the moment this is the assumption so that is what we should state. I would specifically put that the first feature is the edge length (this is what we look at for normalization), and the following features are cartesian coordinates. This encompasses both 2d (x,y) and 3d (x,y,z) edges, which requires less specification changes in the future.

Good point, I will update the text specify that the first feature must be the edge length.

leifdenby:
Ok, I think I am done with this now. In line with our discussions above @joeloskarsson, I have made the following changes to the spec text:

- for mesh node features the first two features must be the spatial (horizontal, e.g. x/y) coordinates, and details that these are normalized together. 
- for the edge features the spec now states that the first must be the edge length (this is used for normalization) and otherwise there must be either two (x, y edge length) or three (x, y, z edge length) features. I made restriction on what the fourth features is allowed to be, we cannot know whether we should normalize also that feature with the edge length.

The final change I have made in the last few commits is to integrate the spec text directly into the validate script and setup a pre-commit hook to generate the spec text markdown file by running the validator script (this script now has a `--write-spec-to-path` flag that saves the spec text to markdown). The reason for doing this is to try to ensure that the spec text and the validation tool (that implements the spec) don't drift apart too quickly, by putting the spec text and the checking functions right next to each other.

With those changes I think this is ready to go in @joeloskarsson - hopefully you think so too! :)

docs/graph_storage_spec.md
```diff
@@ -0,0 +1,179 @@
...
+- Entry `0` MUST always be the bottom mesh level.
+
+Every tensors MUST stored in a manner ameanable to load with `torch.load(...)` (this can most easily be support by using `torch.save(...)` to store tensors to disk) and satisfy the requirements below.  # noqa: E501
+### 3.1 Nodes
+
+The `neural-lam` graph format on disk does not explicitly store node features for grid nodes, as these are expected to be dynamic and stored separately in the dataset. However, static features for mesh nodes MUST be stored in `mesh_features.pt` files (as described below).  # noqa: E501
+
+#### 3.1.1 Node index space
+
+The node indices in edge index tensors MUST be defined so that for each nodeset (for example "mesh nodes level 0") the indices run from `0` to `N-1`, where `N` is the number of nodes in that nodeset, i.e. the node indices for each nodeset MUST be contiguous. For example, if there are `N_0` mesh nodes at level `0`, then the node indices for those nodes MUST be `0` to `N_0 - 1`. If there are `N_1` mesh nodes at level `1`, then the node indices for those nodes MUST be `N_0` to `N_0 + N_1 - 1`, and so on.  # noqa: E501
```
Joltsy10: I think this section is describing two different indexing conventions? The first sentence says that each nodeset runs from '0' to 'N-1', but the example reads like a cumulative offset, ie the indexing for mesh 1 nodes are said to be indexed from 'N_0' to 'N_0 + N_1 - 1', which doesnt seem right.

neural_lam/utils.py
```diff
@@ -230,6 +230,26 @@ def loads_file(fn: str) -> Any:
         "mesh_features.pt"
     )  # List of (N_mesh[l], d_mesh_static)
 
+    # Normalize static mesh features
+    num_features = mesh_static_features[0].shape[1]
+
+    # Coordinates (first two columns) are normalized jointly
+    pos_max = max(torch.max(torch.abs(m[:, :2])) for m in mesh_static_features)
```
Joltsy10: If I'm reading this correctly, old graphs may no longer be rescaled exactly the same way as before, since the previous normalization used the full grid extent and now it uses the mesh-feature extent after load

Joltsy10: I think the two would match only when the mesh coordinates reach the same extrema as the full grid domain. If the mesh stays inside the grid bounds, the loaded graph could be scaled differently; if it gets close to those bounds, the result may only be slightly different rather than exactly the same

docs/graph_storage_spec.md
joeloskarsson: Hmm, yes I think there was maybe some misunderstanding in the discussion https://github.com/mllam/neural-lam/pull/323#discussion_r2879022074. I wrote

> However, they should be contiguous in the sense that at one single level, the indices must lay in the range [start_i, start_i + num_mesh_nodes_at_level], for an arbitrary start_i and with num_mesh_nodes_at_level being the number of nodes on that mesh level.

but this describes the hard constraint of start_i = N_i for a specific N_i (number of nodes in level below). This is not a requirement, start_i can indeed be any arbitrary integer (0 would be most reasonable). I would remove the last part here.
```suggestion
The node indices in edge index tensors MUST be defined so that for each nodeset (for example "mesh nodes level 0") the indices run from `0` to `N-1`, where `N` is the number of nodes in that nodeset, i.e. the node indices for each nodeset MUST be contiguous.  # noqa: E501
```

neural_lam/utils.py
joeloskarsson: Agreed, `pos_max` in `create_graph.py` used to come from the grid, not the mesh coordinates. This will in general lead to a different rescaling. Is this intentional?

neural_lam/utils.py
```diff
@@ -230,6 +230,26 @@ def loads_file(fn: str) -> Any:
         "mesh_features.pt"
     )  # List of (N_mesh[l], d_mesh_static)
 
+    # Normalize static mesh features
+    num_features = mesh_static_features[0].shape[1]
+
+    # Coordinates (first two columns) are normalized jointly
```
joeloskarsson: It is maybe a bit odd how in the edge feature description we opened up for the possibility of 3d coordinates, but here we explicitly assume that nodes have 2d coordinates. Could we not just assume that all node features are coordinates for now? I don't know of a case where they are not.

joeloskarsson:
I think it would be good to look at the questions raised by @Joltsy10 above, otherwise I am happy.

docs/graph_storage_spec.md
leifdenby: ok, yes, those two sentences were contradictory (one stating indices should always start from `0`, the second sentence saying it should start from `N_{i-1}` where `i` is the level number), and also, as you say, we don't need to restrict what the starting value is, because we actually offset the indices so the minimum value is `0`.

The problem here is also that the text doesn't carefully define what a "nodeset" is, in part because we have subsets of nodes depending on graph we are trying to construct (are just the "mesh"-nodes a nodeset? are "mesh nodes on level 1" a nodeset?) Maybe instead of using an ambiguous term like "nodeset" I should instead be explicit about the "set of source nodes" and "set of destination nodes", like so:

> The node indices in edge index tensors MUST be defined so that for both the set of source nodes (where the edges connect from) and the set of destination nodes (where the edges connect to) should be denoted by indices running from `0` to `N_src-1` and `0` to `N_dst-1` respectively, where `N_src` and `N_dst` are the total number of source and destination nodes being connected. This means that the node indices in the edge index tensors are numbered separately, and the nodes in each set are only numbered identically if the set of nodes being connect from and to are the same set (as would be the case for mesh-to-mesh connections in non-hierarchical graphs)

Given our discussion here I think it is also worth putting in a note:

> **NOTE**: The smallest index in any given set of nodes being connect (source or destination) may in fact start at any integer value as `neural-lam` internally offsets the indices to start at zero, the key constraint is that the indices are contiguous.

What do you think about this?

neural_lam/utils.py
leifdenby: ah, I hadn't realised that. That means I should get the grid coordinate values out of the datastore and scale by their range instead. Ok, I will fix that. Thanks for catching this @Joltsy10 !

neural_lam/utils.py
leifdenby: I can imagine a situation where one might expected to be able to include node features containing static attributes that were provided during graph generation. For example say you wanted to include height of topography on mesh nodes. Should those additional features be included inside `neural-lam` after loading the graph, but before the training loop starts? If that was the case then my graph doesn't hold all the static features that are fed to the graph.

But maybe what is more important here is that we guide the person creating the graph towards the intended use. Maybe we create the restriction here to say **only coordinate defining features** are allowed on nodes, and any other static features should be handled after loading the graph in `neural-lam`. That way we make it clear how to include other static features.

docs/graph_storage_spec.md
leifdenby: @Joltsy10 and @joeloskarsson any thoughts on my suggestion above for addressing your comments?

Joltsy10: This looks good to me overall, I agree that describing source and destination nodes is clearer than just an ambiguous nodeset. 
But one small nitpick would be to tweak the 0 TO N_src-1/ 0 to N_dst-1 wording, since it says they MUST be from 0 to N-1, but the note says that neural-lam offsets them anyways. Maybe it would read more cleanly if the main requirements is just that the src and dst indices are contiguous within their respective sets, and then followed by the note that neural-lam automatically offsets them to start at zero.

neural_lam/utils.py
leifdenby: Ok, this is harder than it might seem, for two reasons:

1. The grid coordinates of the input data don't actually become part of the graph static features at all, as-in these coordinate values are not set in the graph as grid-node features, and so are not written to disk as part of the saved graph. Instead these static features are [appended at run-time](https://github.com/mllam/neural-lam/blob/main/neural_lam/models/base_graph_model.py#L307) (from `BaseGraphModel.grid_static_features` which are set by getting these values directly from the datastore). This means what for the graph saved on disk, without any feature normalization as we are planning here, will not contain the grid coordinates to normalize with.
   - => this means the grid coordinates (or rather their spatial range) have to be provided to the `neural_lam.utils.graph_load(...)` call as an argument, and this information can only come from the datastore.
2. Because the mesh node coordinate features are not scaled by their own range, I don't see how to construct a no-op operation here. As in, the only way I can see to create an operation that both normalizes the mesh node coordinate features for the graphs with features stored un-normalized (as the current spec of this PR requires), and also leaves mesh node coordinate features unchanged existing stored graphs (where these features are already normalized) is if for the old graphs these values already are in the [0, 1] range, I compute their range and scale by this value (thereby scaling by 1). However, what we are saying now is that mesh node coordinate features should always be scaled by the grid coordinate range, and that number (since it has to come from the datastore, see above) is always going to be the same, whether the mesh node coordinate features have already been normalized or not.

This honestly is a bit of a mess... but it's quite tricky to handle a situation where part of the graph features have been normalized and some have not.

The only thing I can think of now is to either:

1. only normalize the mesh node coordinate features if it looks like these coordinate features have not be scaled. As in the numerical range is "quite large". This seems dangerous to me. 
2. or, it might be safer to add a warning if the mesh node coordinate features are small (say less than abs 10.0 i value) and create an environment variable (say `NEURAL_LAM_DONT_SCALE_MESH_NODE_FEATURES`) that can be explicitly set to keep the old behaviour where mesh node coordinate features are not normalized on load.
3. or lastly, we could add a sentinel value stored with the graph that indicates that version of `neural-lam` the graph was created with. This would have to be an additional file I think since we can't change the contents of the files without breaking backward compatibility. This file could for example be called `created-with-neural-lam-version` and have the version number inside. In the absence of this file we would raise warning and assume this graph is in the "old" format where mesh node features have been normalized already and skip normalization. That way if people accidentally forget to include this file they will have a clear indication that their mesh node features aren't being normalized.

I think I prefer the last approach because it also leaves us space to evolve the spec in future. Although supporting old revisions may also be a pain... But we can then decide to drop support when that becomes necessary.

Thoughts @joeloskarsson and @Joltsy10 ?

neural_lam/utils.py
leifdenby: I will implement the restriction that only coordinate values are allowed for mesh nodes.

neural_lam/utils.py
Joltsy10: Of the three options, I think the third one looks safest to me as well. The first feels too heuristic, and the second seems a bit fragile since loader behaviour would then depend on an env variable that may differ across developers or CI.

If we are not too opposed to adding an extra file to the graph directory, I wonder if a slightly stronger version of the third option would be to make that file explicit metadata for the graph format rather than a created-with-version sentinel.

The reason I think that may be better is that the metadata file could store the exact graph-format semantics the loader needs, rather than only inferring them indirectly from a version marker.

leifdenby: > exact graph-format semantics the loader needs

This sounds interesting, can you give a bit more detail about in what form this would take @Joltsy10 ? As in, what would be written and to where?

Joltsy10: I was thinking it would be written as one small sidecar metadata file in the graph directory itself, next to the existing .pt files, graph_metadata.json for example.

I think it should include the normalization state of the mesh coordinate features, at minimum it should know whether it is normalized or yet to be normalized, and ideally we could include the exact scale value to use if normalization is still required. The graph generation step already has this so it would just be creating one extra file and load_graph() would be reading one extra file and know how to handle mesh coordinate normalization without guessing or relying on an environment variable.

If we do go this route, I think the same metadata file could also be a good place for a small graph format version and other future format semantics, like coordinate or edge-feature conventions, so that it stays self describing beyond this normalization issue.

Joltsy10: Seeing the consensus from the meeting, if we go with a warning or hard error for older graphs rather than full backward compatibility, then I do not think an extra metadata file makes the most sense for this PR.

I could still see metadata being useful later if we need to distinguish graph semantics more explicitly, e.g. LAM vs global or different coordinate/feature conventions, but that feels better handled when that need is more concrete so we can get this pr out easier and cleaner.

@leifdenby

leifdenby: Ok, I have in https://github.com/leifdenby/neural-lam/commit/6dac0be9a9f77537fc94de08d22af76fbf9f85e5, https://github.com/leifdenby/neural-lam/commit/f6b18e4c766495ebed7a857db102a963a3bfff84 and https://github.com/leifdenby/neural-lam/commit/512acc648303ac04f1b2a05e3e4df316e5699bcb (full diff: https://github.com/leifdenby/neural-lam/compare/6dac0be9a9f77537fc94de08d22af76fbf9f85e5%5E...512acc648303ac04f1b2a05e3e4df316e5699bcb) made the following changes:

- `neural_lam.create_graph(...)` will going forward write a file called `created-with-neural-lam-version` which will contain the version of `neural-lam` that the graph was created with (using `neural_lam.__version__`)
- if this file is absent or the version given is less than or equal to `0.6.0` then this is treated as the "legacy" format and mesh node features are not normalised (they are assumed to be already normalised). We raise a warning to inform the user of this behaviour
- otherwise (i.e. the version given is larger than `0.6.0`) the spec version `0.1.0` applies and mesh node features are normalised at load time.

This means that `neural-lam=0.6.0` will be the last version to write to the "legacy" format where the mesh node features are stored normalised and where we didn't have a formalised specification. `neural-lam=0.7.0` will be the first release to use a specification, with the first version of the specification being `0.1.0`.

I have also made additions to the spec text to clarify the differences between the "legacy" and "0.1.0 spec" format. And finally, I have added test to ensure that the mesh node features are correct normalised or not based on whether the "legacy" or "0.1.0 spec" format is being used.

This approach should also leave us space to evolve the specification and validator, and also `neural-lam` itself, while making decoupling what expectations are about what is written for different revisions of the spec and leave us the ability to drop support graphs from old revisions of `neural-lam` in future.

What do you think @Joltsy10 ?

I will implement the actual correct normalisation by the grid coordinate range next (by passing this range to `neural_lam.utils.load_graph(...)` from the datastore

Joltsy10: This sounds good to me. I was worried it would become a little too neural-lam coupled but this handles it well by just using the versioning for older neural-lam graphs. We just have to make sure the users prenormalize if theyre using their own graphs. Or are external graphs (non-neural-lam graphs) not that much of a priority?

neural_lam/utils.py
leifdenby: Implemented in https://github.com/leifdenby/neural-lam/commit/42a32d3775663fec68b8b8f77747128ce41ac355

neural_lam/utils.py
leifdenby: ah good point, maybe it would better to say the version file should contain the version of the spec that the graph conforms to instead. That way "legacy" could be the format before the first spec was defined, and the definition isn't tied to the neural-lam version. And the absence if the version file could still indicate that the graph conforms to the "legacy" format before the format was formalised in a specification. And this way people could create graphs for neural-lam them selves and they just have to indicate with the now "spec version file" what revision of the spec their graph follows.

docs/graph_storage_spec.md
joeloskarsson: I think that sounds good.

neural_lam/utils.py
leifdenby: Ok, I've made those to changes now, specifically:
1. The version file is used to define to the specification a given graph conforms to, defaulting to `legacy` where mesh-node features are  assumed to already be normalised
2. The maximum span of the grid coordinates is passed in (from the datastore) to the `load_graph(...)` function now to normalised mesh node features for graphs conforming to `0.1.0` of the spec (which will be the first revision).

I've also just done a complete read-through of both the validator script and the rendered markdown.

Could you let me know if you think this is ready to go in now @joeloskarsson and @Joltsy10? Cheers

Joltsy10: This looks good to me now. The versioning split and the mesh-feature scaling change address the main concerns I had earlier, so I think this is good to go in🎉

docs/validate_graph.py
```diff
@@ -0,0 +1,2172 @@
...
+    ## 3. File content requirements
+
+    The content of the files depend on the number of mesh levels, denoted as
+    `L` in the text below, so that for:
+
+    - Non-hierarchical graphs `L == 1`.
+    - Hierarchical graphs `L > 1`.
+    - Entry `0` MUST always be the bottom mesh level.
+
+    Every tensor MUST stored in a manner ameanable to load with
```
prajwal-tech07: Here are the spec-text issues I mentioned:

1. `docs/validate_graph.py` line 1549 (renders as `graph_storage_spec.md` line 117): "Every tensor MUST stored in a manner ameanable to load with..." - two fixes: "MUST be stored", and "ameanable" → "amenable". Since the markdown is auto-generated, the fix goes in this line of the script.
2. Section numbering - the spec jumps from `### 3.2 Edges` straight to `### 3.4 Differences to legacy format graphs`, so there's no 3.3. Either renumber 3.4 → 3.3, or it's fine if a 3.3 was intentionally removed.

leifdenby: Good catch @prajwal-tech07 ! I am impressed you caught this. Going through these suggestions I also realised that the "differences to legacy format" section should really have its own section, which I've done now while addressing your suggestions in https://github.com/mllam/neural-lam/pull/323/commits/366889eed0a43dcd37e200b8f1deb8e177806c34.

Thanks again!

neural_lam/utils.py
```diff
@@ -322,30 +330,115 @@ def loads_file(fn: str) -> Any:
...
+        # m2g and g2m has to be handled specially as not all mesh nodes
+        # might be indexed.
+        m2g_min_indices = m2g_edge_index.min(dim=1, keepdim=True)[0]
+        mesh_first = m2g_min_indices[0] < m2g_min_indices[1]
+        g2m_edge_index = zero_index_g2m(
+            g2m_edge_index, mesh_static_features, mesh_first=mesh_first
+        )
+        m2g_edge_index = zero_index_m2g(
+            m2g_edge_index, mesh_static_features, mesh_first=mesh_first
+        )
```
joeloskarsson, line 441: With all of this checking between the legacy and current graph spec, I guess we don't actually need the special treatment of g2m/m2g anymore. Since there should be no way to build a legacy graph where not all mesh nodes are connected. This was added in preparation for wmg integration, but I guess graphs coming from wmg will anyway already be in current spec and `should_zero_index_edge_indices == False` now.

It doesn't hurt in any way to keep this, just a comment. One could argue that someone could write a separate script that actually conforms to the legacy graph spec and would require this.

joeloskarsson (approved):
I think this all looks good. Added a comment above, but that is more just a note than something needing changing. I read through the whole spec, which I think is where my input is mainly needed. Find nothing needing change there.
I have not read all the code for the validator, but trust that others can verify that this makes sense.

prajwal-tech07 (commented):
Did a full pass over the spec, validator, and loader for this final review - LGTM, I think this is ready to merge. 👍

I have a particularly close view of this from the other side: my `weather-model-graphs` serialization (`to_torch_tensors_on_disk`, mllam/weather-model-graphs#123) writes graphs against exactly this spec, and they validate clean for keisler / graphcast / hierarchical, so the contract holds in practice. A few things I verified:

Spec ↔ validator are in sync - regenerating `graph_storage_spec.md` from `validate_graph.py` reproduces the committed doc exactly, and the new pre-commit hook enforces that going forward. 👍
Loader dual-format path (`utils.py`) matches the spec: missing `metainfo.yaml` → legacy (skip mesh normalization, zero-offset indices); `spec_version: 0.1.0` → normalize mesh features, indices already per-node-set zero-based.
Two tiny non-blocking nits and one question, none of which need hold up the merge:

Nits (grammar, in auto-generated text so the fix is in the validator docstrings):

§2.1: "Each graph MUST identified by a unique `name`" → "MUST be identified".
§3: "Every tensor MUST stored in a manner amenable to load" → "MUST be stored".
Question (clarification for future 3D/global work):

Edge features allow `N_f ∈ {3, 4}` (3D edges with a z-displacement), but `mesh_features` is fixed to exactly 2 columns (`MESH_FEATURE_DIM = 2`). For a genuinely 3D/spherical graph, mesh nodes would have a third coordinate - is 3D edge support intentionally forward-looking for 0.1.0, or should `mesh_features` permit 3 columns in the 3D case? Happy for this to be a follow-up; just flagging so it's a conscious choice before tagging the version.
Thanks for the very thorough write-up - the RFC 2119 phrasing and the index-space notes (§3.1.1) make the contract unambiguous. 🚀

Joltsy10 (commented):
Overall aside from a few typos the code and logic looks solid. Thanks for the amazing work @leifdenby 😊

neural_lam/utils.py
leifdenby: Yes, true. Maybe I should add a comment here. As you said, the most complete way of defining the legacy format would be to write a validator for that. Maybe someone will one day :)

leifdenby:
> §2.1: "Each graph MUST identified by a unique `name`" → "MUST be identified".
> §3: "Every tensor MUST stored in a manner amenable to load" → "MUST be stored".

Thanks @prajwal-tech07! Well done (again) for catching typos like this ⭐ 

> Edge features allow `N_f ∈ {3, 4}` (3D edges with a z-displacement), but `mesh_features` is fixed to exactly 2 columns (`MESH_FEATURE_DIM = 2`). For a genuinely 3D/spherical graph, mesh nodes would have a third coordinate - is 3D edge support intentionally forward-looking for 0.1.0, or should `mesh_features` permit 3 columns in the 3D case?

@prajwal-tech07 I think you have brought up a weak point in the current spec, which is clarity about what physical edge and node features we allow for, and how neural-lam treats these. I for example am not sure about the extent to which even suggesting that z-displacement is supposed even makes sense. By this I mean that having a z-coordinate is to me only relevant when considering the 3D position of points on a sphere in Cartesian coordinates, but I do not think that models will learn very well if they are given unit sphere coordinates for the mesh nodes. 

What I think the intention is with the specification we are adding in this PR is simple to say: you can add a 4th edge feature and if you do it will be normalised like the 2nd and 3rd edge feature, which is to say by the 1st edge feature. And this is based on the idea is that the 1st edge features contains the Euclidean length of the edge.

So in summary: yes, I think it is inconsistent that we don't allow for a third mesh node feature when we allow for up to three Cartesian displacement components, but I think to make a proper general implementation here we actually need to discuss further how to represent what the edge-features and node feature physically are, how/whether to communicate this to `neural-lam`, and what `neural-lam` should do with this info. And, some of this will require collecting information from people about what has worked/not worked for them. So I would see this as an evolution, i.e. the next version of the spec :) I know we will hit this though because `weather-model-graphs` will support creating graphs with the mesh coordinates (and thus features) are geographic (as in lat/lon), so we will need to discuss how to adapt the spec and neural-lam itself to work with these coordinates. But for now I would suggest we leave the spec as-is in this PR.

prajwal-tech07:
Yes Leif, agreed, we can leave the spec as-is for this PR. Deferring the node/edge feature representation (the third mesh-node coordinate, and the geographic lat/lon vs Cartesian question) to the next spec version sounds right. Like you said it needs input on what's actually worked for people, not just a dimensional fix. And yeah, once `weather-model-graphs` starts creating graphs with geographic mesh coordinates we'll definitely run into this, so happy to help work through it when we get to the next version.

Thanks for fixing the wording , Nothing blocking from my side, approval stands, good to merge.

leifdenby:
Ok, that is green lights from all round from @joeloskarsson + @Joltsy10 + @prajwal-tech07. Thanks all! This has been a mammoth task and I will be relieved to see this merged in :)

PR #325, opened 2026-03-03
Support area-agnostic graph loading via --graph_path (closes #299)

### LINKED ISSUE #299: Enhancement: Support area-agnostic graph generation and region configuration
### Problem

Graph loading is currently tightly coupled to the datastore directory structure:

The `--graph` argument only accepts a graph name, which is resolved relative to the datastore.

This introduces unnecessary constraints:
- Graphs must be manually copied into the datastore even if already available elsewhere
- Reusing precomputed graphs across experiments becomes cumbersome
- External graph generation workflows cannot be used directly

The limitation is not in graph definition (which should remain tied to the datastore), but in how graph files are located and loaded.

---

### Proposed Solution

Allow `--graph` to accept either:
1. A graph name (existing behavior), or  
2. A filesystem path to a graph directory  

#### Resolution logic:

```python
from pathlib import Path

graph_arg = Path(args.graph)

if graph_arg.exists():
    graph_dir_path = graph_arg
else:
    graph_dir_path = datastore.root_path / "graph" / args.graph


```
---

Additionally, move graph validation (e.g., node consistency checks) into utils.load_graph() so that:

validation is centralized
model code remains clean
behavior is consistent regardless of graph source

This approach preserves backward compatibility while removing unnecessary storage constraints.

### Brutal check

This version:
- doesn’t fight the datastore design  
- doesn’t introduce bad CLI (`--graph_path`)  
- clearly separates **problem (storage coupling)** vs **non-problem (config ownership)**

Saptami191:
## Describe your changes

This PR updates graph loading to support both datastore-based and filesystem-based graph resolution using the existing `--graph` argument.

Previously, graphs were always loaded from:
`<datastore_root>/graph/<graph_name>/`

This made it inconvenient to reuse graphs generated outside the datastore structure or from independent workflows.

This change:

- Updates `--graph` to accept either:
  - a graph name (existing behavior), or  
  - a filesystem path to a graph directory
- Adds `utils.resolve_graph_path` to centralize graph path resolution
- Refactors `utils.load_graph` to handle validation internally using the datastore
- Removes redundant path resolution and validation logic from model code
- Preserves full backward compatibility (default behavior unchanged)

### Motivation and Context

Graph loading was tightly coupled to the datastore structure. This limited flexibility when:

- Reusing precomputed graphs across experiments  
- Using graphs generated outside the datastore layout  
- Running workflows that do not follow datastore directory conventions  

This change improves modularity by allowing graph loading to be independent of storage layout while keeping datastore configuration as the source of truth.

### Dependencies

No new external dependencies were introduced.

---

## Issue Link

Closes #299

---

## Type of change

---

- [x] For any new/modified functions/classes I have added docstrings that clearly describe their purpose, expected inputs and returned values
- [ ] I have updated the README to cover introduced code changes (can be added if preferred)
- [ ] I have added tests that prove my feature works (can be added if required)
- [ ] I have requested a reviewer and assignee

---

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

---

- [ ] I have added a line to the CHANGELOG describing this change under:
  - *added*

---

## Checklist for assignee

- [ ] the tests pass
- [ ] PR is assigned to the next milestone (if applicable)
- [ ] author has added an entry to the changelog


neural_lam/region.py
```diff
@@ -0,0 +1,32 @@
...
+    projection: str = "latlon"
+
+    @classmethod
+    def from_yaml(cls, path: str):
+        with open(path, "r") as f:
+            data = yaml.safe_load(f)
+        return cls(**data)
+
+
+def generate_xy_from_region(region: RegionConfig):
```
kshirajahere: I think this may not match the current create_graph() coordinate contract.

create_graph() expects xy shaped (Nx, Ny, 2), and mk_2d_graph() reads x from xy[:, :, 0][:, 0] and y from xy[:, :, 1][0, :]. But np.meshgrid(lons, lats) with default indexing produces (Ny, Nx) arrays, so the stacked result becomes (Ny, Nx, 2) instead.

This seems worth validating with a small regression test, because it could produce degenerate geometry for the region-based path.

neural_lam/region.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@dataclass
+class RegionConfig:
+    min_lat: float
+    max_lat: float
+    min_lon: float
+    max_lon: float
+    resolution: float
+    projection: str = "latlon"
```
kshirajahere: Worth checking(i might be wrong): `projection` is added to the config, but I don’t see it being used yet in the actual coordinate generation.

neural_lam/train_model.py
```diff
@@ -17,6 +17,9 @@
 from .models import GraphLAM, HiLAM, HiLAMParallel
 from .weather_dataset import WeatherDataModule
 
+# After ALL imports
+matplotlib.use("Agg")
```
kshirajahere: Minor scope question: is forcing the matplotlib backend needed for the `--graph_path` feature itself?

i am asking mostly cuz this changes global behavior for the training entrypoint, so it may be easier to review separately from the graph-loading changes

kshirajahere (commented):
Nice PR(was going through the code thoroughly as recommended by @Saptami191 (i hope nobody minds the code review :D)) and the `--graph_path` direction makes sense and the fallback behavior looks straightforward.

I think  ideally adding a small regression test around the region path / external graph loading path, would make this much easier to validate :)

neural_lam/region.py
Saptami191: @kshirajahere Thank u for pointing out! the projection field is currently unused. It was added mainly to keep the region configuration extensible in case projected coordinate systems (e.g. Lambert / UTM) are supported later. That's why.

Saptami191:
> Nice PR(was going through the code thoroughly as recommended by @Saptami191 (i hope nobody minds the code review :D)) and the `--graph_path` direction makes sense and the fallback behavior looks straightforward.
> 
> I think ideally adding a small regression test around the region path / external graph loading path, would make this much easier to validate :)

Good suggestion! I’ve also been working on regression testing for GraphLAM outputs in PR #273. I can look into adding a small test covering the --graph_path / external graph loading path as well.

neural_lam/region.py
kshirajahere: in that case, it may be worth either documenting it as a forward-compatibility field in the config/PR description, or leaving it out until projection-aware handling is actually added, just so users do not assume it is already active. thanks :)

Saptami191:
@kshirajahere  Your review makes sense.your thoughtful review. Trying to do better. 
I’ll consider adding one while we wait for feedback from the maintainers.

neural_lam/region.py
Saptami191: That makes sense , thanks for the suggestion. Will check.

kshirajahere:
> @kshirajahere Your review makes sense.your thoughtful review. Trying to do better. I’ll consider adding one while we wait for feedback from the maintainers.

No worries 😃

neural_lam/train_model.py
joeloskarsson: Agreed, this is unrelated and not something that should go in with this PR.

neural_lam/models/base_graph_model.py
```diff
@@ -6,80 +6,98 @@
...
+            loaded_total_nodes = graph_ldict["g2m_edge_index"].max().item() + 1
+            loaded_grid_nodes = loaded_total_nodes - loaded_mesh_nodes
+
+            if loaded_grid_nodes != expected_grid_nodes:
+                raise ValueError(
+                    "Incompatible graph: expected "
+                    f"{expected_grid_nodes} grid nodes but "
+                    f"graph contains {loaded_grid_nodes}."
+                )
+
```
joeloskarsson: It would make sense to do as much validation as possible in utils.load_graph rather than here

neural_lam/region.py
joeloskarsson, line 1: Could you explain the meaning of this file? The region is specified by the datastore, and in the PR description I find no mention of introducing some new type of config.

neural_lam/create_graph.py
```diff
@@ -591,22 +597,57 @@ def cli(input_args=None):
...
         help="Generate hierarchical mesh graph. Otherwise multi-scale.",
     )
+    # new add
+    parser.add_argument(
+        "--output_path",
+        type=str,
+        default=None,
+        help="Optional output directory for graph. "
+        "Overrides datastore graph path.",
+    )
```
joeloskarsson: I would really like to avoid having two arguments: graph path and graph name. We need to figure out a better way to do this.

joeloskarsson (changes requested):
Hi, thanks for your contribution. I think that it could make sense to be able to store graphs separately from the datastore, but we need to think a bit about how to actually give the options for this and what the default behavior should be. Could you outline what your idea is regarding this?

There are many changes in the code that I think are unrelated to the graph path change. Could you go over and make sure that each change is directly related to what this PR is about? It is a lot trickier to review otherwise.

Trying to follow also the discussion in the issue, there seems to be an idea to be able to make graphs without referencing a datastore, but by some other type of config? (Pls explain this more) While I would need to understand better, this is something that I do not really agree with. The point of the datastore is to specify the data setting that you are building a model for. This includes the region you are going to forecast. Why would one build a graph if there is no corresponding datastore providing data to train a model with?

neural_lam/train_model.py
Saptami191: The matplotlib.use("Agg") line was added earlier to avoid display issues when running training in headless environments, but it is not related to the --graph_path functionality.

neural_lam/create_graph.py
Saptami191: my initial idea was to keep --graph for the existing datastore-based behavior and add --graph_path only as an override when graphs are stored externally. However, I agree that having two arguments representing the same concept is not ideal.one option could be to allow --graph to accept either a datastore graph name or a filesystem path, depending on whether the value resolves to a directory. I’m happy to adjust the interface if you have a preferred approach.. my approach was that

neural_lam/region.py
Saptami191: region.py was introduced while experimenting with generating grid coordinates from a simple region configuration (e.g like  bounds and resolution) instead of deriving them directly from the datastore. The idea was to explore workflows where graphs could be generated independently from the datastore layout. I can remove the region.py addition and keep the changes limited to supporting external graph loading.

neural_lam/models/base_graph_model.py
Saptami191: Good point. Moving validation into utils.load_graph() makes sense since the loader should be responsible for verifying the graph structure. I’ll refactor this so the model only handles graph usage.

neural_lam/train_model.py
kshirajahere: > The matplotlib.use("Agg") line was added earlier to avoid display issues when running training in headless environments, but it is not related to the --graph_path functionality.

In that case, I wud agree with joel to keep it out of this PR.

Saptami191:
> Hi, thanks for your contribution. I think that it could make sense to be able to store graphs separately from the datastore, but we need to think a bit about how to actually give the options for this and what the default behavior should be. Could you outline what your idea is regarding this?
> 
> There are many changes in the code that I think are unrelated to the graph path change. Could you go over and make sure that each change is directly related to what this PR is about? It is a lot trickier to review otherwise.
> 
> Trying to follow also the discussion in the issue, there seems to be an idea to be able to make graphs without referencing a datastore, but by some other type of config? (Pls explain this more) While I would need to understand better, this is something that I do not really agree with. The point of the datastore is to specify the data setting that you are building a model for. This includes the region you are going to forecast. Why would one build a graph if there is no corresponding datastore providing data to train a model with?

the purpose of this PR is to make graph loading more flexible by allowing Neural-LAM to load graphs from an explicit filesystem path instead of requiring them to be stored inside the datastore directory structure. Currently, graphs are always loaded from "datastore.root_path/graph/<graph_name>", which tightly couples graph files to the datastore layout. This PR introduces an optional "--graph_path" argument that allows users to specify a graph directory directly, while keeping the existing datastore-based behavior as the default when no path is provided. This enables workflows where graphs are generated externally, reused across experiments, or stored independently from the datastore without changing the current default behavior. It is a long term  experiment , i  was working on, that impact is  it  improves the flexibility of the training workflow without affecting existing behavior. Users can now load graphs stored outside the datastore directory, which makes it easier to reuse graphs across experiments or integrate graphs generated by external pipelines. Since the datastore-based loading remains the default when --graph_path is not provided, the change is fully backward compatible and does not impact current users or existing training setups.I encountered issues when using graphs generated outside the datastore structure, since the current implementation assumes graphs must exist under datastore.root_path/graph/. This PR allows specifying an explicit graph path so externally generated graphs can be used without restructuring the datastore.
BEFORE
                
datastore ──► graph/<name>
                 │
                 ▼
             model loads graph


                AFTER
                -----
            ┌──────────────┐
            │graph_path arg│
            └──────┬───────┘
                   │
        ┌──────────┴───────────┐
        │                      │
   explicit path          datastore path
        │                      │
        ▼                      ▼
    load graph             load graph    here is visualization of my goal.

Saptami191:
Steps used for testing:
Generated a graph:
python -m neural_lam.create_graph \
  --config_path tests/datastore_examples/mdp/danra_100m_winds/config.yaml \
  --name test_graph

Ran training using a custom graph path:

  --config_path tests/datastore_examples/mdp/danra_100m_winds/config.yaml \
  --graph_path tests/datastore_examples/mdp/danra_100m_winds/graph/test_graph

**Result:**

Graph was successfully loaded from the custom path.
Model initialization proceeded without errors.
Default graph loading (datastore.root_path / graph / args.graph) was correctly bypassed when --graph_path was provided.

neural_lam/train_model.py
```diff
@@ -87,8 +86,8 @@ def main(input_args=None):
     parser.add_argument(
         "--precision",
         type=str,
-        default=32,
-        help="Numerical precision to use for model (32/16/bf16)",
+        default="32",
+        choices=["32", "16", "bf16"],
```
joeloskarsson: This change is unrelated

neural_lam/models/base_graph_model.py
joeloskarsson, line 1: Please look over this file, there are a bunch of unrelated changes here.

neural_lam/models/base_graph_model.py
joeloskarsson: Did you resolve this? I don't see any changes in utils.load_graph

neural_lam/models/base_graph_model.py
Saptami191: Thank you! I’m currently working through the changes to clean up the PR and address the comments. I’ll push an updated version shortly.

Saptami191:
@joeloskarsson This PR improves graph loading by removing the dependency on the datastore directory structure.

The --graph argument now supports both graph names (existing behavior) and direct filesystem paths. This makes it possible to use graphs generated outside the datastore, reuse precomputed graphs across experiments, and run workflows that don’t follow the datastore layout.

I also centralized graph path resolution in utils.resolve_graph_path and moved validation into utils.load_graph, which simplifies the model code and keeps consistency checks in one place.

The change is fully backward compatible, so existing usage remains unchanged.

sadamov:
Hi @Saptami191, thanks for the patience on this one. A few things to address if you still have time to drive it forward:

1. Could you rebase onto current main? `base_graph_model.py` moved to `models/step_predictors/graph/base.py` via #208.
2. I'd remove the new `rank_zero_print` in utils.py - we replaced it with `loguru` in v0.6.0, so adding it back is a small regression.
3. @joeloskarsson's design question about the intended API surface for external vs datastore graphs is still open - it would help to outline your thinking there before we dig into the code review.
4. The regression test @kshirajahere suggested would also be great to have.

Worth coordinating with @leifdenby's #323 (graph-on-disk format spec) too, since the two overlap. Let me know if you have the bandwidth, otherwise no worries and we can pick this up later.

Issue #200, opened 2026-02-20
Support Iterating Over All Ensemble Members in WeatherDataset


prajwal-tech07:
**Problem**

When WeatherDataset is initialized with ensemble data (datastore.is_ensemble = True), only the first ensemble member (index 0) is used. All other ensemble members are silently discarded with a warning. This effectively reduces the usable dataset size by a factor of n_ensemble_members and makes it impossible to train ensemble-aware or probabilistic models.

The relevant code in _build_item_dataarrays currently hardcodes:

     i_ensemble = 0
     da_state = self.da_state.isel(ensemble_member=i_ensemble)

Additionally, ensemble forcing data raises a NotImplementedError when an ensemble_member dimension is present:

    if "ensemble_member" in self.da_forcing.dims:
           raise NotImplementedError(
                  "Ensemble member not yet supported for forcing"
             )
However, an existing XXX comment in the code already outlines the intended fix: split the sample index into (time_idx, ensemble_idx) and increase __len__ accordingly.

**Impact**

- Training only on ensemble member 0 wastes all other ensemble members.
- Reduces effective dataset size by a factor of n_ensemble_members.
- Blocks development of ensemble-aware or probabilistic forecasting models.
- __len__ warns about reduced dataset size but cannot correct it without this change.

**Proposed Solution**

1. Update __len__
      When is_ensemble=True, multiply the current dataset length by n_ensemble_members so that each ensemble member is treated as an     
      independent sample.

2. Update _build_item_dataarrays
      Split the flat dataset index using time_idx, ensemble_idx = divmod(idx, n_ensemble_members) and use ensemble_idx to select the correct      
      ensemble member instead of always using index 0.

3. Handle Forcing Data
       If forcing data contains an ensemble_member dimension, select the same ensemble_idx as used for state data rather than raising a    
       NotImplementedError.

4. Backward Compatibility
      Ensure that when is_ensemble=False, the dataset behavior remains unchanged and continues to function exactly as before.

5. Tests
      Add a test that verifies the dataset length scales correctly with n_ensemble_members and confirms that all ensemble members are iterated and      
      accessed during dataset traversal.

I would like to work on this issue and implement the proposed changes, including updating __len__, modifying _build_item_dataarrays, handling ensemble forcing correctly, and adding the necessary tests.

Please feel free to assign this issue to me.!!

observingClouds:
[...]

PR #332, opened 2026-03-04
FEAT:  use all ensemble members by default in WeatherDataset (Addresses #200) 

kshirajahere:
## Describe your changes

This PR adds support for iterating all ensemble members as independent samples in `WeatherDataset`, and makes that the default behavior for ensemble datastores.

### Summary of changes
- Added `use_all_ensemble_members: bool = True` to `WeatherDataset`
  - `True` (default): flat index maps to `(sample_idx, ensemble_idx)` and dataset length scales by number of ensemble members
  - `False`: legacy behavior using only ensemble member `0`
- Updated forcing behavior:
  - if forcing has `ensemble_member` coordinate, select matching `ensemble_idx`
  - otherwise reuse shared forcing across members
- Updated forcing dimension validation to allow optional `ensemble_member` before `grid_index`
- Wired through:
  - `WeatherDataModule(..., use_all_ensemble_members=...)`
  - CLI opt-out flag `--use_only_one_ensemble_member` in `train_model.py`
- Added regression tests for:
  - length scaling (analysis + forecast)
  - index mapping behavior
  - forcing behavior with/without `ensemble_member`
  - CLI/datamodule flag propagation
- Added CHANGELOG entry under `unreleased -> Added`

### Motivation / context
Implements the issue #200 direction discussed by maintainers: ensemble members should be treated as independent samples, with single-member behavior preserved only as an explicit fallback.

### Dependencies
No new external dependencies.

## Issue Link

Refs #200

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [x] the tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -303,4 +303,3 @@ This release introduces Datastores to represent input data from different source
 
 First tagged release of `neural-lam`, matching Oskarsson et al 2023 publication
 (<https://arxiv.org/abs/2309.17370>)
-
--- a/neural_lam/datastore/base.py
+++ b/neural_lam/datastore/base.py
@@ -35,9 +35,11 @@ class BaseDatastore(abc.ABC):
     dimensions (rather than just `time`).
 
     # Ensemble vs deterministic data
-    If state data is ensemble-valued, then the `is_ensemble` attribute should
-    be set to True and returned state data from `get_dataarray` is expected to
-    have an `ensemble_member` dimension. If `has_ensemble_forcing` is True,
+    If the datastore is used to present an ensemble of state realisations, for
+    example for forecast ensembles, then the `is_ensemble` attribute should be
+    set to `True` and returned state data from `get_dataarray` is expected to
+    have an `ensemble_member` dimension. If each ensemble member has its own
+    forcing values, then `has_ensemble_forcing` should be set to `True`, and
     returned forcing data from `get_dataarray` is expected to have an
     `ensemble_member` dimension; otherwise forcing data is expected not to have
     one.
@@ -247,11 +249,11 @@ class BaseDatastore(abc.ABC):
         elapsed_forecast_duration)` dimensions if `is_forecast` is True, or
         `(time)` if `is_forecast` is False.
 
-        If state data is ensemble-valued, the returned state dataarray is
-        expected to have an additional `ensemble_member` dimension. If
-        `has_ensemble_forcing=True`, the returned forcing dataarray is expected
-        to have an additional `ensemble_member` dimension; otherwise it is
-        expected not to have one.
+        If we have multiple ensemble members of state data, the returned state
+        dataarray is expected to have an additional `ensemble_member`
+        dimension. If `has_ensemble_forcing=True`, the returned forcing
+        dataarray is expected to have an additional `ensemble_member`
+        dimension; otherwise it is expected not to have one.
 
         Parameters
         ----------
--- a/neural_lam/datastore/mdp.py
+++ b/neural_lam/datastore/mdp.py
@@ -256,8 +256,8 @@ class MDPDatastore(BaseRegularGridDatastore):
         elapsed_forecast_duration)` dimensions if `is_forecast` is True, or
         `(time)` if `is_forecast` is False.
 
-        If state data is ensemble-valued, the returned state dataarray will
-        have an additional `ensemble_member` dimension. If
+        If we have multiple ensemble members of state data, the returned state
+        dataarray will have an additional `ensemble_member` dimension. If
         `has_ensemble_forcing=True`, the returned forcing dataarray will also
         have an additional `ensemble_member` dimension.
 
--- a/neural_lam/weather_dataset.py
+++ b/neural_lam/weather_dataset.py
@@ -38,10 +38,10 @@ class WeatherDataset(torch.utils.data.Dataset):
         t, given num_past_forcing_steps) are included as forcing inputs at time
         t. Default is 1.
     load_single_member : bool, optional
-        If `False` and the datastore returns ensemble-valued state data,
-        treat each state ensemble member as an independent sample. If `True`,
-        only ensemble member 0 is used. Default is False, so all members are
-        used when available.
+        If `False` and the datastore returns an ensemble of state
+        realisations, treat each state ensemble member as an independent
+        sample. If `True`, only ensemble member 0 is used. Default is False,
+        so all members are used when available.
     standardize : bool, optional
         Whether to standardize the data. Default is True.
     """
@@ -71,6 +71,10 @@ class WeatherDataset(torch.utils.data.Dataset):
         self.da_forcing = self.datastore.get_dataarray(
             category="forcing", split=self.split
         )
+        if self.da_state is None:
+            raise ValueError(
+                "The datastore must provide state data for the WeatherDataset."
+            )
 
         if self.datastore.is_ensemble and self.load_single_member:
             warnings.warn(
@@ -159,7 +163,7 @@ class WeatherDataset(torch.utils.data.Dataset):
         if self.datastore.is_forecast:
             # for now we simply create a single sample for each analysis time
             # and then take the first (2 + ar_steps) forecast times.
-            # If ensemble-valued state data is present and
+            # If the datastore returns an ensemble of state realisations and
             # `load_single_member=False`, each ensemble member is exposed as an
             # independent sample by scaling the base dataset length below.
 
--- a/tests/dummy_datastore.py
+++ b/tests/dummy_datastore.py
@@ -168,9 +168,7 @@ class DummyDatastore(BaseRegularGridDatastore):
         # Stack the spatial dimensions into grid_index
         self.ds = self.ds.stack(grid_index=self.spatial_coordinates)
         self.is_ensemble = "ensemble_member" in self.ds["state"].dims
-        self.has_ensemble_forcing = (
-            "ensemble_member" in self.ds["forcing"].dims
-        )
+        self.has_ensemble_forcing = "ensemble_member" in self.ds["forcing"].dims
 
         # Create temporary directory for storing derived files
         self._tempdir = tempfile.TemporaryDirectory()
@@ -339,10 +337,11 @@ class DummyDatastore(BaseRegularGridDatastore):
         elapsed_forecast_duration)` dimensions if `is_forecast` is True, or
         `(time)` if `is_forecast` is False.
 
-        If state data is ensemble-valued, the returned state dataarray is
-        expected to have an additional `ensemble_member` dimension. If
-        `has_ensemble_forcing=True`, the returned forcing dataarray is
-        expected to have an additional `ensemble_member` dimension.
+        If we have multiple ensemble members of state data, the returned state
+        dataarray is expected to have an additional `ensemble_member`
+        dimension. If `has_ensemble_forcing=True`, the returned forcing
+        dataarray is expected to have an additional `ensemble_member`
+        dimension.
 
         Parameters
         ----------
@@ -359,9 +358,8 @@ class DummyDatastore(BaseRegularGridDatastore):
             The xarray DataArray object with processed dataset.
 
         """
-        da_category = self.ds[category]
         dim_order = self.expected_dim_order(category=category)
-        da_category = da_category.transpose(*dim_order)
+        da_category = self.ds[category].transpose(*dim_order)
 
         if standardize:
             return self._standardize_datarray(da_category, category=category)
@@ -524,9 +522,7 @@ class EnsembleDummyDatastore(BaseDatastore):
                 forcing_has_ensemble,
             )
         self.is_ensemble = "ensemble_member" in self._da_state.dims
-        self.has_ensemble_forcing = (
-            "ensemble_member" in self._da_forcing.dims
-        )
+        self.has_ensemble_forcing = "ensemble_member" in self._da_forcing.dims
 
     # ---- data initialisation helpers ----------------------------------------
 
@@ -729,7 +725,7 @@ class EnsembleDummyDatastore(BaseDatastore):
         return ds
 
     def get_dataarray(
-        self, category: str, split: str, standardize: bool = False
+        self, category: str, split: Optional[str], standardize: bool = False
     ) -> Union[xr.DataArray, None]:
         if category == "state":
             da = self._da_state
--- a/tests/test_datastores.py
+++ b/tests/test_datastores.py
@@ -199,9 +199,8 @@ def test_get_dataarray(datastore_name):
 
             if datastore.is_ensemble and category == "state":
                 expected_dims.append("ensemble_member")
-            elif (
-                category == "forcing"
-                and getattr(datastore, "has_ensemble_forcing", False)
+            elif category == "forcing" and getattr(
+                datastore, "has_ensemble_forcing", False
             ):
                 expected_dims.append("ensemble_member")
 


sadamov:
please have a look at this already existing PR #200 and contribute there

kshirajahere:
@sadamov 
I may be missing it(pardon), but I can’t find an open PR #200 in this repo (I only see issue #200).  
Could you share the target PR link you want contributions on?

I already implemented and tested the core changes in #332 (now closed), and I’m happy to port those commits into the designated PR to avoid duplicate work.

sadamov:
@kshirajahere sorry, that was totally my bad! Got confused and reopened now. Please revert to the original PR template with all checkboxes.

kshirajahere:
hey no worries @sadamov all good! yea i will get back to original template thanks :D

kshirajahere:
@sadamov i have reverted to original template. Its ready for review @observingClouds @leifdenby

joeloskarsson:
I like the described behavior for dealing with forcing when there is ensemble data 😃

neural_lam/train_model.py
```diff
@@ -224,6 +224,14 @@ def main(input_args=None):
         default=1,
         help="Number of future time steps to use as input for forcing data",
     )
+    parser.add_argument(
+        "--use_all_ensemble_members",
```
joeloskarsson: In my opinion it would make more sense that this is the default mode (use all), and only using one member being na option. If we even need that option?

kshirajahere: > In my opinion it would make more sense that this is the default mode (use all), and only using one member being na option. If we even need that option?

I get it @joeloskarsson .
Working on it 🫡

kshirajahere: @joeloskarsson Updated the PR to make using all ensemble members the default behavior. As u suggest now WeatherDataset now use all ensemble members by default, also replaced the old positive CLI flag with an explicit opt-out: `--use_only_one_ensemble_member`, and lastly updated the tests and changelog.md

kshirajahere:
@joeloskarsson @leifdenby Pinging to ask for a rereview 😃

neural_lam/weather_dataset.py
```diff
@@ -123,13 +140,8 @@ def __len__(self):
...
             # This means that for each analysis time we get a single sample
 
-            if self.datastore.is_ensemble:
-                warnings.warn(
-                    "only using first ensemble member, so dataset size is "
-                    " effectively reduced by the number of ensemble members "
-                    f"({self.da_state.ensemble_member.size})",
-                    UserWarning,
-                )
+            if self.datastore.is_ensemble and not self.use_all_ensemble_members:
```
Sir-Sloth-The-Lazy: I think this changes the default behaviour of the WeatherDataset for ensemble datastores, previously only member 0 was used, now all members are used by default. Just making you aware about it since these lines (143-144) which early would warn everytime, now will do so if the user explicitly opts out! This should be a change worth mentioning in the CHANGELOG.md file. cheers :D !

neural_lam/weather_dataset.py
```diff
@@ -154,12 +166,26 @@ def __len__(self):
...
             #     time steps are required for the initial state)
             #   - future forcing: self.num_future_forcing_steps
-            return (
+            base_len = (
                 len(self.da_state.time)
                 - self.ar_steps
                 - max(2, self.num_past_forcing_steps)
                 - self.num_future_forcing_steps
             )
+        if self.datastore.is_ensemble and self.use_all_ensemble_members:
```
Sir-Sloth-The-Lazy: Any existing user with an ensemble datastore who upgrades will silently get a dataset that is N× larger (where N = number of ensemble members) with no warning or error.
To reproduce:
```
# Before this PR
ds = WeatherDataset(datastore=my_ensemble_datastore, split="train")
print(len(ds))  # e.g. 100

# After this PR, same code, no changes
ds = WeatherDataset(datastore=my_ensemble_datastore, split="train")
print(len(ds))  # now silently 300 (if 3 ensemble members)
```
The `_warn_using_single_ensemble_member` warning only fires when explicitly opting out (`use_all_ensemble_members=False`), but there is no warning for users who are silently opted in by the new default.
I think the user should be notified !

kshirajahere: I would respectfully disagree, since using all ensemble members is now the intended behavior and that warning would fire during normal usage for every ensemble datastore.

neural_lam/weather_dataset.py
kshirajahere: Thanks sir sloth. I agree that this should be called out explicitly in CHANGELOG.md. Will update soon.

kshirajahere:
Addressed @Sir-Sloth-The-Lazy Thanks :)

kshirajahere:
@joeloskarsson Please review :)

neural_lam/weather_dataset.py
joeloskarsson: Generally that is a good point, that this changes behavior drastically without giving the user any information. However, to give some context here, I am not aware of anyone actually using datastores with ensemble members at the moment. So I believe that we are quite safe to do this change, without it really impacting anyone's workflow. Of course we can not completely know this, but I would rather make this change now and enable better handling of ensemble data in the future.

neural_lam/train_model.py
```diff
@@ -224,6 +224,14 @@ def main(input_args=None):
         default=1,
         help="Number of future time steps to use as input for forcing data",
     )
+    parser.add_argument(
+        "--use_only_one_ensemble_member",
```
joeloskarsson: This is quite wordy, maybe
```suggestion
        "--load_single_member",
```
or similar?

neural_lam/train_model.py
```diff
@@ -263,6 +271,7 @@ def main(input_args=None):
         standardize=True,
         num_past_forcing_steps=args.num_past_forcing_steps,
         num_future_forcing_steps=args.num_future_forcing_steps,
+        use_all_ensemble_members=not args.use_only_one_ensemble_member,
```
joeloskarsson: Would be better to align the argparse agument and DataModule argument, to avoid the `not` here.

neural_lam/weather_dataset.py
```diff
@@ -87,11 +94,21 @@ def __init__(
...
+                if part == "forcing" and "ensemble_member" in da.dims:
+                    dim_order_with_ensemble = list(expected_dim_order)
+                    grid_index_pos = dim_order_with_ensemble.index(
+                        "grid_index"
+                    )
+                    dim_order_with_ensemble.insert(
+                        grid_index_pos, "ensemble_member"
+                    )
+                    expected_dim_orders.append(tuple(dim_order_with_ensemble))
+                if da.dims not in expected_dim_orders:
```
joeloskarsson: This seems like it should not sit here, but rather that `expected_dim_order` should return dims including ensemble when relevant? Or if this is not suitable we should add some comment here explaining the logic.

neural_lam/weather_dataset.py
```diff
@@ -154,12 +166,26 @@ def __len__(self):
...
                 - self.ar_steps
                 - max(2, self.num_past_forcing_steps)
                 - self.num_future_forcing_steps
             )
+        if self.datastore.is_ensemble and self.use_all_ensemble_members:
+            return base_len * self.da_state.ensemble_member.size
+        return base_len
+
+    def _warn_using_single_ensemble_member(self):
+        if self._warned_single_ensemble_member:
```
joeloskarsson: Could we not just want for this once, during initialization, to avoid having to introduce his `_warned_single_ensemble_member` state into the class? Now it will trigger on requesting the dataset length, which feels a bit arbitrary.

joeloskarsson: I see, it triggers also on other function calls. Still think it would be sufficient to warn during init.

tests/test_cli.py
```diff
@@ -10,3 +10,148 @@ def test_import():
     assert neural_lam is not None
     assert neural_lam.create_graph is not None
     assert neural_lam.train_model is not None
+
+
+def test_train_model_defaults_to_using_all_ensemble_members(monkeypatch):
```
joeloskarsson: This and the next test contain a lot of duplicate code, so should at least be parametrized.

However, are we really sure these tests are needed? From my understanding these mainly test that the argparse argument is actually sent to the data module. While that is can of course be good to check, this is _a lot_ of complexity to add just for this. I lean towards this being too much to add to test something a simple as making sure the argument gets passed along.

joeloskarsson: The more important thing to test here is that we actually do load all the ensemble members in the data as samples.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased](https://github.com/mllam/neural-lam/compare/v0.5.0...HEAD)
 
+### Changed
+- Use all ensemble members by default for ensemble datastores in `WeatherDataset` / `WeatherDataModule`; single-member behavior now requires explicitly opting out.
+
+### Added
+
+- Add support for iterating all ensemble members as independent samples in `WeatherDataset` by default, with matching ensemble-member selection for forcing when available
```
joeloskarsson: This seems like largely the same thing. I would put this all under one entry as "changed". Or alternatively the change to all ens members for state under "Changed" and the ability to use ens forcing as "Added".

joeloskarsson (changes requested):
Thanks for looking into this! I added some comments and ideas 😄

neural_lam/train_model.py
kshirajahere: Renamed this to `--load_single_member` in the latest update. I agree with u about it being too wordy... also it now matches the DataModule / Dataset naming as well.

neural_lam/train_model.py
kshirajahere: Aligned this now. The CLI, `WeatherDataModule`, and `WeatherDataset` all use `load_single_member`, so the `not ...` inversion is gone.

neural_lam/weather_dataset.py
kshirajahere: I agreed that this logic should not live as a forcing-specific exception in `WeatherDataset`.

In the latest update I moved the responsibility into the shared datastore dim-order helper instead: `BaseDatastore.expected_dim_order(...)` now takes whether the concrete array includes `ensemble_member`, and the datastore/data-loading call sites use that. `WeatherDataset` now just validates against the datastore-provided order rather than injecting a local special case for forcing.

neural_lam/weather_dataset.py
kshirajahere: Moved this to dataset initialization now, so it warns once when `load_single_member=True` is used with ensemble data. I also removed the repeated/per-call warning behavior.

tests/test_cli.py
kshirajahere: I dropped the CLI-plumbing style tests and replaced them with dataset-behavior tests instead.

The new tests check that:
- all ensemble members are exposed as independent samples by default,
- flat indexing is time-major across ensemble members,
- forcing follows the same ensemble member when forcing has an `ensemble_member` dimension,
- forcing stays shared when it does not,
- and the forecast case scales dataset length the same way.

CHANGELOG.md
kshirajahere: Consolidated this into a single `Changed` entry now.

kshirajahere:
Latest push addresses the review points:
- renamed the flag to `--load_single_member`,
- aligned the CLI / DataModule / Dataset argument naming,
- moved optional ensemble-dimension handling into the shared datastore dim-order helper instead of keeping it as a local forcing workaround in `WeatherDataset`,
- moved the single-member warning to init-time,
- replaced CLI plumbing tests with dataset behavior tests,
- and consolidated the changelog entry.


@joeloskarsson thank you so much for your review. Please have a look at the latest changes and if any improvements are there let me know always there to help out, sorry for late response 😄. Hoping to hear from you soon 🙏

neural_lam/datastore/base.py
```diff
@@ -391,7 +391,9 @@ def state_feature_weights_values(self) -> List[float]:
 
     @functools.lru_cache
     def expected_dim_order(
-        self, category: Optional[str] = None
+        self,
+        category: Optional[str] = None,
+        has_ensemble_member: Optional[bool] = None,
```
leifdenby: I actually think adding an argument here to set whether the dataset has an ensemble member is dangerous. The reason being that we then have two sources of truth 1) `self.is_ensemble` and 2) this call argument `has_ensemble_member`.

Instead I think we should expect any implementations of the `BaseDatastore` abstract base class to set `self.is_ensemble` during it's `__init__` call, in the same way that we set for example `dim_order` here: https://github.com/mllam/neural-lam/pull/332/changes#diff-2bfda98883376c9b1e398bb0c4dc184138170d4d0d66a994beb7e1b3a2a74968R116. We know that `state` data must be available in the loaded `xr.Dataset` so we can check for an ensemble member dimension there in `__init__` already

kshirajahere: Thanks, @leifdenby . I removed the `has_ensemble_member` call argument from BaseDatastore.expected_dim_order() so there is no longer a per-call override that can disagree with datastore state.

kshirajahere:
> I think we should simplify this a bit and avoid two sources of truth wrt whether a dataset is an ensemble dataset or not

Updated! The idea is that whether the dataset is ensemble-valued still comes from the datastore itself, with no second source of truth at the `expected_dim_order()` call site.

neural_lam/datastore/base.py
```diff
@@ -447,6 +446,20 @@ def expected_dim_order(
 
         return tuple(dim_order)
 
+    def _category_has_ensemble_member(
+        self, category: Optional[str] = None
+    ) -> bool:
+        """Return whether a datastore category includes `ensemble_member`."""
+        if category is None or category == "static":
```
leifdenby: here you always return `False` if category `static` data is requested, but elsewhere you define a dict `_has_ensemble_member` which has a `static` entry, that doesn't seem consistent :)

neural_lam/datastore/base.py
```diff
@@ -447,6 +446,20 @@ def expected_dim_order(
 
         return tuple(dim_order)
 
+    def _category_has_ensemble_member(
```
leifdenby: rather than introducing this `_has_ensemble_member_dim` dict on `self` maybe we add more variable in addition to the `self.is_ensemble` we already have. Since it is only for forcing data we have maybe have an ensemble member `dim` (as well as `state` category data). How about we introduce `self.has_ensemble_of_forcings` or something like that? Then we don't need the `_category_has_ensemble_member` function.

What do you think?

neural_lam/datastore/base.py
kshirajahere: I removed that inconsistency.
`static` no longer participates in any separate ensemble-member metadata path here. Instead of the per-category dict, the base contract now uses:
- `self.is_ensemble` for state data
- `self.has_ensemble_forcing` for forcing data

So `expected_dim_order()` no longer has a hidden second source of truth for `static`, and the only category-specific ensemble case beyond state is forcing.

neural_lam/datastore/base.py
kshirajahere: Yes, I think your suggestion is cleaner, and I updated the branch in that direction.

I removed the _has_ensemble_member_dim dict and the _category_has_ensemble_member() helper. The datastore contract now uses:
- self.is_ensemble for whether state data has an ensemble_member dimension
- self.has_ensemble_forcing for whether forcing data has a separate ensemble_member dimension

WeatherDataset now also uses datastore.has_ensemble_forcing directly when deciding whether forcing should be indexed by member, so the representation is explicit both in the datastore and in the sampling path.
I also updated the dummy datastores and generic datastore tests so the forcing-ensemble case is covered through that explicit attribute.

kshirajahere:
> This is getting there, but I think we should think how represent that a datastore contains separate forcing data for each ensemble member

Thanks @leifdenby, I have updated the PR as requested ! :)

neural_lam/datastore/base.py
```diff
@@ -35,9 +35,11 @@ class BaseDatastore(abc.ABC):
...
 
     # Ensemble vs deterministic data
-    If the datastore is used to represent ensemble data, then the `is_ensemble`
-    attribute should be set to True, and returned data from `get_dataarray` is
-    assumed to have an `ensemble_member` dimension.
+    If state data is ensemble-valued, then the `is_ensemble` attribute should
+    be set to True and returned state data from `get_dataarray` is assumed to
+    have an `ensemble_member` dimension. If forcing data has a separate
+    ensemble-member dimension, that should be represented by the
+    `has_ensemble_forcing` attribute.
```
leifdenby: should we add that the return data array from get_dataarray should have an `ensemble_member` dimension as well?

neural_lam/datastore/base.py
```diff
@@ -243,8 +246,10 @@ def get_dataarray(
         elapsed_forecast_duration)` dimensions if `is_forecast` is True, or
         `(time)` if `is_forecast` is False.
 
-        If the data is ensemble data, the dataarray is expected to have an
-        additional `ensemble_member` dimension.
+        If state data is ensemble-valued, the returned state dataarray is
+        expected to have an additional `ensemble_member` dimension. Forcing
+        data may also have a separate `ensemble_member` dimension when the
```
leifdenby: I don't think should be a "may also have..." isn't the fact either forcing data array should or should not have that an `ensemble_member` dimension depending on the value of `self.has_ensemble_forcing`?

neural_lam/weather_dataset.py
```diff
@@ -48,6 +52,7 @@ def __init__(
         ar_steps: int = 3,
         num_past_forcing_steps: int = 1,
         num_future_forcing_steps: int = 1,
+        load_single_member: bool = False,
```
leifdenby, line 56: if me make using all the ensemble members the default we should make sure to write in the changelog that are changing the default behaviour :)

neural_lam/weather_dataset.py
```diff
@@ -178,12 +185,15 @@ def __len__(self):
...
             #     time steps are required for the initial state)
             #   - future forcing: self.num_future_forcing_steps
-            return (
+            base_len = (
                 len(self.da_state.time)
                 - self.ar_steps
                 - max(2, self.num_past_forcing_steps)
                 - self.num_future_forcing_steps
             )
+        if self.datastore.is_ensemble and not self.load_single_member:
```
leifdenby, line 199: in principle this is correct also if we have an ensemble of analysis steps, but I'm not sure I can think of a situation where this would happen...

neural_lam/weather_dataset.py
```diff
@@ -377,39 +387,33 @@ def _build_item_dataarrays(self, idx):
...
-            # for the ensemble member and then increasing self.__len__ to
-            # include all ensemble members
-            warnings.warn(
-                "only use of ensemble member 0 (the first member) is "
-                "implemented for ensemble data"
-            )
-            i_ensemble = 0
+            n_ensemble_members = self.da_state.ensemble_member.size
+            if not self.load_single_member:
+                sample_idx, i_ensemble = divmod(idx, n_ensemble_members)
```
leifdenby, line 402: neat! I hadn't seen this before :)

tests/test_datasets.py
leifdenby, line 1: these tests are great! well done 🌟

neural_lam/datastore/base.py
kshirajahere: Yes, I updated the class-level contract to make that explicit as well: if is_ensemble is true the returned state dataarray from get_dataarray() is expected to carry ensemble_member, and forcing is handled analogously through has_ensemble_forcing.

neural_lam/datastore/base.py
kshirajahere: Yes, I tightened that wording too.

kshirajahere:
Thanks for the review (sorry for the slack pings 😅 ),I addressed the latest review points.

And thanks for the feedback on the tests. I wanted the behavioural contract around default all-member loading and forcing-member selection to be explicit there as well.

neural_lam/datastore/base.py
kshirajahere: The get_dataarray() contract now states it deterministically: if has_ensemble_forcing=True, the returned forcing dataarray is expected to have ensemble_member; otherwise it is expected not to.

neural_lam/datastore/base.py
```diff
@@ -243,8 +247,11 @@ def get_dataarray(
         elapsed_forecast_duration)` dimensions if `is_forecast` is True, or
         `(time)` if `is_forecast` is False.
 
-        If the data is ensemble data, the dataarray is expected to have an
-        additional `ensemble_member` dimension.
+        If state data is ensemble-valued, the returned state dataarray is
+        expected to have an additional `ensemble_member` dimension. If
```
leifdenby:
```suggestion
        If we have multiple ensemble members of state data, the returned state dataarray is
        expected to have an additional `ensemble_member` dimension. If
```

neural_lam/datastore/base.py
```diff
@@ -35,9 +35,12 @@ class BaseDatastore(abc.ABC):
     dimensions (rather than just `time`).
 
     # Ensemble vs deterministic data
-    If the datastore is used to represent ensemble data, then the `is_ensemble`
-    attribute should be set to True, and returned data from `get_dataarray` is
-    assumed to have an `ensemble_member` dimension.
+    If state data is ensemble-valued, then the `is_ensemble` attribute should
+    be set to True and returned state data from `get_dataarray` is expected to
```
leifdenby: I would keep something closer to the original formulation here, "ensemble-valued" doesn't really make sense, maybe something like this:

```suggestion
    If the datastore is used to present a ensemble of states (rather than a single realisation) for example for forecast ensembles, then the `is_ensemble` attribute should
    be set to `True` and returned state data from `get_dataarray` is expected to
```

neural_lam/datastore/base.py
```diff
@@ -35,9 +35,12 @@ class BaseDatastore(abc.ABC):
...
+    have an `ensemble_member` dimension. If `has_ensemble_forcing` is True,
+    returned forcing data from `get_dataarray` is expected to have an
```
leifdenby:
```suggestion
    have an `ensemble_member` dimension. If each ensemble member has its own forcing values then `has_ensemble_forcing` should be set to `True`,
    and the returned forcing data from `get_dataarray` is expected to have an
```

neural_lam/datastore/mdp.py
```diff
@@ -251,8 +256,10 @@ def get_dataarray(
         elapsed_forecast_duration)` dimensions if `is_forecast` is True, or
         `(time)` if `is_forecast` is False.
 
-        If the data is ensemble data, the dataarray will have an additional
-        `ensemble_member` dimension.
+        If state data is ensemble-valued, the returned state dataarray will
```
leifdenby: adjust this wording similarly to what I suggested for the datastore base class

neural_lam/weather_dataset.py
```diff
@@ -37,6 +37,11 @@ class WeatherDataset(torch.utils.data.Dataset):
         forcing from times t, t+1, ..., t+j-1, t+j (and potentially times before
         t, given num_past_forcing_steps) are included as forcing inputs at time
         t. Default is 1.
+    load_single_member : bool, optional
+        If `False` and the datastore returns ensemble-valued state data,
```
leifdenby: also here, rather than "ensemble-valued" we should write this "an ensemble of state realisations" or something like that

neural_lam/weather_dataset.py
leifdenby: actually, this could make sense in a setting where we are using data assimilation where we create multiple likely initial state estimates, that would be an ensemble of analysis timesteps

tests/dummy_datastore.py
```diff
@@ -352,9 +359,9 @@ def get_dataarray(
             The xarray DataArray object with processed dataset.
 
         """
+        da_category = self.ds[category]
         dim_order = self.expected_dim_order(category=category)
-
-        da_category = self.ds[category].transpose(*dim_order)
+        da_category = da_category.transpose(*dim_order)
```
leifdenby: why is this change required? Defining `da_category` and locally overwriting is error-prone, so unless we need `da_category` to compute the `dim_order` we should leave this unchanged

neural_lam/datastore/base.py
kshirajahere: @leifdenby Applied, I updated the wording to describe the datastore as presenting an ensemble of state realisations, and made the returned `get_dataarray()` contract explicit for that case.

neural_lam/datastore/base.py
kshirajahere: Squeezed it so it is now stated in terms of whether each ensemble member has its own forcing values, via has_ensemble_forcing=True.

neural_lam/datastore/mdp.py
kshirajahere: consolidated there as well to mirror the base datastore wording

neural_lam/weather_dataset.py
kshirajahere: Done

tests/dummy_datastore.py
kshirajahere: Removed that local overwrite and kept the simpler direct transpose, since dim_order does not depend on the already-transposed array, Hope its good now :)

kshirajahere:
@leifdenby Doneee:) Ready to make changes if required

neural_lam/weather_dataset.py
joeloskarsson: Nice, good solution!

neural_lam/weather_dataset.py
joeloskarsson: I think it is generally better to describe what happens when the flag is set to True, rather than when False (which is now default behavior)

kshirajahere:
Thank you @joeloskarsson @leifdenby  Delighted seeing my first PR merge @neural-lam🥳, excited to contribute more

Issue #346, opened 2026-03-06
Validate graph structure before model initialization


Saptami191:
### Problem

Graphs used in Neural-LAM must match the dataset geometry (number of grid nodes and node connectivity). If an incompatible or corrupted graph is loaded, the model may fail during initialization or training because the graph structure does not match the dataset. Neural-LAM relies on graphs that represent spatial relationships between grid and mesh nodes, so structural mismatches can break the training pipeline. :contentReference[oaicite:0]{index=0}

###  Proposed solution

Added **graph metadata validation during graph loading** in `BaseGraphModel`.

A validation function `validate_graph_metadata()` was introduced and executed immediately after loading the graph. This function checks:

- Required graph tensors exist (`g2m_edge_index`, `m2g_edge_index`, `mesh_static_features`)
- The number of grid nodes in the graph matches the dataset grid nodes
- Edge indices do not reference non-existent nodes

If the graph is incompatible with the dataset, the model raises a clear `ValueError` before training starts.This ensures that only structurally valid graphs are accepted during model initialization.

#### Before
Graph loading did not perform validation.
Dataset
↓
Graph generation
↓
Graph loaded in BaseGraphModel
↓
Training starts
#### After
Graph metadata validation is performed immediately after loading the graph.
Dataset
↓
Graph generation
↓
Graph loaded in BaseGraphModel
↓
validate_graph_metadata()
↓
Training starts

Saptami191:
@sadamov  kindly review the PR when you have time and assign it if appropriate. Thank you!

sadamov:
Closing in favour of #408 (graph sanity-check CLI discussion) and #323 (graph-on-disk spec + validator). The load-time enforcement angle is worth revisiting once #323 lands.

PR #349, opened 2026-03-06
Validate graph structure before model initialization

Saptami191:
## Describe your changes

This PR introduces graph metadata validation during graph loading in `BaseGraphModel`.

Previously, graphs were loaded without verifying that their structure matched the dataset geometry. If an incompatible or corrupted graph was loaded, the model could fail later during initialization or training.

This change adds a validation function `validate_graph_metadata()` that runs immediately after loading the graph. The function checks:

- Required graph tensors exist (`g2m_edge_index`, `m2g_edge_index`, `mesh_static_features`)
- The number of grid nodes in the graph matches the dataset grid nodes
- Edge indices do not reference non-existent nodes

If the graph is incompatible, a clear `ValueError` is raised before training begins.

### Motivation

This improves the robustness of the training pipeline by detecting invalid or incompatible graphs early. It is particularly useful when graphs are loaded externally (e.g., via flexible graph loading workflows such as `--graph_path`).

### Dependencies

No new dependencies were introduced.

## Solves issue
#346 

## Type of change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📖 Documentation

- [x] For any new/modified functions/classes I have added docstrings
- [x] I have placed in-line comments where necessary
- [ ] I have added tests that prove my feature works
- [x] I have given the PR a clear name

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented

## Checklist for assignee

- [ ] the tests pass
- [ ] PR assigned to milestone
- [ ] author added changelog entry

### DIFF
--- a/neural_lam/models/base_graph_model.py
+++ b/neural_lam/models/base_graph_model.py
@@ -18,27 +18,20 @@ def validate_graph_metadata(graph_ldict, expected_grid_nodes):
 
     for key in required_keys:
         if key not in graph_ldict:
-            raise ValueError(f"Invalid graph: missing required tensor '{key}'.")
+            raise ValueError(
+                "Invalid graph: missing required tensor " f"'{key}'."
+            )
 
-    mesh_features = graph_ldict["mesh_static_features"]
+    # Only validate grid nodes if grid_static_features exists
+    if "grid_static_features" in graph_ldict:
+        grid_nodes = graph_ldict["grid_static_features"].shape[0]
 
-    if isinstance(mesh_features, torch.Tensor):
-        mesh_nodes = mesh_features.shape[0]
-    else:
-        mesh_nodes = mesh_features[0].shape[0]
-
-    grid_nodes = graph_ldict["grid_static_features"].shape[0]
-
-    if grid_nodes != expected_grid_nodes:
-        raise ValueError(
-            f"Incompatible graph: expected {expected_grid_nodes} grid nodes "
-            f"but graph contains {grid_nodes}."
-        )
-
-    if graph_ldict["g2m_edge_index"].max().item() >= total_nodes:
-        raise ValueError(
-            "Invalid graph: edge indices reference non-existent nodes."
-        )
+        if grid_nodes != expected_grid_nodes:
+            raise ValueError(
+                "Incompatible graph: expected "
+                f"{expected_grid_nodes} grid nodes "
+                f"but graph contains {grid_nodes}."
+            )
 
 
 class BaseGraphModel(ARModel):
--- a/neural_lam/models/base_graph_model.py
+++ b/neural_lam/models/base_graph_model.py
@@ -10,6 +10,7 @@ from .ar_model import ARModel
 
 
 def validate_graph_metadata(graph_ldict, expected_grid_nodes):
+
     required_keys = [
         "g2m_edge_index",
         "m2g_edge_index",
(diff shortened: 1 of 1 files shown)


neural_lam/models/base_graph_model.py
```diff
@@ -9,6 +9,31 @@
...
+
+    for key in required_keys:
+        if key not in graph_ldict:
+            raise ValueError(
+                "Invalid graph: missing required tensor " f"'{key}'."
+            )
+
+    # Only validate grid nodes if grid_static_features exists
+    if "grid_static_features" in graph_ldict:
+        grid_nodes = graph_ldict["grid_static_features"].shape[0]
```
kshirajahere:
```suggestion
        if "grid_static_features" in graph_ldict:
                    grid_nodes = graph_ldict["grid_static_features"].shape[0]
```
I think node-count check may never run in current code path as `utils.load_graph()` does not seem to return `grid_static_features` in `graph_ldict`.
So unless I’m missing another path(possible), this branch is never entered, which means the PR currently does not validate graph-vs-dataset grid-node compatibility as described. Although i would love thoughts of maintainers on this aswell...

neural_lam/models/base_graph_model.py
```diff
@@ -9,6 +9,31 @@
...
 
 
+def validate_graph_metadata(graph_ldict, expected_grid_nodes):
+    required_keys = [
+        "g2m_edge_index",
+        "m2g_edge_index",
+        "mesh_static_features",
+    ]
+
+    for key in required_keys:
```
kshirajahere, line 20:
```suggestion
    for key in required_keys:
            if key not in graph_ldict:
                  raise ValueError(
                         "Invalid graph: missing required tensor " f"'{key}'."
                  )
```
I got  a doubt as `utils.load_graph()` loads these graph files before `validate_graph_metadata()` is called, wouldn’t a missing required tensor/file already fail during loading?

kshirajahere (commented):
Hello, so current implementation misses two of three checks it claims to add. I mapped them to the inline comments.

neural_lam/models/base_graph_model.py
Saptami191: Good point thanks for pointing that out.u r right that utils.load_graph() already loads the graph tensors from disk, so a missing file wwould likely fail earlier during loading. thee intention of this check was mainly to validate the structure of the loaded dictionary rather than file existence, but I agree that this may be redundant depending on how load_graph handles missing files.and grid_static_features, u r also correct that in the current code path utils.load_graph() does not appear to include it in graph_ldict. That means the node-count validation branch would not run in practice.my intention here was to make the validation tolerant to graphs that do not include grid_static_features while still performing the node count check when the tensor is available. but given the current behavior of load_graph(), this check may indeed never trigger.i woud appreciate thoughts from maintainers on whether grid_static_features is expected to be part of graph_ldict in future graphs or whether validation against datastore.num_grid_points should be implemented differently.

Saptami191:
> Hello, so current implementation misses two of three checks it claims to add. I mapped them to the inline comments.

I’ll update the implementation to ensure the intended validations actually occur

Saptami191:
@joeloskarsson  Kindly give a review!

neural_lam/models/base_graph_model.py
```diff
@@ -25,6 +49,10 @@ def __init__(self, args, config: NeuralLAMConfig, datastore: BaseDatastore):
         self.hierarchical, graph_ldict = utils.load_graph(
```
Omnagar, line 49: Nice addition — validating the graph metadata early should definitely help catch
incompatible graphs before training starts.

One small question: since the graph tensors are registered as buffers
immediately after `utils.load_graph()`, does the validation run before or
after these buffers are attached to the module?

If validation fails, would it make sense to perform the metadata checks before
registering the buffers, so the model isn’t partially initialized with invalid
graph tensors?

Omnagar (commented):
Left a small question regarding when graph metadata validation
runs relative to buffer registration. Overall this looks like a
useful safeguard for catching invalid graphs earlier in the
training pipeline.

sadamov:
Thanks @Saptami191. The validation here largely overlaps with #323, which adds a full graph-on-disk format spec and validator with CI integration, and that was the path we settled on in #408. The "validate at model init" angle could still be useful as a follow-up once #323 lands (i.e. enforcing the spec at load time inside `BaseGraphModel`), but I'd prefer not to add a parallel shallower validator in the meantime. I propose to close this in favour of #323. If you'd like to help, please pitch in there.

Issue #83, opened 2024-10-28
Add graph creation functionality using weather-model-graphs


sadamov:
### Overview
This PR adds functionality to create and save different graph architectures using the `weather-model-graphs` package. The implementation includes support for three graph types (currently):
- Keisler graphs (single-scale mesh)
- GraphCast (multi-scale)
- Oskarsson hierarchical graphs

*Reference*
https://github.com/mllam/neural-lam/pull/66#issuecomment-2434482120

### New Features
- Script to generate graphs from datastore xy coordinates
- Integration with `weather-model-graphs` package
- Support for multiple datastore types (MDP, NPYFilesMEPS)
- Graph serialization to PyTorch format

### Implementation Details
```python
from pathlib import Path
import weather_model_graphs as wmg
from neural_lam.datastore import init_datastore

def create_graphs(datastore):
    xy = datastore.get_xy("state", stacked=False)
    graphs = {
        'keisler': wmg.create.archetype.create_keisler_graph(xy_grid=xy),
        'graphcast': wmg.create.archetype.create_graphcast_graph(xy_grid=xy),
        'hierarchical': wmg.create.archetype.create_oscarsson_hierarchical_graph(xy_grid=xy)
    }
    return graphs
```

### Known Issues
1. Graph serialization with `wmg.save.to_pyg()` not working correctly
2. Need to determine correct edge/node feature naming convention

### TODO
- [ ] Expand/simplify PyG serialization
- [ ] Adapt tests for graph creation with wgm
- [ ] Document graph properties and usage

### Questions
1. What's the preferred format for storing generated graphs?
2. Should we include visualization tools?
3. How closely do we want to interate `wgm` into `nl` (slang for neural-lam, obviously)

joeloskarsson:
I will take this on, as there is a need to more tightly integrate neural-lam and wmg for #84 . Or, in particular, neural-lam will have to call wmg with specific arguments, so we can add a dependency from neural-lam to wmg and then graph-creation scripts in neural-lam that call wmg. 

For #84 this will be needed to impose a specific ordering between interior grid nodes and boundary grid nodes. By calling wmg directly from neural-lam we can assure that these calls always place all interior nodes before all boundary nodes.

muskan-1301:
Hi @joeloskarsson I see this issue has been open for a while and I'm interested in helping move it forward. I've been looking at the weather-model-graphs package and the proposed integration.

[...]

PR #373, opened 2026-03-10
Fix square mesh assumption in create_graph to support non-square domains

### LINKED ISSUE #371: `create_graph` always generates square mesh grids, ignoring data grid aspect ratio
## Problem
In line 244-259, the mesh graph is always square regardless of the data grid dimensions:

```py
# line 243
nx = 3  # number of children =nx**2
nlev = int(np.log(max(xy.shape[:2])) / np.log(nx)) # Uses largest dimension
nleaf = nx**nlev  

# line 258
n = int(nleaf / (nx**lev))
g = mk_2d_graph(xy, n, n) # Same n for both x and y
```

For a data grid of shape `(Nx, Ny)`, the mesh is always `n x n`, where n is derived from `max(Nx, Ny)`. In addition, the `dm` calculation only measures the x-direction, assuming that the x direction is the largest:
```py
dm = np.sqrt(
    np.sum((vm.data("pos")[(0, 1, 0)] - vm.data("pos")[(0, 0, 0)]) ** 2)
)
```
Since the mesh may not be uniform, this wont correctly capture the y-direction relationship.

## Why it matters
Using the largest dimension means that for a case where the data grid has a non-square aspect ratio will produce multiple redundant node, wasting computation. This also comes with the added side effect of having uneven spacing, meaning that the weather moving diagonally will look different depending on the direction relative to the mesh, because it treats x and y as different resolution.

For example, for a 200x30 domain (6.7:1 aspect ratio), the original code produces a 27x27 mesh where y-nodes would be 6.7x more densely packed than x-nodes.

## How to reproduce/visualize
The following is the script that is used to visualize the problem
```py
import torch
import tempfile
import os
import numpy as np
from neural_lam.create_graph import create_graph

Nx, Ny = 200, 30
x = np.linspace(0, Nx, Nx)
y = np.linspace(0, Ny, Ny)
xx, yy = np.meshgrid(x, y, indexing='ij')
xy = np.stack([xx, yy], axis=-1)

print(f"Data grid: {Nx}x{Ny} (aspect ratio {Nx/Ny:.1f}:1)")

with tempfile.TemporaryDirectory() as tmp:
    create_graph(tmp, xy)

    mesh = torch.load(os.path.join(tmp, 'mesh_features.pt'))
    pos  = mesh[0]

    x_unique = len(torch.unique(pos[:, 0]))
    y_unique = len(torch.unique(pos[:, 1]))

    dx_phys = Nx / x_unique
    dy_phys = Ny / y_unique

    print(f"Mesh shape: {x_unique}x{y_unique} = {pos.shape[0]} nodes")
    print(f"x spacing: {dx_phys:.1f} units/node")
    print(f"y spacing: {dy_phys:.1f} units/node")
    print(f"Spacing ratio: {dx_phys/dy_phys:.1f}:1  (ideal: 1.0:1)")
    print(f"Redundant y nodes: ~{y_unique - round(y_unique * Ny/Nx)}"
          f" of {y_unique} rows cover the same grid points")
```
The output given for `200x30` is:
```
Mesh shape: 27x27 = 729 nodes
x spacing: 7.4 units/node
y spacing: 1.1 units/node
Spacing ratio: 6.7:1
Redundant y nodes: ~23 of 27 rows cover the same grid points 
```

## Proposed implementation
At each mesh level, both `n_x` and `n_y` should be scaled independently to match the physical aspect ratio of the domain, rather than using the same value n for both dimensions. The dm connection radius should also be updated to correctly reflect the non-square mesh cell geometry to ensure full grid coverage.

## Related

Issue #4

osten-antonio:
## Describe your changes

`create_graph` previously generated square mesh graphs regardless of the data grid's physical aspect ratio, by passing the same value `n` for both x and y dimensions of `mk_2d_graph`. This caused two concrete problems, as stated in issue #371 :
1. Uneven mesh spacing: for a 200x30 domain (6.7:1 aspect ratio), the original 27x27 mesh packs y-nodes 6.7x more densely than x-nodes. This means the model sees different effective resolutions in each direction — weather features moving diagonally will appear different depending on their orientation relative to the mesh.
2. Redundant nodes: some of the rows of the smaller dimension in the original mesh cover the same grid points, wasting computation with no benefit to model accuracy.

**Changes:**
- At each mesh level, `n_x` and `n_y` are now derived independently by scaling the larger dimension proportionally to the physical 
  domain extent (`x_extent / y_extent`), assigning `n_larger` to whichever axis has greater physical extent.
- `mesh_dims` tracks `(n_x, n_y)` per level, replacing the square assumption in the merge loop reshape.
- `dm` is now computed from the mesh cell diagonal (`sqrt(dx**2 + dy**2)`) rather than x-direction spacing only. This would also guarantee full grid coverage with zero orphan node, which was a problem when migrating to a non square resolution.

No new dependencies.

## Issue Link

Closes #371 

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/neural_lam/create_graph.py
+++ b/neural_lam/create_graph.py
@@ -256,17 +256,24 @@ def create_graph(
 
     # multi resolution tree levels
     G = []
-    mesh_dims = []  # for tracking per-level dimensions, used for reshape and dm
+    # for tracking per-level dimensions, used for reshape and dm
+    mesh_dims: list[tuple[int, int]] = []
     for lev in range(1, mesh_levels + 1):
-        n_larger = int(nleaf / (nx**lev))
-        # scale n_x and n_y proportionally to physical domain aspect ratio,
-        # assigning n_larger to whichever direction has greater physical extent
-        if x_extent >= y_extent:
-            n_x = n_larger
-            n_y = max(1, round(n_larger * y_extent / x_extent))
+        if lev == 1:
+            n_larger = int(nleaf / (nx**lev))
+            # scale n_x and n_y proportionally to physical domain aspect ratio,
+            # assign n_larger to whichever direction has greater physical extent
+            if x_extent >= y_extent:
+                n_x = n_larger
+                n_y = max(1, round(n_larger * y_extent / x_extent))
+            else:
+                n_y = n_larger
+                n_x = max(1, round(n_larger * x_extent / y_extent))
         else:
-            n_y = n_larger
-            n_x = max(1, round(n_larger * x_extent / y_extent))
+            # derive from actual slice of previous level
+            n_x_prev, n_y_prev = mesh_dims[lev - 2]
+            n_x = len(range(1, n_x_prev, nx))
+            n_y = len(range(1, n_y_prev, nx))
 
         if (n_y < 3 or n_x < 3) and lev > 1:  # always allow level 1
             mesh_levels = lev - 1
@@ -448,7 +455,16 @@ def create_graph(
     # distance between mesh nodes
     dx = x_extent / mesh_dims[0][0]  # cell width at finest mesh level
     dy = y_extent / mesh_dims[0][1]  # cell height at finest mesh level
-    dm = np.sqrt(dx**2 + dy**2)  # diagonal of mesh cell (coverage radius basis)
+
+    # compare using mesh_dims because dx dy can cause rounding issue
+    if mesh_dims[0][0] == mesh_dims[0][1]:
+        dm = np.sqrt(
+            np.sum((vm.data("pos")[(0, 1, 0)] - vm.data("pos")[(0, 0, 0)]) ** 2)
+        )
+    else:
+        dm = np.sqrt(
+            dx**2 + dy**2
+        )  # diagonal of mesh cell (coverage radius basis)
 
     # grid nodes
     Nx, Ny = xy.shape[:2]
--- a/neural_lam/create_graph.py
+++ b/neural_lam/create_graph.py
@@ -245,6 +245,8 @@ def create_graph(
     nlev = int(np.log(max(xy.shape[:2])) / np.log(nx))
     nleaf = nx**nlev  # leaves at the bottom = nleaf**2
 
+    x_extent = xy[:, :, 0].max() - xy[:, :, 0].min()
+    y_extent = xy[:, :, 1].max() - xy[:, :, 1].min()
     mesh_levels = nlev - 1
     if n_max_levels:
         # Limit the levels in mesh graph
@@ -254,14 +256,28 @@ def create_graph(
 
     # multi resolution tree levels
     G = []
+    mesh_dims = []  # for tracking per-level dimensions, used for reshape and dm
     for lev in range(1, mesh_levels + 1):
-        n = int(nleaf / (nx**lev))
-        g = mk_2d_graph(xy, n, n)
+        n_larger = int(nleaf / (nx**lev))
+        # scale n_x and n_y proportionally to physical domain aspect ratio,
+        # assigning n_larger to whichever direction has greater physical extent
+        if x_extent >= y_extent:
+            n_x = n_larger
+            n_y = max(1, round(n_larger * y_extent / x_extent))
+        else:
+            n_y = n_larger
+            n_x = max(1, round(n_larger * x_extent / y_extent))
+
+        if (n_y < 3 or n_x < 3) and lev > 1:  # always allow level 1
+            mesh_levels = lev - 1
+            break
+        g = mk_2d_graph(xy, n_x, n_y)
         if create_plot:
             plot_graph(from_networkx(g), title=f"Mesh graph, level {lev}")
             plt.show()
 
         G.append(g)
+        mesh_dims.append((n_x, n_y))
 
     if hierarchical:
         # Relabel nodes of each level with level index first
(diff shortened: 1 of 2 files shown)


neural_lam/create_graph.py
```diff
@@ -254,14 +256,28 @@ def create_graph(
 
     # multi resolution tree levels
     G = []
+    mesh_dims = []  # for tracking per-level dimensions, used for reshape and dm
     for lev in range(1, mesh_levels + 1):
-        n = int(nleaf / (nx**lev))
-        g = mk_2d_graph(xy, n, n)
+        n_larger = int(nleaf / (nx**lev))
```
kshirajahere: This rounding scheme is not compatible with the existing [1::3, 1::3] parent-selection logic below. For a rectangular grid like xy.shape == (100, 600), this produces level dims (14, 81) and then (4, 27), but sampling the previous level yields (5, 27) parents, so create_graph(..., n_max_levels=2) crashes at line 399 with ValueError: cannot reshape array of size 270 into shape (108,2). The coarse dims need to be derived from the sampled previous level, or the relabel/compose logic needs to change.

tests/test_graph_creation.py
```diff
@@ -117,3 +118,24 @@ def test_graph_creation(datastore_name, graph_name):
                         assert r.shape[0] == 2  # adjacency matrix uses two rows
                     elif file_id.endswith("_features"):
                         assert r.shape[1] == d_features
+
+
+def test_graph_creation_non_square_aspect_ratio():
```
kshirajahere, line 123: This test only covers the n_max_levels=1 case, so it misses the actual breakage introduced by the new sizing logic. Please add at least one rectangular multiscale case (n_max_levels=2) and one square-grid regression check for unchanged g2m connectivity, otherwise both failures above still pass CI.

neural_lam/create_graph.py
```diff
@@ -429,9 +446,9 @@ def create_graph(
     vm = G_bottom_mesh.nodes
     vm_xy = np.array([xy for _, xy in vm.data("pos")])
     # distance between mesh nodes
-    dm = np.sqrt(
-        np.sum((vm.data("pos")[(0, 1, 0)] - vm.data("pos")[(0, 0, 0)]) ** 2)
-    )
+    dx = x_extent / mesh_dims[0][0]  # cell width at finest mesh level
```
kshirajahere, line 456: This changes g2m connectivity for every existing square graph, not just non-square ones. Previously the radius was based on one mesh-node spacing; here it becomes the cell diagonal, which is sqrt(2) larger on square meshes. On an 81x81 square grid, g2m edges jump from 9009 on main to 17757 here. That is a topology regression unless it was explicitly intended and rebase lined everywhere.

kshirajahere (changes requested):
The aspect-ratio fix is not safe to merge as written. It introduces a hard failure for non-square multiscale graphs because the rounded per-level mesh dimensions no longer match the existing 3x downsampling/relabeling assumption in the non-hierarchical path. A concrete repro is create_graph(..., xy.shape=(100, 600), n_max_levels=2), which now fails at create_graph.py:399 with a reshape error.

It also changes g2m topology globally by switching the coverage radius from single-axis neighbor spacing to the mesh-cell diagonal. On a square 81x81 grid, that increases g2m edges from 9009 on main to 17757 in this PR. The added test only exercises the one-level rectangular case, so neither regression is caught.

I would block this until:

- Multi-level rectangular meshes are generated with level sizes that are consistent with the parent mapping logic.
- Square-grid g2m connectivity is preserved, or the topology change is made explicit and covered with updated tests.

osten-antonio:
Hi @kshirajahere , thanks for the review! I have addressed both issue; fixed n_max_levels = 2 crash by using the previous level's coarse dimension, and on square grids I reverted to using the original dm formula. I have also added the tests covering multi-level rectangular mesh, and square grid g2m edge count preservation. Let me know if there is anything else to address :D

joeloskarsson:
Sorry for not looking at this earlier, but I would not put effort into such improvements to create_graph.py, as this functionality already exists and is much more robust in weather-model-graphs. The plan is to drop create_graph.py in its current form once #83 is complete.

Pratz1337 (changes requested):
Please ignore this review

neural_lam/create_graph.py
```diff
@@ -429,9 +453,18 @@ def create_graph(
...
+    dy = y_extent / mesh_dims[0][1]  # cell height at finest mesh level
+
+    # compare using mesh_dims because dx dy can cause rounding issue
+    if mesh_dims[0][0] == mesh_dims[0][1]:
```
kshirajahere, line 460: This condition should be based on spacing equivalence (for example dx vs dy with tolerance), not only n_x == n_y. Rounded equal dims can happen on non-square domains and cause coverage regressions.

neural_lam/create_graph.py
```diff
@@ -254,14 +256,35 @@ def create_graph(
...
     for lev in range(1, mesh_levels + 1):
-        n = int(nleaf / (nx**lev))
-        g = mk_2d_graph(xy, n, n)
+        if lev == 1:
+            n_larger = int(nleaf / (nx**lev))
+            # scale n_x and n_y proportionally to physical domain aspect ratio,
+            # assign n_larger to whichever direction has greater physical extent
+            if x_extent >= y_extent:
+                n_x = n_larger
+                n_y = max(1, round(n_larger * y_extent / x_extent))
```
kshirajahere, line 268: Please guard degenerate coordinate extents before division. A zero extent on one axis should fail fast with a clear error or use a safe fallback.

kshirajahere (changes requested):
Thanks for addressing the earlier blockers.
The multiscale reshape crash concern and square-grid g2m topology regression were both handled, and the added tests for those paths are useful.
I am requesting some more changes please address them.

kshirajahere:
Also add changelog entry

osten-antonio:
Hi @kshirajahere thanks for the rereview! I think I will follow @joeloskarsson's advice seeing that there is much more development on weather-model-graph end with its multiple layout types.

sadamov:
Thanks @osten-antonio. Closing per your own follow-up - going with @joeloskarsson's earlier suggestion to drop changes to `create_graph.py` in favour of the weather-model-graphs migration tracked in #384 / #596. If any of these fixes still apply on the wmg side once #596 lands, please port them there.

PR #376, opened 2026-03-10
Improve metric heatmap readability and scaling (#375)

### LINKED ISSUE #375: Improve metric heatmap readability and scaling
`plot_error_map` in `neural_lam/vis.py` produces a heatmap of per-variable, per-lead-time errors.
I have three issues with this function that have been bothering me for a while now:

### 1. Misleading function name (and least important :D)

The name `plot_error_map` implies a spatial (geographic) map
**Suggestion**: rename to `plot_error_heatmap` and update all call sites in `ar_model.py`.

### 2. Colormap not suitable for cross-variable comparison

Each row is currently normalised independently to `[0, 1]`. This means:
The colour of a cell encodes *relative* position within a single variable's range. A cell that is dark red for `u100m` and dark red for `t2m` could have very different absolute errors. It is impossible to visually compare error magnitudes across variables.
The numeric annotations partially compensate, but the background heatmap is then misleading rather than helpful.

**Options to consider**:
- Normalise all values to a single global scale
- Use perceptually uniform colormap calibrated to the rescaled errors
- ...

### 3. Font sizes do not adapt to the number of variables

The font size of both axis tick labels and in-cell annotations is currently fixed. With few variables (e.g. 5-10) the text is undersized relative to the cell height; with many variables (30+) the y-tick labels and annotations overlap and become unreadable. Similarly for many lead_times.

Image

**Suggestion**: derive font sizes automatically from the number of variables and prediction steps. Also the size of the table must dynamically adapt the cell size.

kshirajahere:
## Describe your changes

This PR improves the metric heatmap plotting used during evaluation.

### Summary of changes
- renamed `plot_error_map` to `plot_error_heatmap`
- updated the `ARModel` call site and related wording to use “heatmap” consistently
- replaced per-row normalization with a single shared color scale across all variables
- switched to a perceptually uniform colormap
- added a colorbar for absolute-value interpretation
- made figure size adapt to the number of variables and lead times
- made tick-label and in-cell annotation font sizes adapt to heatmap dimensions
- rotate x tick labels automatically for dense lead-time grids
- added focused plotting regression tests for:
  - shared global color scaling
  - adaptive figure/font sizing behavior
- added a changelog entry under `unreleased -> Fixed`

### Motivation / context
Fixes the three issues raised in #375:
- misleading function naming
- heatmap colors not being comparable across variables
- non-adaptive text sizing/layout for dense heatmaps

### Dependencies
No new external dependencies.

## Issue Link

Closes #375

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass

## Tests

Ran locally:
- `python -m pytest -q tests/test_plotting.py -k "heatmap"` -> passed
- `python -m py_compile neural_lam/vis.py neural_lam/models/ar_model.py tests/test_plotting.py` -> passed

Notes:
- `python -m pytest -q tests/test_plotting.py` hits an existing cartopy/Natural Earth data path permission issue on this machine when the older integration plotting tests try to access coastline data. The new focused heatmap tests pass.


sahilkr31:
Hi @sadamov, and thanks @kshirajahere for the PR.

I went through the changes in this PR and the switch to a shared global color scale makes sense, since it allows the heatmap colours to reflect absolute error magnitudes and makes comparisons across variables easier.

One thing I was thinking about regarding point 2 is whether global scaling might sometimes reduce the contrast for variables whose error range is much smaller than others. In those cases a row could end up looking almost uniform even if there are differences across lead times.

Would it make sense to support both behaviours (for example using global scaling by default but optionally allowing per-variable normalization when someone wants to highlight within-variable patterns), or do you think keeping a single global scale is preferable for the evaluation plots here?

Curious to hear your thoughts also.

Joltsy10:
The contrast problem @sahilkr31 points out exists because raw absolute errors across variables with different physical units aren't comparable to begin with. 500 Pa of surface pressure error and 1 K of temperature error sit on completely different scales, so a shared raw colormap will always be dominated by high-magnitude variables regardless of how you normalize it.

Normalizing each variable by its climatological standard deviation before applying the global scale fixes this properly. The colormap then encodes normalized error, how large is this error relative to natural variability, which is actually consistent across variables. Same color, same forecast skill, regardless of units.

The per-variable toggle is just letting the user pick which misleading view they want rather than solving the underlying issue.😄
Happy to be corrected if raw absolute errors are preferred here for a specific reason.

sadamov:
> 500 Pa of surface pressure error and 1 K of temperature error sit on completely different scales, so a shared raw colormap will always be dominated by high-magnitude variables regardless of how you normalize it.
Normalizing each variable by its climatological standard deviation before applying the global scale fixes this properly. The colormap then encodes normalized error, how large is this error relative to natural variability, which is actually consistent across variables. Same color, same forecast skill, regardless of units.

Agreed, this sounds like a good way to do it. We can implement this cleanly by expanding plot_error_map to accept both physical and normalized errors separately. Physical values are always shown as text annotations on the heatmap (unchanged). When errors_norm is also provided, it drives the colormap instead of the default per-variable-max normalization:

```python
def plot_error_map(
    errors,
    datastore: BaseRegularGridDatastore,
    title=None,
    errors_norm=None,
):
    """
    errors: (pred_steps, d_f) — physical-unit values, shown as text annotations
    errors_norm: (pred_steps, d_f) — optional values used for colormap scaling.
        When provided (e.g. errors divided by climatological std), the colormap
        encodes skill relative to natural variability, making colors comparable
        across variables with different units. When None, each variable is
        normalized by its own maximum error.
    """
```

The caller in ar_model passes metric_tensor_averaged (already in normalized units) as errors_norm and metric_tensor_averaged * state_std as errors, so no redundant rescaling occurs.

sadamov:
@kshirajahere could you share some pictures of the new heatmap for low/high number of vars and lead_times?

kshirajahere:
@sadamov Sure 
massive_grid_120leads_40vars
full_real_subset_8vars_65leads


you will need to open in new tab to see how it comes now, without overlapping, i tried to make this test as rigorous as possible.
Some simpler heatmaps-->
small_real_subset_5vars_6leads
medium_real_subset_8vars_20leads


 I've generated some sample heatmaps using the current code from the PR. I used the MEPS sample data to showcase real-world variable structures along with a simulated "massive grid" scenario like the one mentioned in the issue.

kshirajahere:
@sahilkr31 @Joltsy10 @sadamov Thanks for the great discussion here. I agree that comparing raw values with varying physical units like Pressure vs Temp can misrepresent skill when they share a single color scale. 

Based on this, I've ensured the heatmap code correctly handles whatever normalization metric is passed to it by the `ARModel`. The plotting function now uses a global color scale for the matrix provided, meaning that if it receives climatology-normalized errors from the training/evaluation loop (where natural variability is standardized), the heatmap's color mapping will automatically reflect unit-independent forecast skill! And, of course, the annotations continue to show the absolute numeric values properly, avoiding overlapping and adjusting depending on cell size.

joeloskarsson:
Hi, thanks for working on this, and great that you raised the issue @sadamov. These plots have often looked quite bad.

I was a bit worried by the discussion of using a joint colormap, since (as then later pointed out) this would not look reasonable at all due to the magnitude of variables being so different. I like the current idea though, to have the colormap based on error in standardized units and then always the original numbers in text. From the examples it looks like some variables still dominate though. Might need to mix in the variance of the single-step-delta in the standardization to get something reasonable.

Personally I am not much of a fun of the switch to a perceptually uniform colormap for this. The change from white to red clearly shows that things got worse, and with `viridis` one has to consult the color bar to recall e.g. that lighter is worse and that green is not good. But I understand this is a bit personal preference.

kshirajahere:
Thanks Joel @joeloskarsson 
The latest update keeps the numeric annotations in the original units, but changes the background colors to use a relative cross-variable scale instead of raw physical magnitudes. Please review @sadamov @joeloskarsson 

Concretely:
- the heatmap now normalizes the background colors using datastore standardization stats,
- and when `state_diff_std_standardized` is available it uses that as well, so the colors reflect magnitude relative to typical 1-step variability for each variable,
- while the cell annotations still show the original rescaled values.

I also switched the colormap away from `viridis` to a white-to-red sequential scale, since that seemed more in line with your point that “worse” should read visually as red without needing to decode the colorbar.

I added tests for:
- relative cross-variable color scaling,
- the white-to-red colormap behavior,
- the adaptive layout/font sizing,
- dense-grid annotation suppression,
- and the deprecated wrapper.

I’ve attached:
1. a raw-global-scale MEPS-style baseline,
image

2. the updated relative-scale version on the same matrix,
image


3. and a dense-grid readability example.
image

neural_lam/vis.py
```diff
@@ -12,23 +15,154 @@
...
+    cell_h_pt = (fig_height / max(n_rows, 1)) * 72
+    show_annotations = (
+        n_rows * n_cols <= _MAX_CELLS_FOR_ANNOTATIONS
+        and min(cell_w_pt, cell_h_pt) >= _MIN_CELL_SIZE_FOR_ANNOTATIONS
+    )
+
+    return {
+        "fig_width": fig_width,
+        "fig_height": fig_height,
+        "tick_label_size": float(np.clip(15.0 - 0.18 * max_dim, 7.0, 14.0)),
```
sadamov: 7 pt is at the edge of readability for large heatmaps (e.g. 30 vars × 40 steps). Suggest raising the floor to 9 pt.

```suggestion
        "tick_label_size": float(np.clip(15.0 - 0.18 * max_dim, 9.0, 14.0)),
```

neural_lam/vis.py
```diff
@@ -12,23 +15,154 @@
...
+def _get_heatmap_color_values(
+    errors_np: np.ndarray, datastore: BaseRegularGridDatastore
+) -> tuple[np.ndarray, str]:
+    """Normalize heatmap colors to a cross-variable relative scale."""
+    try:
+        ds_state_stats = datastore.get_standardization_dataarray(
+            category="state"
+        )
+    except (AttributeError, KeyError, TypeError, ValueError):
+        return errors_np, "Relative scale"
```
sadamov: When standardization stats are unavailable the fallback returns raw error values, which are on an absolute scale — not a relative one. The label is misleading.

```suggestion
        return errors_np, "Absolute scale"
```

neural_lam/vis.py
```diff
@@ -12,23 +15,154 @@
...
+
+    n_vars = errors_np.shape[0]
+    state_std = _get_feature_scale(ds_state_stats, "state_std", n_vars)
+    if state_std is None:
+        return errors_np, "Relative scale"
```
sadamov: Same issue as above.

```suggestion
        return errors_np, "Absolute scale"
```

neural_lam/vis.py
```diff
@@ -12,23 +15,154 @@
...
+
+    scale = state_std
+    colorbar_label = "Relative scale (state stds)"
+
+    state_diff_std_standardized = _get_feature_scale(
+        ds_state_stats, "state_diff_std_standardized", n_vars
+    )
+    if state_diff_std_standardized is not None:
+        scale = scale * state_diff_std_standardized
+        colorbar_label = "Relative scale (1-step diff stds)"
```
sadamov: The label is correct but i would prefer `error / σ(1-step change)` where σ(1-step change) is `state_std × state_diff_std_standardized` — the std of 1-step differences in physical units, i.e. roughly the error a persistence forecast makes after one step. A value of 1.0 means the model error equals a typical hour-to-hour change; 0.5 means twice as accurate as persistence.

```suggestion
        colorbar_label = "Error / σ(1-step change)"
```

neural_lam/vis.py
```diff
@@ -40,36 +174,7 @@ def plot_on_axis(
...
-        If True, crop the plot to the interior region.
-
-    Returns
-    -------
-    matplotlib.collections.QuadMesh
-        The mesh object created by pcolormesh.
-
-    """
-
+    """Plot weather state on a projection-aware axis using datastore metadata."""
```
sadamov: The original `plot_on_axis` docstring had a full NumPy-style Parameters/Returns section that is used by IDEs and documentation generators. The PR replaced it with a single-line summary unrelated to the PR goals — please restore it.

```suggestion
    """Plot weather state on a projection-aware axis using datastore metadata.

    Parameters
    ----------
    ax : matplotlib.axes.Axes
        The axis to plot on. Should have a cartopy projection.
    da : xarray.DataArray
        The data to plot. Should have shape (N_grid,).
    datastore : BaseRegularGridDatastore
        The datastore containing metadata about the grid.
    vmin : float, optional
        Minimum value for color scale.
    vmax : float, optional
        Maximum value for color scale.
    ax_title : str, optional
        Title for the axis.
    cmap : str or matplotlib.colors.Colormap, optional
        Colormap to use for plotting.
    boundary_alpha : float, optional
        If provided, overlay boundary mask with given alpha transparency.
    crop_to_interior : bool, optional
        If True, crop the plot to the interior region.

    Returns
    -------
    matplotlib.collections.QuadMesh
        The mesh object created by pcolormesh.
    """
```

neural_lam/vis.py
```diff
@@ -155,62 +257,139 @@ def plot_on_axis(
...
+                f"got {color_values_np.T.shape} and {errors_np.T.shape}"
+            )
 
-    ax.imshow(
-        errors_norm,
-        cmap="OrRd",
-        vmin=0,
-        vmax=1.0,
+    if colorbar_label is None:
+        colorbar_label = default_colorbar_label
```
sadamov: `color_values` and `colorbar_label` are never used at the only call site (`ar_model.py`) and add ~20 lines for zero current benefit. Three small removals needed:

**1. Remove the two params from the signature (lines 264–265):**
```python
# delete these two lines
    color_values=None,
    colorbar_label=None,
```

**2. Remove their docstring entries (lines 278–284):**
```python
# delete these seven lines
    color_values : torch.Tensor, optional
        Optional values used only for the background colors. If omitted,
        colors are normalized from ``errors`` using datastore state-variable
        standardization statistics.
    colorbar_label : str, optional
        Optional label for the colorbar. If omitted, an automatic label is
        chosen based on the color normalization used.
```

**3. Replace the if/else block (lines 293–307) with the direct call:**
```python
    color_values_np, colorbar_label = _get_heatmap_color_values(
        errors_np, datastore
    )
```

sadamov (changes requested):
Almost good to go — a few small fixes needed before merge.

neural_lam/vis.py
sadamov: If you prefer `Std` instead of `σ` that's also great :)

kshirajahere:
@sadamov Thanks for the review, I will fix these and get back to you ASAP :)

sadamov:
Make sure to also run pre-commits (and pytest) locally before the next commit :pray:

neural_lam/vis.py
sadamov: Or `(1-step difference)` works, too. Up to you!

kshirajahere:
I reran the plotting tests locally on the latest branch state before pushing, and `tests/test_plotting.py -q` passes on my side. The current push also has all pre-commit jobs green.
@sadamov Ready for a re-review :)

sadamov (commented):
lookin good just a few minor things to fix before merge

neural_lam/vis.py
```diff
@@ -12,23 +15,154 @@
...
+)
 
-def _tex_safe(s: str) -> str:
-    """Escape TeX special characters in s if TeX rendering is currently active.
 
-    Needed because % is a TeX comment character; without escaping it would
-    silently truncate any text that follows it (e.g. the title for r2m (%)).
-    """
+def _tex_safe(s: str) -> str:
+    """Escape TeX special characters in s if TeX rendering is active."""
```
sadamov: One-liner loses the explanation of *why* `%` needs escaping.

```suggestion
    """Escape TeX special characters in s if TeX rendering is currently active.

    Needed because % is a TeX comment character; without escaping it would
    silently truncate any text that follows it (e.g. the title for r2m (%)).
    """
```

neural_lam/vis.py
```diff
@@ -12,23 +15,154 @@
...
+    """Build state-variable labels, padding defensively if metadata is short."""
+    var_names = list(datastore.get_vars_names(category="state"))
+    var_units = list(datastore.get_vars_units(category="state"))
+
+    if len(var_names) < n_vars:
+        var_names.extend(
+            [f"state_feature_{i}" for i in range(len(var_names), n_vars)]
+        )
+    if len(var_units) < n_vars:
+        var_units.extend([""] * (n_vars - len(var_units)))
```
sadamov: `n_vars` always equals `errors.shape[0]` (from the same datastore that provides var names), so the padding block can never trigger. Simplify to:

```suggestion
    var_names = datastore.get_vars_names(category="state")
    var_units = datastore.get_vars_units(category="state")
    return [
        _tex_safe(f"{n} ({u})" if u else n)
        for n, u in zip(var_names, var_units)
    ]
```

neural_lam/vis.py
```diff
@@ -12,23 +15,154 @@
...
+def _get_heatmap_color_values(
+    errors_np: np.ndarray, datastore: BaseRegularGridDatastore
+) -> tuple[np.ndarray, str]:
+    """Normalize heatmap colors to a cross-variable relative scale."""
+    try:
+        ds_state_stats = datastore.get_standardization_dataarray(
+            category="state"
+        )
+    except (AttributeError, KeyError, TypeError, ValueError):
+        return errors_np, "Absolute scale"
```
sadamov: Silent fallback, users won't know why relative scaling disappeared. Add a `warnings.warn` before the return:

```suggestion
    except (AttributeError, KeyError, TypeError, ValueError) as exc:
        warnings.warn(
            f"Could not load standardization stats ({exc}); "
            "falling back to absolute scale.",
            UserWarning,
            stacklevel=3,
        )
        return errors_np, "Absolute scale"
```

neural_lam/vis.py
```diff
@@ -12,23 +15,154 @@
...
+
+    n_vars = errors_np.shape[0]
+    state_std = _get_feature_scale(ds_state_stats, "state_std", n_vars)
+    if state_std is None:
+        return errors_np, "Absolute scale"
```
sadamov: Same silent fallback — add a warning here too:

```suggestion
    if state_std is None:
        warnings.warn(
            "state_std not available; falling back to absolute scale.",
            UserWarning,
            stacklevel=3,
        )
        return errors_np, "Absolute scale"
```

neural_lam/vis.py
```diff
@@ -155,62 +284,117 @@ def plot_on_axis(
 
 
 @matplotlib.rc_context(utils.fractional_plot_bundle(1))
-def plot_error_map(errors, datastore: BaseRegularGridDatastore, title=None):
+def plot_error_heatmap(
```
sadamov, line 352: The figure size set here is immediately overridden by `figsize=` below, so this looks like dead code. The context applies NeurIPS font/text settings - add a one-liner to make that clear:

```suggestion
# rc_context applies NeurIPS font/text settings; figure size is overridden
# by the explicit figsize= below but font family and usetex stay in effect.
@matplotlib.rc_context(utils.fractional_plot_bundle(1))
def plot_error_heatmap(
```

neural_lam/vis.py
```diff
@@ -109,14 +241,12 @@ def plot_on_axis(
     )
 
     if boundary_alpha is not None:
-        # Overlay boundary mask
         mask_da = datastore.boundary_mask
```
sadamov, line 305: Explanatory comment was removed — please restore:

```suggestion
    if boundary_alpha is not None:
        # Overlay boundary mask
        mask_da = datastore.boundary_mask
```

neural_lam/vis.py
```diff
@@ -109,14 +241,12 @@ def plot_on_axis(
...
         mask_values = mask_da.values
         if mask_values.ndim == 2 and mask_values.shape[1] == 1:
             mask_values = mask_values[:, 0]
         mask_2d = mask_values.reshape(grid_shape)
 
-        # Create overlay: 1 where boundary, NaN where interior
         overlay = np.where(mask_2d == 1, 1.0, np.nan)
```
sadamov, line 312:
```suggestion
        # Create overlay: 1 where boundary, NaN where interior
        overlay = np.where(mask_2d == 1, 1.0, np.nan)
```

neural_lam/vis.py
```diff
@@ -129,7 +259,6 @@ def plot_on_axis(
         )
 
     if crop_to_interior:
-        # Calculate extent of interior
         mask_da = datastore.boundary_mask
```
sadamov, line 325:
```suggestion
    if crop_to_interior:
        # Calculate extent of interior
        mask_da = datastore.boundary_mask
```

neural_lam/vis.py
```diff
@@ -285,7 +469,6 @@ def plot_spatial_error(
     colorbar_label: str = "",
 ):
     """Plot spatial error with projection-aware axes."""
-
     error_np = error.detach().cpu().numpy()
```
sadamov: Blank line after docstring was removed - restore it for PEP 257 compliance:

```suggestion
    """Plot spatial error with projection-aware axes."""

    error_np = error.detach().cpu().numpy()
```

tests/test_plotting.py
```diff
@@ -22,6 +23,8 @@
 from tests.conftest import init_datastore_example
 from tests.dummy_datastore import DummyDatastore
 
+plt.switch_backend("Agg")  # non-interactive backend for headless test runs
```
sadamov: Module-level `plt.switch_backend` runs on import and can affect other test modules in the same pytest session. Move to a session-scoped autouse fixture:

```suggestion
@pytest.fixture(scope="session", autouse=True)
def _set_agg_backend():
    """Use non-interactive backend for all plotting tests."""
    matplotlib.use("Agg")
```

tests/test_plotting.py
```diff
@@ -123,6 +171,128 @@ def test_plot_error_map() -> None:
     assert actual_y_ticklabels == expected_y_ticklabels
 
     assert len(ax.texts) == pred_steps * d_f
+    assert colorbar_ax.get_ylabel() == "Error / Std(1-step change)"
```
sadamov: `"Error / Std(1-step change)"` couples the test to `dummydata` exposing `state_diff_std_standardized`. Weaken the assertion to decouple it:

```suggestion
    assert colorbar_ax.get_ylabel() != ""
```

tests/test_plotting.py
```diff
@@ -123,6 +171,128 @@ def test_plot_error_map() -> None:
...
+
+    fig = vis.plot_error_heatmap(errors, datastore=datastore)
+    colorbar = fig.axes[1]
+
+    assert colorbar.get_ylabel() == "Absolute scale"
+
+    plt.close(fig)
+
+
+def test_plot_error_heatmap_uses_white_to_red_colormap():
```
sadamov: This tests the value of `_HEATMAP_CMAP` (a module constant), not observable behavior. If the colormap is intentionally changed, the test fails by design. Remove it.

tests/test_plotting.py
```diff
@@ -123,6 +171,128 @@ def test_plot_error_map() -> None:
...
+    ax = fig.axes[0]
+
+    assert len(ax.texts) == 0
+    assert fig.get_size_inches()[0] > 18.0
+    assert ax.get_xticklabels()[0].get_fontsize() == pytest.approx(9.0)
+
+    plt.close(fig)
+
+
+def test_plot_error_map_deprecated_wrapper():
```
sadamov: `test_plot_error_map` already checks `pytest.warns(DeprecationWarning)` and `isinstance(fig, plt.Figure)`. This wrapper test is fully redundant — remove it.

tests/test_plotting.py
```diff
@@ -123,6 +171,128 @@ def test_plot_error_map() -> None:
...
+        < small_ax.get_yticklabels()[0].get_fontsize()
+    )
+    assert large_ax.texts[0].get_fontsize() < small_ax.texts[0].get_fontsize()
+    assert large_ax.get_xticklabels()[0].get_rotation() == 45.0
+
+    plt.close(small_fig)
+    plt.close(large_fig)
+
+
+def test_plot_error_heatmap_skips_annotations_for_very_dense_grids():
```
sadamov: `test_plot_error_heatmap_adapts_figure_and_font_sizes` and `test_plot_error_heatmap_skips_annotations_for_very_dense_grids` share the same setup and exercise the same `_compute_heatmap_layout` path. Merge into one:

```suggestion
def test_plot_error_heatmap_adapts_layout_for_grid_size():
    """Dense heatmaps get larger figures, smaller fonts, and suppress annotations."""
    small_fig = vis.plot_error_heatmap(torch.ones((4, 5)), datastore=HeatmapDatastore(n_vars=5))
    large_fig = vis.plot_error_heatmap(torch.ones((20, 30)), datastore=HeatmapDatastore(n_vars=30))
    dense_fig = vis.plot_error_heatmap(torch.ones((40, 50)), datastore=HeatmapDatastore(n_vars=50))

    assert large_fig.get_size_inches()[0] > small_fig.get_size_inches()[0]
    assert large_fig.axes[0].get_yticklabels()[0].get_fontsize() < small_fig.axes[0].get_yticklabels()[0].get_fontsize()
    assert large_fig.axes[0].get_xticklabels()[0].get_rotation() == 45.0
    assert len(dense_fig.axes[0].texts) == 0
    assert dense_fig.get_size_inches()[0] > 18.0

    plt.close(small_fig); plt.close(large_fig); plt.close(dense_fig)
```

tests/test_plotting.py
```diff
@@ -123,6 +171,128 @@ def test_plot_error_map() -> None:
...
+
+    errors = torch.tensor([[1.0, 2.0], [3.0, 4.0]])
+    datastore = NoStatsHeatmapDatastore(n_vars=errors.shape[1])
+
+    fig = vis.plot_error_heatmap(errors, datastore=datastore)
+    colorbar = fig.axes[1]
+
+    assert colorbar.get_ylabel() == "Absolute scale"
+
+    plt.close(fig)
```
sadamov, line 261: The branch where `state_std` exists but `state_diff_std_standardized` is absent (`colorbar_label = "Relative scale (state stds)"`) has no test. Add one after this test:

```suggestion
    plt.close(fig)


def test_plot_error_heatmap_uses_state_std_only_when_diff_std_absent():
    class StateStdOnlyDatastore(HeatmapDatastore):
        def get_standardization_dataarray(self, category):
            return super().get_standardization_dataarray(category).drop_vars("state_diff_std_standardized")

    errors = torch.tensor([[2.0, 4.0], [1.0, 3.0]])
    fig = vis.plot_error_heatmap(errors, datastore=StateStdOnlyDatastore(n_vars=2, state_std=[2.0, 1.0]))
    assert fig.axes[1].get_ylabel() == "Relative scale (state stds)"
    plt.close(fig)
```

tests/test_plotting.py
```diff
@@ -194,15 +362,12 @@ def test_plot_spatial_error_crop_to_interior_changes_extent() -> None:
 
 @pytest.fixture
 def model_and_batch(tmp_path, time_step, time_unit):
-    """Setup a model and dataset for testing plot_examples"""
-    # Create timedelta with specified step length
+    """Setup a model and dataset for testing plot_examples."""
```
sadamov, line 406: Step-by-step comments were removed from `model_and_batch` (`# Create timedelta with specified step length`, `# Create datastore with specified step_length`, `# Create minimal model args`, `# Create config`, `# Create model`, `# Create dataset to get a sample batch`, `# Get a batch (just use one sample)`). These help readers follow the multi-step setup — please restore them.

tests/test_plotting.py
```diff
@@ -272,33 +433,27 @@ class ModelArgs:
 def test_plot_examples_integration_saves_figure(
     model_and_batch, time_step, time_unit, t_i
 ):
-    """Integration test that saves actual figure for manual inspection"""
+    """Integration test that saves an actual figure for manual inspection."""
     model, batch, datastore, tmp_path = model_and_batch
 
-    # Reset plotted examples counter
     model.plotted_examples = 0
```
sadamov, line 518: Step-by-step comments were also removed from `test_plot_examples_integration_saves_figure` (`# Reset plotted examples counter`, `# Verify that the model correctly inferred time step from datastore`, `# Generate prediction`, `# Rescale to original data scale`, `# Get first example`, `# Create DataArrays`, `# Get vranges`, `# Create plot for specified timestep and first variable`, `# Save for inspection`, `# Verify the figure was created`). Please restore them.

CHANGELOG.md
```diff
@@ -30,6 +30,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replace `shell=True` subprocess call in `compute_standardization_stats.py` with a safe argument list and Python-side hostname parsing to prevent command injection via `SLURM_JOB_NODELIST` [\#264](https://github.com/mllam/neural-lam/pull/264) @ashum9
 
 - Avoid NaN when standardizing fields with zero std [#189](https://github.com/mllam/neural-lam/pull/189) @varunsiravuri
+- Improve metric heatmaps by renaming `plot_error_map`, using a relative
+  cross-variable color scale while keeping original values in annotations, and
+  scaling figure size, tick labels, and annotation text with the number of
+  variables and lead times so larger evaluation outputs remain readable
+  ([#375](https://github.com/mllam/neural-lam/issues/375))
```
sadamov: Cross-variable normalization changes how every existing heatmap looks — that's a `Changed` item, not a bug fix. Move this entry under `### Changed`:

```suggestion
- Change metric heatmap (`plot_error_map`, now `plot_error_heatmap`) to use a
  shared cross-variable color scale instead of per-row normalization, add a
  colorbar, and scale figure size and font sizes with grid dimensions
  ([#375](https://github.com/mllam/neural-lam/issues/375))
```

neural_lam/vis.py
kshirajahere: Restored the fuller `_tex_safe` docstring explaining why `%` needs escaping under TeX rendering.

neural_lam/vis.py
kshirajahere: Added a `UserWarning` when standardization stats cannot be loaded, so the fallback to absolute scale is explicit rather than silent.

neural_lam/vis.py
kshirajahere: Also added a warning for the branch where `state_std` itself is unavailable before falling back to absolute scale.

CHANGELOG.md
kshirajahere: Moved the changelog entry under `### Changed` and updated the wording to match the new heatmap behavior.

tests/test_plotting.py
kshirajahere: I have restored the step-by-step comments in both `model_and_batch` and the plotting integration test.

kshirajahere:
Hey @sadamov addressed latest review changes, ready for review.

kshirajahere:
Also ran all tests locally before

sadamov (approved):
I really like these new maps! they will prove very useful. This PR is proposed for v0.7.0 and will be discussed at the next dev meeting and is marked as ready. @kshirajahere thanks for working on this over the course of the past weeks!

sadamov:
@kshirajahere I saw you added the docstrings asked for during the dev meeting. great! now we wait for #208

sadamov:
Hi @kshirajahere,

After the dev-meeting I took another look at some realistic heatmaps using the COSMO checkpoint from the boundary paper. And as @joeloskarsson mentioned, the heatmaps are really dominated by a few variables. Also, I don't think that the three-level silent fallback can be a bit dangeours as the user might not get the heatmap they expected. Because we already completed the PR review I implemented the changes here for you, not to cause you more work. But I would appreciate your review and if you agree merge with #376. Here are four plots for four different colormaps based on different scalings/normalizations (the skill one is not part of the PR, because we said we want only red tones :red_circle: ) 

[cosmo_skill_heatmap.pdf (omitted from PR)](https://github.com/user-attachments/files/26836372/cosmo_skill_heatmap.pdf)
[cosmo_per_var_max_heatmap.pdf (legacy fallback)](https://github.com/user-attachments/files/26836373/cosmo_per_var_max_heatmap.pdf)
[cosmo_one_step_std_heatmap.pdf (optional via new argument)](https://github.com/user-attachments/files/26836375/cosmo_one_step_std_heatmap.pdf)
[cosmo_global_state_std_heatmap.pdf (default)](https://github.com/user-attachments/files/26836376/cosmo_global_std_heatmap.pdf)

In this PR here we had one step diff as the default. But as you can see this is basically dominated by the pressure varaibles that change only very slightly in one forcast step. Sea Level pressure is actually a very important variable to assess model skill also in traditional NWP models. It contains the information about the whole atmospheric column and therefore about the layering and horizontal gradients -> wind.
But, I would argue that just seeing "yes pressure is important" is not what these score cards are used for (I am exaggerating a bit). Rather we would like to see a somewhat useful comparison between variables. And the the global state diff allows for that. We actually see the typical dynamic (i.e. fast changing) variables to be the most red, like wind, vertical motion, humidity.

If you agree with my logic here and the PR into your branch, feel free to merge. Or let me know if you see some other reason why my default is bad. https://github.com/kshirajahere/neural-lam/pull/1

kshirajahere:
Hey @sadamov went through the stacked PR. It looks good from what I have reviewed. (just few questions before merging, have raised them on https://github.com/kshirajahere/neural-lam/pull/1) 

This feels like a cleaner match to the dev-meeting discussion and should make the heatmap behavior easier to reason about from the API and the colorbar label. Thanks a lot ❤️

kshirajahere:
Merged https://github.com/kshirajahere/neural-lam/pull/1 into this PR, Thanks @sadamov :D

sadamov:
#208 was just merged, could you resolve conflicts here @kshirajahere so that we can merge the PR?

kshirajahere:
On it @sadamov

kshirajahere:
@sadamov pls check

sadamov:
@kshirajahere we did it! :partying_face: lovely new heatmaps are in main

kshirajahere:
Yay 🎉🍾 cheers!

PR #405, opened 2026-03-16
docs: fix invalid pip install command in README

Riteesh-NITT:
Hello Guys, I noticed a small typo error in the Using pip section of the README file. The instructions included says --group flag (python -m pip install --group dev -e.), Which is specific to uv and causes pip to throw a unrecognized option error. I've updated it to use the standard pip install -e ".[dev]" syntax so future contributors don't hit an installation crash.

Looking forward to contributing more to the project!


sadamov:
pip install --group is a real, standard pip flag, not uv-specific. It was added in pip 25.1 (April 2025) implementing [PEP 735 dependency groups](https://peps.python.org/pep-0735/).
probably just need to update your pip version

PR #239, opened 2026-02-23
Normalize data on GPU (#25)

Sharkyii:
## Describe your changes

- Move data normalization from CPU to GPU
- Add on_after_batch_transfer hook for GPU-side normalization
- Remove CPU normalization from Dataset

## Issue Link
solves #25

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [ ] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented (including return types and parameters)
- [ ] the code is easy to maintain

## Author checklist after completed review

- [ ] I have added a line to the CHANGELOG describing this change, in a section
  reflecting type of change (add section where missing):
  - *added*: when you have added new functionality
  - *changed*: when default behaviour of the code has been changed
  - *fixes*: when your contribution fixes a bug
  - *maintenance*: when your contribution is relates to repo maintenance, e.g. CI/CD or documentation

## Checklist for assignee

- [ ] PR is up to date with the base branch
- [ ] the tests pass
- [ ] (if the PR is not just maintenance/bugfix) the PR is assigned to the next milestone. If it is not, propose it for a future milestone.
- [ ] author has added an entry to the changelog (and designated the change as *added*, *changed*, *fixed* or *maintenance*)
[...]

PR #411, opened 2026-03-17
Fix IndexError and silent tensor-shape corruption in distributed standardization stats

### LINKED ISSUE #412: [Bug] IndexError in distributed flux stats gathering in compute_standardization_stats.py
## Description

Running `compute_standardization_stats.py --distributed` with `world_size > 1` crashes with an `IndexError` in the rank-0 block that aggregates 
per-rank flux statistics.

## Reproduction

Use `--distributed` with any realistic dataset (N samples >> world_size).

## Error

```
IndexError: index N is out of bounds for dimension 0 with size world_size
```

## Root cause

`dist.all_gather_object(flux_means_gathered, flux_means)` produces a `list[list[0-dim tensor]]` of shape `(world_size, n_batches_per_rank)`.
The original code converts this to a 2-D tensor:

```python
flux_means_gathered = torch.tensor(flux_means_gathered)  # (world_size, n_batches)
flux_means = [flux_means_gathered[i] for i in original_indices]  # crashes
```

`original_indices` contains one entry per *dataset sample* (can be thousands), but the tensor's first dimension is `world_size` (typically 4–8).
Indexing with `i >= world_size` immediately raises `IndexError`.

## Fix

Flatten all per-batch scalars across ranks without sample-level depadding:

```python
flux_means = [
    torch.cat([torch.stack(rank_flux) for rank_flux in flux_means_gathered])
]
```

Fixed in PR #411.

### LINKED ISSUE #413: [Bug] Diff stats silently corrupted by wrong tensor shape in distributed compute_standardization_stats.py
## Description

Running `compute_standardization_stats.py --distributed` with `world_size > 1` silently writes incorrect `diff_mean.pt` and `diff_std.pt` files with no
exception raised.

## Root cause

After gathering, `diff_means_gathered` has shape `(total_window_samples, d_features)`. The original code uses a Python list comprehension to select rows:

```python
diff_means = [diff_means_gathered[i] for i in original_indices]
```

Each `diff_means_gathered[i]` is a 1-D `(d_features,)` tensor (a single row). `torch.cat(dim=0)` on a list of 1-D tensors concatenates along the *only*
dimension, producing a flat `(N * d_features,)` tensor instead of the expected `(N, d_features)` 2-D tensor.

`save_stats` then calls `torch.mean(tensor, dim=0)` on this 1-D tensor, which returns a *scalar* mean of all values instead of the `(d_features,)`
per-variable mean. The incorrect scalar is written to disk without any error.

## Impact

All per-variable diff statistics (`diff_mean`, `diff_std`) computed in distributed mode are wrong — they collapse to a single scalar that is
broadcast as the mean/std for every variable.

## Fix

Use PyTorch fancy indexing to preserve both dimensions:

```python
diff_means = [diff_means_gathered[original_indices]]  # (N, d_features)
```

`torch.cat` on a single `(N, d_features)` tensor is a no-op on `dim=0`,
giving the correct `(N, d_features)` shape for `save_stats`.

Fixed in PR #411.

RajdeepKushwaha5:
## Describe your changes

Fix two bugs in `compute_standardization_stats.py` that only manifest when running with `--distributed` and `world_size > 1`:

1. **Flux stats (`IndexError`):** `torch.tensor(flux_means_gathered)` creates a `(world_size, n_batches_per_rank)` tensor, but `original_indices` can have length greater than `world_size`, causing an `IndexError`. Fixed by concatenating per-rank batch lists with `torch.cat`/`torch.stack` — flux stats are per-batch scalars and do not need sample-level depadding.

2. **Diff stats (silent shape corruption):** `.view(-1, *diff_means[0].shape)` followed by a list comprehension `[gathered[i] for i in original_indices]` drops the feature dimension, so `torch.cat` produces a flat `(N*d_features,)` tensor instead of `(N, d_features)`. Fixed by removing `.view()` and using tensor fancy-indexing (`gathered[original_indices]`) which preserves the 2-D shape.

Added 14 unit tests reproducing both bugs and verifying the fixes.

## Issue Link

Closes #412
Closes #413

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass


sadamov:
Hi @RajdeepKushwaha5 and thanks for your contributions here! Could you quickly check if #239 would solve your problem here as well?

RajdeepKushwaha5:
> Hi @RajdeepKushwaha5 and thanks for your contributions here! Could you quickly check if #239 would solve your problem here as well?

Thanks for the pointer, @sadamov! I did a thorough comparison of both PRs.

They are completely independent — both in purpose and in the lines they touch in `compute_standardization_stats.py`:

**What PR #239 does in `compute_standardization_stats.py`:**
- Adds a `StandardizedDatastore` wrapper class at the top of the file
- Replaces `datastore=datastore, standardize=True` with `datastore=StandardizedDatastore(datastore)` in the second `WeatherDataset(...)` call

(And notably, based on your latest review comments, you've asked for that `StandardizedDatastore` change to be reverted — so after #239's revision cycle, it may not touch `compute_standardization_stats.py` at all.)

**What PR #411 does in `compute_standardization_stats.py`:**
- Fixes the distributed rank-0 **flux stats aggregation block** (~lines 248–270): removes `torch.tensor(flux_means_gathered)` which created a `(world_size, n_batches)` tensor that was then indexed by sample-level `original_indices`, crashing with `IndexError` on any realistic dataset
- Fixes the distributed rank-0 **diff stats aggregation block** (~lines 368–385): removes the `.view()` that made the gathered tensor 3-D and replaces the list comprehension with tensor fancy-indexing to preserve the `(N, d_features)` shape, preventing silent stat corruption

These code paths are not touched anywhere in #239 — not in the current version, not in the revised version after your review. Even if #239 merged today in its current state, both bugs fixed in #411 would still crash or silently corrupt output on any multi-GPU `--distributed` run.

The two PRs can coexist and merge independently. If #239 lands first I'm happy to rebase.

joeloskarsson:
Hi, please update to use our PR template. Is there an issue tracking this problem?

RajdeepKushwaha5:
> Hi, please update to use our PR template. Is there an issue tracking this problem?

Hello Sir @joeloskarsson, thanks for the heads-up! I've updated the PR description to follow the template.

The two issues tracking these bugs are #412 and #413:
- **#412** — IndexError in distributed flux stats gathering
- **#413** — Diff stats silently corrupted by wrong tensor shape in distributed mode

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -246,20 +246,31 @@ def main(
...
+                )
+            ]
+            flux_squares = [
+                torch.cat(
+                    [
+                        torch.stack(rank_flux)
+                        for rank_flux in flux_squares_gathered
+                    ]
+                )
+            ]
```
sadamov, line 264: Flux stats are now not depadded at all, while `means`/`squares` are still filtered by `original_indices` three lines above. When `PaddedWeatherDataset` adds padding, the padded batches contribute duplicate values to the flux mean, making `flux_stats.pt` world-size-dependent. The bias is small (at most `world_size - 1` near-duplicate entries) but the inconsistency is worth fixing.

```suggestion
            n_real_per_rank = len(ds.get_original_indices()) // world_size
            flux_means = [
                torch.cat(
                    [
                        torch.stack(rank_flux[:n_real_per_rank])
                        for rank_flux in flux_means_gathered
                    ]
                )
            ]
            flux_squares = [
                torch.cat(
                    [
                        torch.stack(rank_flux[:n_real_per_rank])
                        for rank_flux in flux_squares_gathered
                    ]
                )
            ]
```

sadamov (changes requested):
Two real bugs, both correctly identified, thanks! The bugs only fire when someone explicitly passes --distributed on a multi-node SLURM job (SLURM_NTASKS > 1).  The fixes eliminate the crash and the shape corruption. There are two correctness concerns worth addressing before merge.

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -355,21 +366,16 @@ def main(
...
             )
-            diff_means, diff_squares = (
-                [diff_means_gathered[i] for i in original_indices],
-                [diff_squares_gathered[i] for i in original_indices],
-            )
+            # Fancy-index to keep the 2-D shape (N, d_features)
+            # instead of the list-comprehension that collapsed
+            # each row to 1-D.
+            diff_means = [diff_means_gathered[original_indices]]
+            diff_squares = [diff_squares_gathered[original_indices]]
```
sadamov: `get_original_window_indices` returns `[i // step_int for i in range(N_total * step_int)]`, which for `step_int=3` gives `[0, 0, 0, 1, 1, 1, ...]`. With the new fancy-indexing, `diff_means_gathered[original_indices]` repeats each source row `step_int` times instead of selecting the first `N_total * step_int` distinct rows. For `step_int=1` (the common case) this is fine, but for larger step lengths the diff stats will be wrong.

The correct index is simply a contiguous slice, since padded rows are at the end of the gathered tensor:

```suggestion
            n_original_windows = len(ds_standard.get_original_indices()) * step_int
            diff_means = [diff_means_gathered[:n_original_windows]]
            diff_squares = [diff_squares_gathered[:n_original_windows]]
```

This is simpler and correct for all `step_int`. `get_original_window_indices` can then be removed if nothing else uses it.

tests/test_compute_standardization_stats.py
```diff
@@ -0,0 +1,158 @@
...
+        result = data[indices]
+        assert torch.mean(result, dim=0).shape == (d_f,)
+
+    def test_fix_preserves_values(self):
+        """Fancy-index selects correct rows."""
+        d_f, total = 5, 10
+        data = torch.arange(total * d_f, dtype=torch.float32).view(total, d_f)
+        indices = [0, 2, 4, 6]
+
+        assert torch.equal(data[indices], data[[0, 2, 4, 6]])
```
sadamov: This test is a tautology — both sides of the assertion index `data` with `[0, 2, 4, 6]` so it can never fail. Should compare against explicitly constructed expected values.

```suggestion
    def test_fix_preserves_values(self):
        """Fancy-index selects correct rows."""
        d_f, total = 5, 10
        data = torch.arange(total * d_f, dtype=torch.float32).view(total, d_f)
        indices = [0, 2, 4, 6]
        result = data[indices]
        expected = torch.stack([data[0], data[2], data[4], data[6]])
        assert torch.equal(result, expected)
```

sadamov:
please fix the changelog and then we should be ready to merge soon

sadamov (commented):
Thanks @RajdeepKushwaha5  ready to merge with one small change. Both bugs are fixed correctly enough that the distributed flux/diff stats paths produce sensible output. The one residual issue is a unit mismatch in the flux slice (samples vs. batches) that is a silent no-op at default `batch_size`, but would actively drop real data at `batch_size=1`. Easy to remove.

sadamov:
Pushed 2f51ea8 directly to drop the `n_real_per_rank` slice — sorry for the churn, my earlier suggestion was wrong. Nothing else needed from your side, this looks ready to merge.

sadamov (approved):
looks good. cleaned up tests a little bit

sadamov:
@RajdeepKushwaha5 thanks alot!!

Issue #278, opened 2026-02-27
Replace assert with if ... raise ValueError(...) for runtime validation


Sir-Sloth-The-Lazy:
# Replace `assert` with `if ... raise ValueError(...)` for runtime validation

## Problem

Multiple `assert` statements are used for input validation that should be enforced at runtime. Python's `assert` is disabled when running with optimization flags (`python -O`), which would **silently skip these checks**. These should use `if ... raise ValueError(...)` instead to ensure validation is always active.

## Affected Files & Locations

| File | Line(s) | Context |
|------|---------|---------|
| [neural_lam/train_model.py](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/train_model.py) | L229–L233 | Validation of `val_steps_to_log` range |
| [neural_lam/train_model.py](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/train_model.py) | L237–L241 | Validation of `var_leads_metrics_watch` range |
| [neural_lam/utils.py](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/utils.py) | L132–L137 | Consistency checks in [load_graph](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/utils.py#47-198) (level counts) |
| [neural_lam/utils.py](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/utils.py) | L213 | MLP blueprint validation in [make_mlp](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/utils.py#200-225) |
| [neural_lam/create_graph.py](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/create_graph.py) | L596–L598 | Config path assertion in [cli()](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/create_graph.py#562-609) |
| [neural_lam/interaction_net.py](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/interaction_net.py) | L48 | Aggregation method validation in `InteractionNet.__init__` |
| [neural_lam/interaction_net.py](file:///Users/sirsloththelazy/Projects/GSOC/neural-lam/neural_lam/interaction_net.py) | L143–L145 | Chunk size validation in `SplitMLPs.__init__` |

### Before (assert — stripped by `python -O`)

```python
# train_model.py
[...]

PR #193, opened 2026-02-20
Fix single-GPU testing bottleneck, tensor shape error, and DDP/NCCL Windows compatibility

AdityaKumarSethia:
## Describe your changes

**Summary of the changes:**
- **`tests/test_training.py`**: Removed the hardcoded `devices=2` in the `pl.Trainer` initialization. The device count is now dynamically assigned using `devices=2 if torch.cuda.device_count() >= 2 else 1`.
- **`neural_lam/vis.py`**: Added a dimension check in `plot_error_map` to gracefully handle 1D `(pred_steps,)` tensors generated during single-device testing. Tensors are explicitly unsqueezed to 2D before transposition.

**Relevant motivation and context:**
Local integration testing on single-GPU developer machines was failing with a `MisconfigurationException` because the trainer strictly expected 2 GPUs. Modifying this to 1 GPU revealed an unpacking error (`ValueError: not enough values to unpack`) in `vis.py`, because metrics from a single device are not stacked by PyTorch Lightning's `all_gather` and remain a 1D tensor. Un-squeezing the tensor resolves the crash and also fixes a PyTorch `.T` deprecation warning.

**Dependencies:**
None.

## Issue Link

Closes #192

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [x] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

[...]

PR #422, opened 2026-03-17
Guard output_std against incompatible losses (Fixes #363)

### LINKED ISSUE #363: [BUG] Incompatible objective: --output_std with wmse or wmae creates degenerate learned-variance training
## Summary

When `--output_std` is enabled together with `--loss wmse` or `--loss wmae`, the model can minimize the loss by inflating `pred_std` rather than improving the forecast itself.
This makes the learned variance head degenerate: `pred_std` is incentivized to grow without bound, while the objective approaches zero trivially. This is a direct issue for probabilistic forecasting work, since the codebase currently allows this combination without warning or validation.

## Why I believe this is a bug

With `--output_std`, the model predicts a standard deviation head:

- `pred_std = softplus(raw)` in [neural_lam/models/base_graph_model.py](neural_lam/models/base_graph_model.py#L348)

The weighted losses then use that `pred_std` as a divisor:

- `wmse`: [neural_lam/metrics.py](neural_lam/metrics.py#L56)
- `wmae`: [neural_lam/metrics.py](neural_lam/metrics.py#L111)

**`wmse`** computes:

$$\frac{(y - \hat{y})^2}{\sigma^2}$$

**`wmae`** computes:

$$\frac{|y - \hat{y}|}{\sigma}$$

For fixed residuals, both objectives are monotone decreasing in $\sigma$.

**For WMSE:**

$$\frac{\partial}{\partial \sigma}\left(\frac{(y-\hat{y})^2}{\sigma^2}\right) = -\frac{2(y-\hat{y})^2}{\sigma^3} \leq 0$$

**For WMAE:**

$$\frac{\partial}{\partial \sigma}\left(\frac{|y-\hat{y}|}{\sigma}\right) = -\frac{|y-\hat{y}|}{\sigma^2} \leq 0$$

So the variance head is always rewarded for increasing `pred_std`. There is no balancing term such as $\log(\sigma)$.
By contrast, `nll` includes a $\log(\sigma)$ term and has a proper optimum, and `crps_gauss` also has non-degenerate dependence on `pred_std`.

## Current code path

The problematic combination is currently allowed by default CLI behavior:
- `--output_std` is enabled via [neural_lam/train_model.py](neural_lam/train_model.py#L116)
- the default loss is `wmse` in [neural_lam/train_model.py](neural_lam/train_model.py#L132)
- no compatibility validation appears in [neural_lam/models/ar_model.py](neural_lam/models/ar_model.py#L95)

This means a user can enable `--output_std` while leaving the default loss unchanged and get a degenerate probabilistic objective without any warning.

## Reproduction of Bug

A simple optimization that updates only the variance head shows the issue clearly.

Using fixed residuals and optimizing only shared per-variable `pred_std`:
- with `wmse`, the learned `pred_std` keeps increasing and the loss decreases trivially
- with `nll`, the learned `pred_std` converges near the empirical residual scale

Observed behavior from a local reproduction:

| Loss | Step 0 `pred_std` | Step 5000 `pred_std` | True residual std | Behavior |
|------|-------------------|----------------------|-------------------|----------|
| WMSE | `[0.69, 0.69, 0.69]` | `[7.16, 7.16, 7.16]` | `[1.04, 1.96, 0.50]` | diverges |
| NLL  | `[0.69, 0.69, 0.69]` | `[1.04, 1.96, 0.50]` | `[1.04, 1.96, 0.50]` | converges |

The important point is not the exact numbers, but that the `wmse`/`wmae` objective gives the variance head a one-sided incentive to grow.

## Expected behavior

If `output_std=True`, the loss should be compatible with a learned predictive standard deviation.

At minimum, the code should reject or warn on combinations where `pred_std` is used in a way that makes the objective degenerate.

---

## Actual behavior

The code currently allows:

- `--output_std --loss wmse`
- `--output_std --loss wmae`

even though these objectives encourage unbounded variance inflation.

## Existing test coverage gap

There appears to be no test that validates `output_std=True` with a proper probabilistic loss.

The current output-std training test uses `loss = "mse"` in [tests/test_training.py](tests/test_training.py#L67), which ignores `pred_std` entirely and therefore does not exercise the learned-variance behavior.

## Suggested fix

At minimum:

1. Add a validation guard that rejects `output_std=True` with `loss in {"wmse", "wmae"}`.
2. Allow `output_std=True` only with losses that properly regularize the variance head, such as `nll` and likely `crps_gauss`.
3. Add training coverage for valid probabilistic configurations.

Possible locations for the guard:

- [neural_lam/models/ar_model.py](neural_lam/models/ar_model.py)
- or [neural_lam/train_model.py](neural_lam/train_model.py)

## Possible follow-up tests

- Assert that `output_std=True` with `wmse` raises a clear error
- Assert that `output_std=True` with `wmae` raises a clear error
- Add at least one training smoke test for `output_std=True` with `nll`
- Optionally add one for `crps_gauss`

---

## Affected files

- [neural_lam/metrics.py](neural_lam/metrics.py)
- [neural_lam/models/base_graph_model.py](neural_lam/models/base_graph_model.py)
- [neural_lam/models/ar_model.py](neural_lam/models/ar_model.py)
- [neural_lam/train_model.py](neural_lam/train_model.py)
- [tests/test_training.py](tests/test_training.py)

kshirajahere:
## Describe your changes

- Add explicit loss metadata in `metrics.py` so support for learned predictive std-dev is defined per loss.
- Reject incompatible `--output_std` and loss combinations early in `train_model.py` with a clear CLI error.
- Keep the same compatibility guard in `ARModel` so direct model construction is validated too.
- Add focused regression tests for:
  - CLI rejection of incompatible `--output_std` losses,
  - training/model construction with valid probabilistic losses.
- Add CHANGELOG entry under `## [unreleased] -> ### Fixed`.

## Issue Link

Fixes #363

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass


neural_lam/models/ar_model.py
```diff
@@ -123,6 +123,17 @@ def __init__(
...
+            args.loss
+        ):
+            supported_losses = ", ".join(
+                metrics.get_output_std_compatible_metrics()
+            )
+            raise ValueError(
+                "--output_std requires a loss that can train the predictive "
+                f"std-dev head, but got loss='{args.loss}'. Supported "
+                f"losses: {supported_losses}."
+            )
```
joeloskarsson: I would reword this to make it more model-agnostic (`pred_std` could be produced in different ways).


```suggestion
            raise ValueError(
                "Training with --output_std requires a loss that incorporates predicted "
                f"std-dev. The current choice of {args.loss} does not. Supported "
                f"losses: {supported_losses}."
            )
```

neural_lam/metrics.py
```diff
@@ -228,10 +212,45 @@ def crps_gauss(
...
+    """
+    metric_name_lower = metric_name.lower()
+    assert is_defined_metric(metric_name), f"Unknown metric: {metric_name}"
+    return DEFINED_METRICS[metric_name_lower]["fn"]
+
+
+def metric_supports_output_std(metric_name):
+    """Return whether the metric can train a learned predictive std-dev."""
+    metric_name_lower = metric_name.lower()
+    assert is_defined_metric(metric_name), f"Unknown metric: {metric_name}"
```
joeloskarsson: In line with https://github.com/mllam/neural-lam/issues/278 it's probably better to use ValueErrors instead of asserts here.

neural_lam/train_model.py
```diff
@@ -242,6 +264,12 @@ def main(input_args=None):
         int(k): v for k, v in json.loads(args.var_leads_metrics_watch).items()
     }
 
+    compatibility_error = get_loss_output_std_compatibility_error(
+        args.loss, args.output_std
+    )
+    if compatibility_error is not None:
+        parser.error(compatibility_error)
+
```
joeloskarsson: Seems like we are not checking this in two places, which seems unnecessary. I would be fine with just checking it here in train_model.py

tests/test_training.py
```diff
@@ -44,7 +49,9 @@ def run_simple_training(datastore, set_output_std):
         # XXX: `devices` has to be set to 2 otherwise
         # neural_lam.models.ar_model.ARModel.aggregate_and_plot_metrics fails
         # because it expects to aggregate over multiple devices
-        devices=2,
+        devices=devices,
```
joeloskarsson: I think making changes to devices here will just lead to conflicts with https://github.com/mllam/neural-lam/pull/193

joeloskarsson (changes requested):
Thanks for the work on this, I like the approach! Left some comments here

neural_lam/models/ar_model.py
kshirajahere: Addressed by removing the duplicate check from `ARModel` entirely and keeping the validation only in `train_model.py`, per your other comment below. That also avoids having to maintain two error-message variants.

neural_lam/metrics.py
kshirajahere: Changed this to raise `ValueError` instead of using `assert` for unknown metric names.

neural_lam/train_model.py
kshirajahere: Fair, the compatibility check now lives only in `train_model.py`, and I removed the duplicate validation from `ARModel`.

tests/test_training.py
kshirajahere: Reverted the trainer/device churn here and reduced the diff back toward `main`. The remaining semantic change is just that the `output_std` training path now uses `loss="nll"`.

kshirajahere:
Resolved the merge conflict with the latest main and reran CI; all checks are now green. The earlier review comments should all be addressed, so this should be ready for another look whenever you have time. @joeloskarsson

tests/test_training.py
joeloskarsson: Still some differences here, but they might be disappear once this is back up to date with main.

neural_lam/train_model.py
```diff
@@ -140,7 +159,10 @@ def main(input_args=None):
         "--loss",
         type=str,
         default="wmse",
-        help="Loss function to use, see metric.py",
+        help=(
+            "Loss function to use, see metric.py. When --output_std is set, "
+            "only losses that support learned predictive std-dev are valid."
+        ),
     )
```
joeloskarsson, line 206: Add the allowed loss options as explicit argparse choices here, that way the user can't choose something incompatible.

neural_lam/train_model.py
```diff
@@ -24,6 +24,25 @@
...
+
+    if metrics.metric_supports_output_std(loss):
+        return None
+
+    supported_losses = ", ".join(metrics.get_output_std_compatible_metrics())
+    return (
+        "Training with --output_std requires a loss that incorporates "
+        f"predicted std-dev. The current choice of --loss '{loss}' does not. "
+        f"Supported losses: {supported_losses}."
+    )
```
joeloskarsson: Not sure if this function is needed. If we have explicit options for losses, then we can assume that the metrics is defined, and the check is just e.g. `if output_std and output_std_not_supported`. The error message is great though :)

joeloskarsson (changes requested):
Just had another look at this now. This taking so long has meant that there are a lot of changes to main that has caused conflicts. Happy to keep reviewing this though if you are ok with getting it back up to date with main.

kshirajahere:
Hi @joeloskarsson 😁, Yes sorry for the inactivity, caught up in exams. I will update my PR soon upto main and take into view ur recent review. Thank you so much

kshirajahere:
@joeloskarsson Please have a look when you get time sir!

joeloskarsson (approved):
Nice! This looks good now. Will run this through the next dev meeting (forgot to add it to the roadmap during the last one 😅), but see no reason not to merge this in.

neural_lam/metrics.py
```diff
@@ -330,3 +329,18 @@ def crps_gauss(
...
+    """Return whether the metric can train a learned predictive std-dev."""
+    metric_name_lower = metric_name.lower()
+    if metric_name_lower not in DEFINED_METRICS:
+        raise ValueError(f"Unknown metric: {metric_name}")
+    return metric_name_lower in OUTPUT_STD_COMPATIBLE_METRICS
+
+
+def get_output_std_compatible_metrics():
+    """Return the loss names that can train a learned predictive std-dev."""
+    return sorted(OUTPUT_STD_COMPATIBLE_METRICS)
```
joeloskarsson, line 346: In the meantime, just a small improvement: Could you make these docstrings follow the numpy style and explicitly specify parameters and return values/types? Trying to be more strict with this in the codebase going forward.

joeloskarsson:
This is approved and ready to be merged now :) Could you fix my small docstring suggestion and the test conflict and I'll go ahead and merge this in.

PR #420, opened 2026-03-17
Fix : --metrics_watch  prefix  mismatch in eval phase

archit7-beep:
## Describe your changes

This PR adds a warning when metrics specified in --metrics_watch are not matched during validation or test phases.

Previously, if a user provided a metric with an incorrect prefix (e.g., val_rmse during --eval test), the metric would be silently ignored with no feedback. This could lead to confusion, as no scalar metrics would be logged without any indication of misconfiguration.

This change tracks which metrics are actually matched during evaluation and emits a warnings.warn() at the end of the epoch for any unmatched entries.

## Issue Link

Fixes :  #418 

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [ ] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [x] the code is readable
- [x] the code is well tested
- [x] the code is documented (including return types and parameters)
- [x] the code is easy to maintain

## Author checklist after completed review

- [x] I have added a line to the CHANGELOG describing this change, in a section
  reflecting type of change (add section where missing):
  - *added*: when you have added new functionality
  - *changed*: when default behaviour of the code has been changed
  - *fixes*: when your contribution fixes a bug
  - *maintenance*: when your contribution is relates to repo maintenance, e.g. CI/CD or documentation

## Checklist for assignee

[...]

PR #432, opened 2026-03-18
Implement automatic prefix normalization for metrics_watch

### LINKED ISSUE #418: Bug/Docs: `--metrics_watch` silently logs nothing when prefix doesn't match eval phase
## Problem

`--metrics_watch` accepts metric names including their prefix (e.g. `val_rmse`), as documented in the CLI help:

```
--metrics_watch: List of metrics to watch, including any prefix (e.g. val_rmse)
```

The matching logic in `ar_model.py` is:

```python
full_log_name = f"{prefix}_{metric_name}"  # e.g. "val_rmse" or "test_rmse"
if full_log_name in self.args.metrics_watch:
    ...  # log per-variable scalar values
```

The `prefix` is `"val"` during validation and `"test"` during testing (`--eval test`).

This means:

| Command | `--metrics_watch` value | Result |
|---|---|---|
| Training/validation | `val_rmse` | ✅ works |
| `--eval test` | `val_rmse` | ❌ **silently logs nothing** |
| `--eval test` | `test_rmse` | ✅ works |

**There is no error, no warning, and no indication that the watched metric was never matched.** A user who trains with `--metrics_watch val_rmse` and then switches to `--eval test` without changing the flag will lose all watched metric output without any notice.

## Steps to reproduce

```bash
    --config_path <config> \
    --model graph_lam \
    --eval test \
    --load <checkpoint> \
    --metrics_watch val_rmse \          # wrong prefix for test phase
    --var_leads_metrics_watch '{"0": [1, 2]}'
```

No error is raised. No scalar metrics are logged. The `--metrics_watch` configuration is silently ignored.

## Expected behaviour

At least one of:

1. **Documentation fix**: The `--metrics_watch` help text should explicitly state that the prefix must match the evaluation phase — `val_rmse` for validation, `test_rmse` for `--eval test`.
2. **Warning at epoch end**: After each validation/test epoch, if any name in `self.args.metrics_watch` was never matched against any `full_log_name`, emit a `warnings.warn()` listing the unmatched keys.
3. **Automatic prefix normalisation**: Strip any leading `val_`/`test_` prefix from the configured value and re-apply it based on the active phase — so `val_rmse` and `test_rmse` both work regardless of eval mode.

Option 2 (a `warnings.warn()`) provides the lowest-friction safety net and can be added with minimal code changes.

## Context

Discovered while auditing `ar_model.py` during the work on PR #414.

ManasMishra4:
This PR implements automatic prefix normalization for the --metrics_watch argument. It allows the CLI to handle metric names (e.g., val_rmse) interchangeably across validation and test phases, resolving the silent logging issue identified in #418.

Technical Improvements

Metric Normalization: Modified ar_model.py to strip and re-apply phase prefixes (val_/test_) dynamically.

Dimensionality Robustness: Added guards in vis.py and ar_model.py to handle 1D tensors, preventing IndexError during evaluation on small datasets like meps_example.

PyTorch 2.6 Support: Updated train_model.py with safe_globals to ensure compatibility with PyTorch 2.6+ model loading security.

Motivation
While #420 introduces a warning for prefix mismatches, this PR implements the suggested automatic normalization to improve the user experience and prevent data loss during evaluation.

Fixes #418

### DIFF
--- a/neural_lam/models/ar_model.py
+++ b/neural_lam/models/ar_model.py
@@ -622,14 +622,27 @@ class ARModel(pl.LightningModule):
                 delimiter=",",
             )
 
+        # --- FIX STARTS HERE ---
         # Check if metrics are watched, log exact values for specific vars
         var_names = self._datastore.get_vars_names(category="state")
-        if full_log_name in self.args.metrics_watch:
-            for var_i, timesteps in self.args.var_leads_metrics_watch.items():
+        
+        # Normalize watch list to handle 'val_' vs 'test_' interchangeably
+        watches_clean = [m.replace("val_", "").replace("test_", "") for m in self.args.metrics_watch]
+
+        # Use the smart match: check full name OR the cleaned base name
+        if full_log_name in self.args.metrics_watch or metric_name in watches_clean:
+            for var_i_str, timesteps in self.args.var_leads_metrics_watch.items():
+                var_i = int(var_i_str) # Ensure index is an integer
                 var_name = var_names[var_i]
                 for step in timesteps:
                     key = f"{full_log_name}_{var_name}_step_{step}"
-                    log_dict[key] = metric_tensor[step - 1, var_i]
+                    
+                    # Robustness: Handle 1D tensors (common in test/small runs) 
+                    # and 2D tensors (standard training/eval)
+                    if metric_tensor.ndim == 1:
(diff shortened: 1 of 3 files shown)


sadamov:
@ManasMishra4 please contribute directly to #420 and add your input in that PR instead.

PR #435, opened 2026-03-18
Clear test_metrics in on_test_epoch_end for consistency with val_metrics

### LINKED ISSUE #433: Clear test_metrics in on_test_epoch_end for consistency with val_metrics cleanup pattern
### Description

In `ARModel.on_test_epoch_end()`, `self.test_metrics` is never cleared after aggregation. This is inconsistent with:

- `on_validation_epoch_end()`, which explicitly clears `self.val_metrics`
- `self.spatial_loss_maps`, which IS cleared at the end of the same `on_test_epoch_end()` method

This is **not an active bug** — in the standard neural-lam workflow, `trainer.test()` is only called once (via `--eval`). The clearing of `val_metrics` is necessary because validation runs multiple epochs during `trainer.fit()` — that's not the case for test.

This is a **code consistency improvement** to align `on_test_epoch_end` with the cleanup pattern used in `on_validation_epoch_end`.

### Relevant code

`on_validation_epoch_end` (clears metrics):

```python
def on_validation_epoch_end(self):
    self.aggregate_and_plot_metrics(self.val_metrics, prefix="val")
    # Clear lists with validation metrics values
    for metric_list in self.val_metrics.values():
        metric_list.clear()
```

`on_test_epoch_end` (does NOT clear test_metrics):

```python
def on_test_epoch_end(self):
    self.aggregate_and_plot_metrics(self.test_metrics, prefix="test")
    # ... spatial loss map plotting ...
    self.spatial_loss_maps.clear()  # <-- this IS cleared
    # test_metrics is NOT cleared  <-- inconsistency
```

### Proposed fix

Add after the `spatial_loss_maps.clear()` call in `on_test_epoch_end`:

```python
for metric_list in self.test_metrics.values():
    metric_list.clear()
```

RajdeepKushwaha5:
### Describe your changes

`on_test_epoch_end` does not clear `test_metrics` after aggregation. This is inconsistent with `on_validation_epoch_end`, which clears `val_metrics`, and with `spatial_loss_maps`, which IS cleared in the same method.

This is a consistency improvement, not a bug fix — `trainer.test()` is only called once in normal usage. Adding the `.clear()` call aligns the test cleanup pattern with validation.

### Issue Link

closes #433

### Type of change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,8 @@ This release introduces new features including GIF animation support, wandb run 
 
 ### Maintenance
 
+- Clear `test_metrics` lists in `on_test_epoch_end` for consistency with `val_metrics` clearing in `on_validation_epoch_end` [\#433](https://github.com/mllam/neural-lam/issues/433) [\#435](https://github.com/mllam/neural-lam/pull/435) @RajdeepKushwaha5
+
 - Update PR template to clarify milestone/roadmap requirement and maintenance changes [\#186](https://github.com/mllam/neural-lam/pull/186) @joeloskarsson
 
 - Update CI/CD to use python 3.13 for testing and full range of current python versions for linting (3.10 - 3.14) [\#173](https://github.com/mllam/neural-lam/pull/173) @observingClouds
--- a/neural_lam/models/ar_model.py
+++ b/neural_lam/models/ar_model.py
@@ -824,6 +824,10 @@ class ARModel(pl.LightningModule):
         self.matched_metrics = set()
         self.spatial_loss_maps.clear()
 
+        # Clear lists with test metrics values
+        for metric_list in self.test_metrics.values():
+            metric_list.clear()
+
     def on_load_checkpoint(self, checkpoint):
         """
         Perform any changes to state dict before loading checkpoint
--- a/tests/test_training.py
+++ b/tests/test_training.py
@@ -1,6 +1,8 @@
 # Standard library
+import tempfile
 import warnings
 from pathlib import Path
+from unittest.mock import patch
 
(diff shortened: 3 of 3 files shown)


joeloskarsson:
When would one possibly call trainer.test multiple times? It would be better if you open an issue first, then we could discuss there if this is actually a bug and how to address it.

RajdeepKushwaha5:
> When would one possibly call trainer.test multiple times? It would be better if you open an issue first, then we could discuss there if this is actually a bug and how to address it.

You're right — in the standard workflow, `trainer.test()` is only called once, so this doesn't affect normal usage.

The reason `val_metrics` is cleared in `on_validation_epoch_end` is because validation runs multiple times during `trainer.fit()` — without clearing, epoch N+1's metrics would include epoch N's batches. That's not the case for test, which only runs a single epoch per `trainer.test()` call.

So this is more of a defensive consistency fix than an active bug. The asymmetry with `spatial_loss_maps` (which IS cleared in the same `on_test_epoch_end`) made it look like an oversight, but the practical impact is minimal.

I've opened #433 to track this (will update the PR description to link it). Happy to keep this as a low-priority cleanup or close it if you'd prefer — let me know how you'd like to proceed.

sadamov:
Thanks @RajdeepKushwaha5. As you note in #433 this isn't an active bug, and @joeloskarsson already weighed in there saying it's a pure consistency change. We'll keep #437 since the `plotted_examples` reset is the actual fix that matters when running `trainer.test()` repeatedly. I propose to close this one to keep review surface small. Happy to fold the one-line `.clear()` into #437 if you'd like.

PR #469, opened 2026-03-20
Fix get_integer_time zero case and improve numerical stability

Saptami191:
## Describe your changes

This PR fixes the behavior of `get_integer_time` in `neural_lam/utils.py`.

### Changes made:
- Added explicit handling for `timedelta(0)` to return `(0, "seconds")` instead of `(0, "weeks")`
- Replaced float-based computation with integer microseconds to avoid precision issues
- Added unit tests covering:
  - standard cases (days, hours)
  - zero timedelta
  - milliseconds conversion
  - negative timedelta

---

## Issue Link

Closes #468 

---

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

---

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch
- [x] I have performed a self-review of my code
- [x] I have added docstrings where needed
- [x] I have added tests that prove my fix is effective
- [ ] I have updated the README (not required for this change)
- [x] PR title clearly describes the change

---

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested
- [x] the code is documented
- [x] the code is easy to maintain

---

## Author checklist after completed review

- [x] I have added a line to the CHANGELOG (can be added if requested by maintainer)

---

## Checklist for assignee

- [ ] PR is up to date with the base branch
- [ ] tests pass

PR #494, opened 2026-03-22
Fix get_integer_time precision and zero handling

### LINKED ISSUE #468: Improve get_integer_time: handle zero timedelta and avoid float precision issues
## Problem
The current implementation of `get_integer_time` in `neural_lam/utils.py` has two potential issues:

### 1. Ambiguous behavior for zero timedelta
For `timedelta(0)`, the function returns:
```
(0, "weeks")
```
This occurs because `0 % unit == 0` for all units, and the function returns the first match ("weeks").
While mathematically valid, this behavior may be unintuitive, since zero has no inherent scale and the choice of "weeks" is arbitrary.

### 2. Floating-point precision risk
The function relies on:
```python
total_seconds = tdelta.total_seconds()
```
and performs modulo operations with float-derived values.
Floating-point arithmetic can introduce precision inconsistencies in edge cases. For example:
```python
get_integer_time(timedelta(days=0.001))  # returns (1, 'unknown') instead of (86400, 'milliseconds')
```

---

## Steps to Reproduce
```python
from datetime import timedelta
from neural_lam.utils import get_integer_time

print(get_integer_time(timedelta(0)))
```

### Current Output
```
(0, "weeks")
```

---

## Expected Behavior
For `timedelta(0)`, returning:
```
(0, "seconds")
```
is more intuitive and avoids arbitrary unit selection.

---

### 1. Handle zero explicitly
```python
if total_microseconds == 0:
    return 0, "seconds"
```

### 2. Use integer microseconds for computation
Compute total microseconds directly from `timedelta`'s integer fields instead of using `total_seconds()`:
```python
total_microseconds = (
    tdelta.days * 86400_000000
    + tdelta.seconds * 1_000000
    + tdelta.microseconds
)
```
Python guarantees `.days`, `.seconds`, and `.microseconds` are integers, so this avoids floating-point precision issues entirely.

### 3. Maintain unit definitions in integer microseconds
```python
units = {
    "weeks":        604800_000000,
    "days":          86400_000000,
    "hours":          3600_000000,
    "minutes":          60_000000,
    "seconds":           1_000000,
    "milliseconds":          1_000,
    "microseconds":              1,
}
```

---

## Benefits
- Improves clarity for zero timedelta handling
- Eliminates floating-point precision issues
- Ensures consistent and predictable behavior
- Preserves existing `dict` structure with minimal changes

---

## Testing
The behavior has been verified with unit tests covering:
- Standard cases (days, hours)
- Unit conversion (milliseconds → seconds)
- Edge cases:
  - zero timedelta
  - negative timedelta

All tests pass after applying the proposed changes.

Saptami191:
## Describe your changes

Fix get_integer_time precision issues and zero timedelta handling.

Uses integer microseconds instead of float computation to avoid rounding errors and ensure correct unit detection.

No additional dependencies.

## Issue Link

Closes #468 

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

  reflecting type of change:
  - fixes

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Close the `PIL.Image` file handle and delete the temporary `.png` after upload in `CustomMLFlowLogger.log_image`, fixing a resource leak and temp-file accumulation in CWD; replace `sys.exit(1)` on `NoCredentialsError` with a re-raise so callers can handle the failure [\#496](https://github.com/mllam/neural-lam/pull/496) @Zrahay
 
+- Refactor `get_integer_time` to use integer microsecond arithmetic instead of `total_seconds()` floats, with explicit handling for `timedelta(0)` (now returns `(0, 'seconds')` instead of `(0, 'weeks')`) and correct support for sub-second and negative timedeltas. Current callers all pass whole-hour `datastore.step_length` so behaviour is unchanged in practice; this is a defensive correctness improvement [\#494](https://github.com/mllam/neural-lam/pull/494) @Saptami191
+
 ### Maintenance
 
 - Add comprehensive type hints to `neural_lam/metrics.py` [\#447](https://github.com/mllam/neural-lam/pull/447) @sidhantpande
--- a/neural_lam/utils.py
+++ b/neural_lam/utils.py
@@ -654,22 +654,35 @@ def get_integer_time(tdelta: datetime.timedelta) -> tuple[int, str]:
     >>> get_integer_time(timedelta(milliseconds=1000))
     (1, 'seconds')
     >>> get_integer_time(timedelta(days=0.001))
-    (1, 'unknown')
+    (86400, 'milliseconds')
+    >>> get_integer_time(timedelta(0))
+    (0, 'seconds')
+    >>> get_integer_time(timedelta(days=-7))
+    (-1, 'weeks')
     """
-    total_seconds = tdelta.total_seconds()
+    # Use integer microseconds rather than tdelta.total_seconds() (a float)
+    # to avoid floating-point precision errors when checking divisibility.
+    total_microseconds = (
+        tdelta.days * 86400_000_000
+        + tdelta.seconds * 1_000_000
+        + tdelta.microseconds
+    )
+
+    if total_microseconds == 0:
+        return 0, "seconds"
 
     units = {
-        "weeks": 604800,
-        "days": 86400,
-        "hours": 3600,
-        "minutes": 60,
-        "seconds": 1,
-        "milliseconds": 0.001,
-        "microseconds": 0.000001,
+        "weeks": 604800_000_000,
+        "days": 86400_000_000,
+        "hours": 3600_000_000,
+        "minutes": 60_000_000,
+        "seconds": 1_000_000,
+        "milliseconds": 1_000,
+        "microseconds": 1,
     }
 
-    for unit, unit_in_seconds in units.items():
-        if total_seconds % unit_in_seconds == 0:
-            return int(total_seconds / unit_in_seconds), unit
+    for unit, unit_us in units.items():
+        if total_microseconds % unit_us == 0:
+            return total_microseconds // unit_us, unit
 
     return 1, "unknown"


Saptami191:
@joeloskarsson Recreated this PR after fixing earlier issues with the template and workflow. The code, tests, and linting are now clean and aligned with project guidelines. Would appreciate a review.

neural_lam/utils.py
```diff
@@ -601,22 +601,31 @@ def get_integer_time(tdelta) -> tuple[int, str]:
         >>> get_integer_time(timedelta(milliseconds=1000))
         (1, 'seconds')
         >>> get_integer_time(timedelta(days=0.001))
-        (1, 'unknown')
+        (86400, 'milliseconds')
+        >>> get_integer_time(timedelta(0))
+        (0, 'seconds')
     """
-    total_seconds = tdelta.total_seconds()
+    total_microseconds = (
```
kshirajahere, line 665: The new implementation is fine, but this helper’s contract is still inconsistent across the tree: `compute_standardization_stats.py` still calls `get_integer_time(step_length.total_seconds())`, i.e. with a float rather than a `timedelta`. I think we should either update that caller in the same slice or explicitly tighten/document the contract here.

tests/test_utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+def test_hours():
+    assert get_integer_time(timedelta(hours=5)) == (5, "hours")
+
+
+def test_zero():
+    assert get_integer_time(timedelta(0)) == (0, "seconds")
+
+
+def test_milliseconds():
```
kshirajahere: Since this PR is specifically about float-precision avoidance, I think it would be good to add one direct regression test for the issue-body example (`timedelta(days=0.001)` -> `(86400, "milliseconds")`) here as well, rather than leaving that behavior covered only by the doctest.

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix typo in `ar_model.py` that causes `AttributeError` during evaluation [\#204](https://github.com/mllam/neural-lam/pull/204) @ritinikhil
 
+- Fix `get_integer_time` to avoid floating-point precision issues and correctly handle zero timedelta [#469](https://github.com/mllam/neural-lam/pull/469) @Saptami191
```
kshirajahere: Shud point to #494 instead of #469

kshirajahere (commented):
I think the integer-microseconds direction is the right fix, but I think there is still few thing worth addressing before merge.

CHANGELOG.md
Saptami191: Thanks for the review! I hve addressed all three points.

tests/test_utils.py
observingClouds: This is actually already tested in the docstring.

tests/test_utils.py
observingClouds, line 1: All these newly added tests are already covered by the doctest. No need to replicate them here. If you think a case is missing in the docstring, please add them there.

observingClouds (changes requested):
Thanks for this PR. I left a few comments and suggest a few changes.

tests/test_utils.py
Saptami191: @observingClouds thanks for the review! yeah all are tested in docstring. i will remove this test file .actually it was part of proving the fix is correct.
Should I also add a doctest for negative timedelta
 >>> get_integer_time(timedelta(days=-7))
    (-1, 'weeks')  

it will represent a duration in the past 7 days ago.
 or is that out of scope for this PR?

Saptami191:
@observingClouds Should `get_integer_time` explicitly support negative timedeltas (e.g., `timedelta(days=-7) -> (-1, "weeks")`), or is that out of scope for this PR?

tests/test_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    assert get_integer_time(timedelta(milliseconds=1000)) == (1, "seconds")
+
+
+def test_negative():
+    assert get_integer_time(timedelta(days=-7)) == (-1, "weeks")
```
observingClouds: Yes, please add this test to the doctest and then all is covered by the doctest and we can remove all tests here

observingClouds (changes requested):

Saptami191:
@observingClouds Hii! Requested changes are done and all tests have been passed

sadamov (approved):
@observingClouds @Saptami191 I just rebased this PR onto main. Feel free to have a look and merge if you agree.

PR #507, opened 2026-03-24
Adds `PropagationNet` GNN layer and makes it optionally usable in existing deterministic models

Sir-Sloth-The-Lazy:
## Describe your changes

Adds `PropagationNet` GNN layer and makes it optionally usable in existing deterministic models, as outlined in #62.

It is integrated into the existing model hierarchy from #208 and can be enabled via the `vertical_propnets` flag.

Depends on #208.
For changes on top of #208 only, see:
https://github.com/Sir-Sloth-The-Lazy/neural-lam/compare/refactor/model-class-hierarchy-issue-49...refactor/batch-fold-ensemble-prep

## Issue Link

Contributes to #62

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass


Sir-Sloth-The-Lazy:
@joeloskarsson @sadamov @observingClouds please have a look , if this qualifies as the next step in ensemble prep 😄 . Would be grateful for your feedback !

neural_lam/interaction_net.py
```diff
@@ -139,6 +131,83 @@ def aggregate(self, inputs, index, ptr, dim_size):
...
+        (optionally) edge_rep: (M, d_h), updated vector representations
+            of edges
+        """
+        # Always concatenate to [rec_nodes, send_nodes] for propagation,
+        # but only aggregate to rec_nodes
+        node_reps = torch.cat((rec_rep, send_rep), dim=-2)
+        edge_rep_aggr, edge_diff = self.propagate(
+            self.edge_index, x=node_reps, edge_attr=edge_rep
+        )
+        rec_diff = self.aggr_mlp(
```
Debadri-das: `PropagationNet` handles receivers with no incoming edges differently than `InteractionNet`.
The problem is in :

```python
# In InteractionNet:
def forward(self, send_rep, rec_rep, edge_rep):
    ...
    rec_diff = self.aggr_mlp(torch.cat((rec_rep, edge_rep_aggr), dim=-1))
    rec_rep = rec_rep + rec_diff   #Better as it Preserves original node values


# In PropagationNet :
def forward(self, send_rep, rec_rep, edge_rep):
    ...
    rec_diff = self.aggr_mlp(torch.cat((rec_rep, edge_rep_aggr), dim=-1))
    rec_rep = edge_rep_aggr + rec_diff  #Loses the original values of node

**Comment by Debadri-das** [comment]
@joeloskarsson @sadamov  would request your further review on this PR!

File: neural_lam/models/base_graph_model.py [hunk 2]
```diff
@@ -5,23 +5,66 @@
...
+        datastore: BaseDatastore,
+        graph_name: str = "multiscale",
+        hidden_dim: int = 64,
+        hidden_layers: int = 1,
+        processor_layers: int = 4,
+        mesh_aggr: str = "sum",
+        num_past_forcing_steps: int = 1,
+        num_future_forcing_steps: int = 1,
+        output_std: bool = False,
+        vertical_propnets: bool = False,
```
joeloskarsson: I think we need more fine-grained options for where to use propnets here (g2m, m2g, both). We should also make this more future-proof by not making it a boolean inet/propnet, but rather having the argument be the GNN type to use (as a string? enum? not sure about best design).

joeloskarsson: This goes also for other model classes with this argument.

neural_lam/interaction_net.py
```diff
@@ -139,6 +131,83 @@ def aggregate(self, inputs, index, ptr, dim_size):
...
+        )
+
+        # Residual connections
+        rec_rep = edge_rep_aggr + rec_diff  # residual is to aggregation
+
+        if self.update_edges:
+            edge_rep = edge_rep + edge_diff
+            return rec_rep, edge_rep
+
+        return rec_rep
```
joeloskarsson: There is quite a lot of repeated code from the InteractionNets forward here. Could this be refactored to avoid as much repetition, while keeping some clarity of the differences between the two classes?

joeloskarsson (changes requested):
Thanks for looking at this, and sorry for being slow with reviews 😅 Shared some first thoughts here on how I think we can best integrate this. Happy to hear input also from others, as there are some non-trivial design choices around this (e.g. how to choose the GNN type for each sub-graph).

Sir-Sloth-The-Lazy:
## Refactor: replace `vertical_propnets` bool with fine-grained GNN type strings 

### Overview

Replaced the single `--vertical_propnets` boolean flag with four per-direction string parameters, enabling independent GNN selection per message-passing direction.

---

### Changes

- **Per-direction GNN type parameters**
  Replaced `--vertical_propnets` with four string args: `--g2m_gnn_type`, `--m2g_gnn_type`, `--mesh_up_gnn_type`, `--mesh_down_gnn_type`. Each defaults to "InteractionNet" and can be set to "PropagationNet".
- **`GNN_TYPES` registry + `get_gnn_class()` lookup** (`interaction_net.py`)
  Added a registry to resolve GNN type strings to their corresponding classes, making it straightforward to register new GNN types in the future.

- **`PropagationNet` forward() deduplication**
  Eliminated the duplicated `forward()` method by extracting a `node_residual_target()` hook:
  - `InteractionNet.node_residual_target()` → returns the receiver representation (rec_rep)
  - `PropagationNet.node_residual_target()` → returns the aggregated edge messages (edge_rep_aggr)
  -   Shared `forward()` now delegates to this hook.

- **CLI wiring + propagation through model hierarchy**
  Threaded the per-direction types through `train_model.py`, `BaseGraphModel`, `BaseHiGraphModel`, `HiLAM`, `HiLAMParallel`, and `GraphLAM`.

- **Integration tests updated**
  Tests now exercise the string-based API, including per-direction overrides and invalid-type error handling.

---

### Files Changed (8)

| File | Nature of change |
|------|-----------------|
| `interaction_net.py` | Added `GNN_TYPES` registry, `get_gnn_class()`, `node_residual_target()` hook; unified `forward()` |
| `base_graph_model.py` | Threaded per-direction GNN type args |
| `base_hi_graph_model.py` | Threaded per-direction GNN type args |
| `graph_lam.py` | Wired new GNN type params |
| `hi_lam.py` | Wired new GNN type params |
| `hi_lam_parallel.py` | Wired new GNN type params |
| `train_model.py` | Replaced `--vertical_propnets` with four `--*_gnn_type` CLI args |
| `test_propagation_net.py` | Updated tests for string-based API and error handling |

Sir-Sloth-The-Lazy:
## Proposal: GNN type selection per sub-graph

### Context

With the current commit (`3472387`), four per-direction CLI args cover the encoder/decoder
and inter-level GNNs. However, three sub-graphs remain hardcoded to `InteractionNet`:

| Sub-graph | Location | Currently configurable? |
|-----------|----------|------------------------|
| m2m same-level | `HiLAM.make_same_gnns` | No |
| m2m processor | `GraphLAM.__init__` | No |
| processor (all edges) | `HiLAMParallel.__init__` | No |

> **Note on m2m same-level edges:** The original `--vertical_propnets` flag deliberately
> excluded same-level m2m processing, nodes at the same mesh level are peers with no
> clear sender→receiver directionality, so `PropagationNet` (designed for directional
> vertical message passing) may not be semantically meaningful here. Any global default
> that switches everything would affect m2m same-level too, which is worth keeping in mind.

---

### Option A: Single `--gnn_type` default + per-direction overrides

Add one global `--gnn_type` arg (default `"InteractionNet"`). Per-direction args
(`--g2m_gnn_type`, `--m2m_gnn_type`, etc.) default to `None` and override the global
when set. Currently-hardcoded sub-graphs in `HiLAM` and `GraphLAM` fall back to the
global default.

**Resolution logic:**
```python
g2m_gnn_type = g2m_gnn_type or gnn_type  # per-direction wins if set
```

**Pros:**
- Clean UX for the common case — `--gnn_type PropagationNet` switches everything in one flag
- Fine-grained control still available via per-direction overrides
- Small diff on top of the current implementation
- Closes the hardcoded gaps (except `HiLAMParallel`) without adding more args

**Cons:**
- `None`-vs-default fallback logic adds a layer of indirection to understand
- A blanket `--gnn_type PropagationNet` would also switch m2m same-level edges, which may
  not be desirable (mitigated by `--m2m_gnn_type InteractionNet` override, but the user
  has to know to do this)
- Still CLI-arg-driven, which will eventually need migration if the codebase moves fully
  to config files

---

### Option B: YAML config-driven (via `NeuralLAMConfig`)

Move GNN type selection into the existing `NeuralLAMConfig` YAML config. The codebase
already uses `dataclass_wizard.YAMLWizard` with `Union`-based type selection for training
config (loss weighting, output clamping). Model architecture config would follow the same
pattern:
```yaml
model:
  hidden_dim: 64
  processor_layers: 4
  gnn_types:
    g2m: PropagationNet
    m2g: InteractionNet
    mesh_up: PropagationNet
    mesh_down: InteractionNet
    m2m: InteractionNet
```

**Pros:**
- Consistent with the direction the codebase is already moving (`NeuralLAMConfig` handles
  datastore, loss weighting, output clamping via YAML)
- Scales cleanly to any number of sub-graphs or future model config without CLI bloat
- Config files are versioned, shareable, and reproducible — easier to track experiment setups
- No fallback/override logic needed; each sub-graph has an explicit entry
- Makes the m2m same-level choice explicit rather than inheriting from a global default

**Cons:**
- Larger refactor — needs new dataclasses, migration of existing CLI model args
- Scope creep risk if done within this PR (`--hidden_dim`, `--processor_layers` would
  ideally move at the same time)
- Diverges from the current CLI-driven workflow that other model args still use

---

### Recommendation

**Option A now, Option B later.**

Option A is a small, self-contained change that fits the existing CLI pattern and closes
the hardcoded gaps in `HiLAM` and `GraphLAM` immediately. Option B is the right
long-term direction but is better done as a dedicated refactor that moves all model
hyperparameters into `NeuralLAMConfig` at once, rather than partially migrating just
GNN types.

> **`HiLAMParallel` remains out of scope for both options** , its processor fuses all
> edge types (m2m + up + down) into a single `InteractionNet` with
> `edge_chunk_sizes`/`aggr_chunk_sizes`, so per-direction GNN type selection is not
> applicable without rethinking its architecture.

neural_lam/train_model.py
```diff
@@ -128,6 +123,34 @@ def main(input_args=None):
...
+        help="GNN type for upward mesh message passing in hierarchical "
+        "models (e.g. InteractionNet, PropagationNet)",
+    )
+    parser.add_argument(
+        "--mesh_down_gnn_type",
+        type=str,
+        default="InteractionNet",
+        help="GNN type for downward mesh message passing in hierarchical "
+        "models (e.g. InteractionNet, PropagationNet)",
+    )
```
joeloskarsson, line 188: Could be nice to explicitly add the possible choices in argparse, using the `choices` keyword with the registered GNN dict.

neural_lam/gnn_layers.py
joeloskarsson, line 1: Should we now rename this file to something like "gnn_layers.py"

joeloskarsson (changes requested):
Looking really nice now! Previous comments addressed, I just added some small suggestions here that is really just polishing.

The outstanding question is indeed what to do with the remaining GNN layers, if we should make them all configurable. In my opinion there is not that much point in making the GNNs for m2m edges propagation-networks, so for most practical use cases the present options are sufficient. But it is a bit strange that some of the GNNs can be changed but not all. I think I am leaning towards trying to merge this as is atm, and then leaving a further generalization (e.g. through a model-config file) as a future change. But this requires some discussion I think. I will propose this for the next release version, and also see what others at the dev meeting think about the configurability question.

Sir-Sloth-The-Lazy:
Hi @joeloskarsson made the changes , agreed with your verdict and hope you found the proposal a little helpful ! ;)

joeloskarsson:
From the dev-meeting there seems to be agreement to stick to the current design of cli-arguments for choosing some gnn layers, and leaving the broader config to later.

Sir-Sloth-The-Lazy:
Understood , I would love to work on that when the time comes 😄

joeloskarsson (approved):
The implementation here is all good now, so approving this. This should however go in after #208, so need to wait after possible conflict resolution with that (although might not be much as it is built on top of that).

The one thing missing is a changelog entry :)

joeloskarsson:
Now that #208 is in only the conflicts have to be resolved here (+ changelog entry), and then we can merge this as well.

Sir-Sloth-The-Lazy:
On it !

Sir-Sloth-The-Lazy:
@joeloskarsson I have resolved the merge conflicts, I hope the branch is ready to be merged.

neural_lam/gnn_layers.py
```diff
@@ -62,15 +62,18 @@ def __init__(
...
-        # The edge indices used in this GNN layer are defined as:
+        # Lay out the combined node tensor as [receivers | senders]:
         #   receivers → [0 .. num_rec-1]
         #   senders   → [num_rec .. num_rec+num_snd-1]
-        # Hence, sender indices from the input edge_index are offset
-        # by num_rec to obtain the indices used in this layer.
+        # Offset sender indices accordingly.
         edge_index = torch.stack(
             (edge_index[0] + self.num_rec, edge_index[1]), dim=0
         )
```
joeloskarsson: I think this might be a regression in the merge? With the current state on main, we should not need this re-indexing of edge_index that is re-introduced here.

.github/workflows/install-and-test.yml
```diff
@@ -62,9 +62,20 @@ jobs:
...
+          echo "Minimum torch version required: ${MIN_VERSION}"
+          # Find the latest version satisfying that minimum in the target index.
+          # This returns the highest version present in the wheel index (e.g. cu128)
+          # rather than the highest version on PyPI.
+          TORCH_VERSION=$(python -m pip install --dry-run "torch>=${MIN_VERSION}" \
+            --index-url ${{ matrix.torch_index }} 2>&1 \
+            | grep "Would install" | grep -o 'torch-[0-9.]*' \
+            | awk -F'-' '{print $2}' | tail -n 1)
+          echo "Torch version to install: ${TORCH_VERSION}"
+          echo "TORCH_VERSION=${TORCH_VERSION}" >> $GITHUB_ENV
```
joeloskarsson, line 78: How does this workflow change relate to the GNN layers?

joeloskarsson (changes requested):
Thanks for fixing the conflicts. I think the reindexing of edge_indexis something that needs another look. Unless I am confusing things, this shouldn't be needed now, and I don't think this is done on main. Then I also just had a question about the workflow change.

Rest is looking good, just as before! So we can just tidy up these last things and then be ready to merge :)

.github/workflows/install-and-test.yml
Sir-Sloth-The-Lazy: @joeloskarsson, you're right that the workflow change is unrelated to the GNN layers. I can revert it from this PR. I wanted to explain why it was added and what the options are.

**The problem**

The GPU CI job was failing with:


`ERROR: Could not find a version that satisfies the requirement torch==2.12.0
(from versions: 2.7.0+cu128 ... 2.11.0+cu128)`
The "Determine PyTorch version" step runs `pip install --dry-run "."` which queries PyPI and resolves `torch==2.12.0 (the latest release)`. The subsequent install step then tries to fetch that exact version from the cu128 wheel index, which only carries up to `2.11.0+cu128`.

This is a general infrastructure issue not specific to this PR and it will affect every future PR until the cu128 index catches up to PyPI.

### What the workflow change does ?

Instead of resolving the torch version from PyPI, it reads the lower bound directly from pyproject.toml (`torch>=2.3.0`) and queries the target wheel index for the latest version available there. For `cu128` that gives `2.11.0`; for CPU it gives whatever is current on PyPI. This ensures the pinned version is always present in whichever index the install uses.

**Options going forward**

Revert from this PR, open a separate PR, Keep #507 clean, fix the CI in a standalone PR against main. The GPU job on this PR will likely keep failing in the meantime.
Keep it here. Merge the fix as part of this PR since the CI is actively broken and needs to unblock testing.

Happy to do whichever you prefer. If you'd like me to revert it from here and open a separate PR I can do that right away. If I am deducing something wrong please correct me

neural_lam/gnn_layers.py
Sir-Sloth-The-Lazy: Thanks for catching that @joeloskarsson. You were right, the normalization in InteractionNet.__init__ was masking a bug elsewhere rather than being the correct fix.

**Root cause**

`create_graph.py` builds the combined `g2m/m2g` graph by placing all mesh nodes first (all hierarchy levels combined), then grid nodes. So for a 3-level hierarchical graph with `729 + 81 + 9 = 819 total mesh nodes`, grid node indices in the raw graph files start at `819`.

`zero_index_g2m` and `zero_index_m2g` in `utils.py` are responsible for zero-basing those grid indices before the edge_index reaches InteractionNet. But both functions were computing the `offset using mesh_static_features[0].shape[0]`, the bottom-level mesh count (729), instead of the total across all levels (819). This left grid indices at [90, 7769] instead of [0, 7679], which then caused an `IndexError` during the HiLAM forward pass.

Fix (`f4b9e3e`)

Changed `mesh_static_features[0].shape[0]` to sum(`sf.shape[0]` for sf in mesh_static_features) in both helpers. For a flat (single-level) graph the sum reduces to the same value as before, so non-hierarchical behaviour is unchanged. For hierarchical graphs it now correctly offsets by the full mesh node count.

With load_graph producing properly zero-based indices, the normalization inside `InteractionNet.__init__` is no longer needed and has been removed.

.github/workflows/install-and-test.yml
joeloskarsson: I think fixing this in a separate PR would be better, if nothing else it is more urgent than this PR. I'll coordinate on slack and ping you there :)

joeloskarsson: Moved to #639

neural_lam/gnn_layers.py
joeloskarsson: Ok, so if I understand correctly this is a bug currently on main? If so we should not just throw in a bugfix here, but rather have a separate issue documenting this and a separate PR to fix it. But I am not entirely convinced that this is indeed the case, specifically the way that m2g/g2m are saved now, are the grid nodes really offset by the number of mesh nodes in all levels? Since only the bottom level is actually in the networkx object converted to pyg. It also seems to me that we would have tests catching this, but could be that it really slips through somehow.
Can you share what error you run into? And, assuming that this is the case also on main with InteractionNets, make a separate issue about this.

neural_lam/gnn_layers.py
```diff
@@ -63,14 +63,11 @@ def __init__(
...
-        #   edge_index[1]: receiver indices in [0 .. num_rec-1]
-        # The edge indices used in this GNN layer are defined as:
+        # Expects zero-based, locally-indexed edge_index from the caller
+        # (load_graph in utils.py normalises g2m/m2g before passing here).
+        # Lay out the combined node tensor as [receivers | senders]:
         #   receivers → [0 .. num_rec-1]
         #   senders   → [num_rec .. num_rec+num_snd-1]
-        # Hence, sender indices from the input edge_index are offset
-        # by num_rec to obtain the indices used in this layer.
         edge_index = torch.stack(
```
joeloskarsson: Don't see a good reason to change this comment to something less descriptive.

tests/test_prediction_model_classes.py
joeloskarsson, line 1: Sorry if I missed this earlier, but do the changes to this file really belong with this PR? Does this realate to the PropNets?

neural_lam/gnn_layers.py
Sir-Sloth-The-Lazy: Thanks @joeloskarsson for the question , it helped me to verify the issue with certainity, I dug into this carefully and want to lay out exactly what's happening so we can decide on the right path.

**Is it a bug on main?** **Yes, but it only affects hierarchical graphs (n_levels > 1).**

The inconsistency is between how `create_graph.py` writes the `g2m/m2g` graphs and how `utils.py` reads them back:

- In `create_graph.py`, for the hierarchical branch, `joint_mesh_graph = networkx.union_all([graph for graph in G])` builds all_mesh_nodes from every level, and `G_grid.add_nodes_from(all_mesh_nodes)` adds all of them before `from_networkx(G_g2m)`.
- In `utils.py`, both `zero_index_g2m/zero_index_m2g` subtract only `mesh_static_features[0].shape[0]`, the bottom-level count.

**On the specific point about only the bottom level being in the pyg object**: that's true for the edges (`vm = G_bottom_mesh.nodes` only drives edge creation), but not for the nodes. The node set added to `G_g2m/G_m2g` is `all_mesh_nodes` from the `union_all` over all levels. Since mesh keys are `(0,…)/(1,…)/(2,…)` and grid keys are `(1000,…)`, the `sorted(...)` call places all mesh levels before the grid nodes, so `from_networkx` numbers grid nodes after the full mesh count. For a 729+81+9 graph that's an offset of 819, while the loader only subtracts 729.

For flat graphs `len(mesh_static_features) == 1`, so `sum(...)` reduces exactly to `mesh_static_features[0].shape[0]` and behaviour is unchanged which is why this has gone unnoticed.

On why tests don't catch it: `test_graph_creation.py` does parametrize over hierarchical, but it only asserts the saved files exist and have the right shapes it never calls `load_graph()` (the `zero_index_*` path) and never runs a hierarchical forward pass. So the indexing mismatch isn't exercised by the existing suite.

The error: with grid indices left offset (e.g. `[90, 7769]` instead of `[0, 7679]`), the `InteractionNet` scatter/gather during the HiLAM forward goes out of bounds and raises an `IndexError`. (I can attach the full traceback.)

On process fully agree.  I'll pull it out of this PR.

neural_lam/gnn_layers.py
Sir-Sloth-The-Lazy: Agreed ! Sorry for this 😅

tests/test_prediction_model_classes.py
Sir-Sloth-The-Lazy: you are correct these don't belong here. There are two unrelated edits to this file: a test rename/docstring (`test_forecaster_module_old_checkpoint` → test_old_checkpoint_load) that's leftover #208 checkpoint-refactor churn, and a `test_fold_unfold_equivalence` test for the batch-fold work. Neither is PropNet-related, and the batch-fold production code is already in main, **so these were stale test-only remnants from when this branch carried the #208 history.**

I've rebased this branch onto current main (dropping the already-merged #208 commits), which reverts `tests/test_prediction_model_classes.py` back to identical with main. Once I force-push, this file will drop out of the PR diff entirely.

neural_lam/gnn_layers.py
joeloskarsson: I see, yea that is indeed quite bad. I thought that we also ran a training step with Hi-LAM, but that does not seem to be the case (we should do this imo). Hmm, I wonder if this will break things in a switch between ´create_graph´ here and WMG for making graphs. This seems to relate strongly to #323.

Anyhow, this is unrelated to the propagation-nets, so we should not make the change here, and try to have the discussion elsewhere. Could you open an issue with the context from this discussion?

Sir-Sloth-The-Lazy:
@joeloskarsson I have done the changes. One test failing is exactly the issue being discussed in #642. Hope this is okay :)

tests/test_prediction_model_classes.py
Sir-Sloth-The-Lazy: This is my fault ! Sorry

joeloskarsson: Seems to still be in the diff atm?

neural_lam/gnn_layers.py
joeloskarsson: Resolving this, as we should fix this in #643, and then things should work here as well.

joeloskarsson:
Great, can you have another look at `tests/test_prediction_model_classes.py`, as I think the unrelated changes there are left? Then as soon as #642 is solved we can get this up to date and merge.

tests/test_prediction_model_classes.py
Sir-Sloth-The-Lazy: Will fix that 😅

joeloskarsson (approved):
All looks good now, great to get the PropagationNets merged! Good job with this :D

Sir-Sloth-The-Lazy:
Thank you very much ! Will start work on getting the encoder/decoder from prob_model_lam merged

PR #529, opened 2026-03-27
[ENH] Add config validation with descriptive error messages

### LINKED ISSUE #528: [ENH] Improve config validation error messages for missing or invalid fields
## Summary

When running `neural-lam` with a misconfigured YAML file, the current behavior raises raw `KeyError` or `AttributeError` exceptions that don't tell the user *which* field is missing or *why* it's required. This makes debugging config issues unnecessarily difficult, especially for new users.

## Current behavior

For example, if `training.state_feature_weighting` is missing from the config, the traceback exposes internal xarray/Python internals rather than pointing to the config field:
```
KeyError: 'state_feature_weighting'
```

## Expected behavior

The error should clearly state:
```
ValueError: Missing required config field: 'training.state_feature_weighting'. This field must be set to define how state features are weighted during training. See the config documentation at <link>.
```

## Proposed fix

Add a `validate()` method or validation step in `neural_lam/config.py` that checks for all required fields before any training or graph creation begins, and raises `ValueError` with a descriptive message for each missing or incorrectly typed field.

This would also make it easier to extend config validation when new fields are added — for example, fields needed for probabilistic forecasting models (GSoC 2026 project).

## Steps to reproduce

1. Remove `state_feature_weighting` from your config YAML
2. Run `python -m neural_lam.train --config your_config.yaml`
3. Observe the unhelpful traceback

## Additional context

I'm a GSoC 2026 applicant working on the Probabilistic Forecasting Models project. I noticed this while reading through `neural_lam/config.py` and the training config schema. I want to take this on — I'll open a draft PR once the approach is confirmed.

Saurabh6266:
## Summary

Closes #528

Adds a `validate_config()` function to `neural_lam/config.py` that checks for common config problems before training or graph creation begins, and raises `InvalidConfigError` with clear, actionable messages instead of raw tracebacks.

## Motivation

Two common config errors currently produce hard-to-debug tracebacks:

1. `datastore.config_path` pointing to a non-existent file causes a deep `FileNotFoundError` inside `init_datastore`, with no indication of which config field is wrong.
2. `ManualStateFeatureWeighting` with an empty `weights` dict produces a silent no-op during training rather than an early error.

Both are now caught at startup with a message pointing directly to the relevant field and showing an example fix.

## Changes

- `neural_lam/config.py`: Added `validate_config(config, config_path)` called inside `load_config_and_datastore` after YAML parsing
- `tests/test_config.py`: Added five tests for `validate_config` covering existing path (passes), missing path (raises with correct field name), error message content, empty ManualStateFeatureWeighting (raises), and non-empty ManualStateFeatureWeighting (passes). Existing tests untouched.

## Why this matters for future development

The probabilistic forecasting project and other extensions will introduce new required config fields. Having a single validation function makes it straightforward to add new checks without risk of new fields producing cryptic runtime errors deep in the training loop.

## Testing
```bash
pytest tests/test_config.py -v
pre-commit run --all-files
```

## Checklist
- [x] Tests pass locally
- [x] pre-commit passes
- [x] Existing tests unaffected

### DIFF
--- a/neural_lam/config.py
+++ b/neural_lam/config.py
@@ -153,6 +153,69 @@ class InvalidConfigError(Exception):
     pass
 
 
+def validate_config(config: "NeuralLAMConfig", config_path: str) -> None:
+    """Validate a loaded NeuralLAMConfig and raise descriptive errors.
+
+    This runs after YAML parsing so dataclass defaults are already applied.
+    It catches issues that would otherwise produce cryptic runtime tracebacks,
+    such as a missing or non-existent datastore config file.
+
+    Parameters
+    ----------
+    config : NeuralLAMConfig
+        The fully loaded config object.
+    config_path : str
+        Path to the neural-lam YAML config file. Used to resolve the
+        datastore config path, which is relative to this file.
+
+    Raises
+    ------
+    InvalidConfigError
+        If any required field is missing, invalid, or points to a
+        non-existent file.
+    """
+    errors = []
+
+    # datastore.config_path must resolve to an existing file
+    resolved = Path(config_path).parent / config.datastore.config_path
+    if not resolved.exists():
+        errors.append(
+            f"Missing required config field: 'datastore.config_path'.\n"
+            f"  Resolved path does not exist: {resolved}\n"
+            f"  This path is resolved relative to your neural-lam config "
+            f"file at: {config_path}\n"
+            f"  Check that 'config_path' in the 'datastore' section is "
+            f"correct."
+        )
+
+    # if ManualStateFeatureWeighting, weights must not be empty
+    weighting = config.training.state_feature_weighting
+    if isinstance(weighting, ManualStateFeatureWeighting):
+        if not weighting.weights:
+            errors.append(
+                "Invalid config field: 'training.state_feature_weighting.weights'.\n"
+                "  ManualStateFeatureWeighting requires at least one weight "
+                "entry.\n"
+                "  Example:\n"
+                "    training:\n"
+                "      state_feature_weighting:\n"
+                "        __config_class__: ManualStateFeatureWeighting\n"
+                "        weights:\n"
+                "          u100m: 1.0\n"
+                "          v100m: 1.0"
+            )
+
+    if errors:
+        error_list = "\n\n".join(
+            f"  [{i + 1}] {e}" for i, e in enumerate(errors)
+        )
+        raise InvalidConfigError(
+            f"neural-lam config validation failed "
+            f"({len(errors)} error(s)):\n\n{error_list}\n\n"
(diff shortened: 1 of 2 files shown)


princekumarlahon:
This is a really nice improvement catching config issues early with clear error messages will make things much easier to debug, especially for new users. I also like that validation happens right after loading the config and that you added targeted tests for the new cases. I had a few thoughts mainly around how this might scale as more config options get added:

1. Right now all the checks seem to live inside `validate_config`. That works well for now, but I wonder if it might get a bit crowded over time. Maybe grouping validations by sections (like `datastore`, `training`, etc.) could help keep things organized as it grows.

2. It might also be worth standardizing the error messages a bit more for example, always including the full config path (`training.state_feature_weighting`) and keeping a consistent format for suggested fixes.

3. Since this is meant to support future additions (like probabilistic forecasting configs), it could be useful to have a clear pattern for adding new validation rules, so contributors don’t have to modify one large function every time.

Out of curiosity, do you plan to expand validation to more fields over time, or keep it focused on the most common failure cases? Overall though, this is a great step toward making config errors much more user-friendly.

Saurabh6266:
Thanks @princekumarlahon, really appreciate the detailed feedback!

These are all good points. To respond to each:

1. **Grouping by section** — agreed that a flat `validate_config` will get crowded as more fields are added. One approach would be breaking it into `_validate_datastore` and `_validate_training` helpers called from the main function, so each section stays self-contained. Happy to restructure it that way in a follow-up commit if that direction makes sense to the maintainers.

2. **Consistent error message format** — makes sense. Right now the path format is consistent (`datastore.config_path`, `training.state_feature_weighting`) but the suggested fix wording varies slightly between checks. I can tighten that into a fixed template — something like `[field]: <what's wrong>. Example fix: <yaml snippet>` — so all messages follow the same structure.

3. **Extensibility pattern** — the intent was exactly this: new fields get registered in one place with their error description, rather than buried in ad-hoc conditionals. If the current structure isn't clear enough on that, I can add a short comment above the validation calls explaining the pattern.

On your question about scope: the plan is to keep the initial PR focused on the two most failure-prone cases (missing datastore file, empty manual weights) and expand incrementally as new config fields are introduced — especially the ones needed for the probabilistic forecasting extension. That way each addition stays reviewable and testable on its own.

Let me know if you'd like me to address any of these before review, or if it's better to handle them as follow-up PRs.

princekumarlahon:
This sounds great I like the direction. Splitting validation into section-based helpers and standardizing error messages both make sense. And +1 to keeping this PR focused and handling structural improvements as follow up. I think this is in a good place

sadamov:
i am closing this implementation for now, see linked issue for context.

PR #252, opened 2026-02-24
docs: 100 percent coverage of docstrings

Mohit-Lakra:
## Summary
Improves docstring coverage and quality across core neural-lam modules as 
groundwork for the API reference documentation system described in Issue #61.

## Changes
- `neural_lam/__init__.py`: Added module-level docstring
- `neural_lam/datastore/__init__.py`: Added module-level docstring  
- `neural_lam/models/__init__.py`: Added module-level docstring
- `neural_lam/config.py`: Converted `NeuralLAMConfig` and `InvalidConfigError` 
  to full NumPy-style docstrings
- `neural_lam/metrics.py`: Converted `mask_and_reduce_metric` and `wmse` 
  to full NumPy-style docstrings with Parameters/Returns sections
- `neural_lam/interaction_net.py`: Added complete NumPy-style docstring to 
  `InteractionNet.__init__` with all parameters documented

## Metrics
| Metric | Before | After |
|--------|--------|-------|
| Docstring coverage | 67.5% (158/234) | 100% (234/234) |
| Missing docstrings | 76 | 0 |

## Motivation
This PR is the first step toward the structured documentation system 
requested in Issue #61. Clean, consistent NumPy-style docstrings are a 
prerequisite for `sphinx-autoapi` to generate a useful API reference — 
the Jupyter Book PR (coming next) depends on this baseline quality.

PR #533, opened 2026-03-28
docs: improve clarity of graph construction comments and docstrings

Yusufiabichi:
## Summary
Improve documentation in `create_graph.py` to help new contributors understand graph mesh construction, particularly for Limited Area Modeling (LAM) vs global forecasting contexts (relates to #63).

## Changes
1. Clarify "avoid nodes on border" comment with explanation of LAM boundary handling
2. Add docstrings to helper functions: `sort_nodes_internally`, `from_networkx_with_start_index`
3. Document edge feature format in `save_edges` function
4. Expand TODO comment about directed edge visualization

## Type
Documentation / Clarity improvement (no functional changes)

This is an initial contribution and I welcome feedback on direction.

## Related
Relates to #63 (global graph support): borders/boundary handling differs between LAM and global domains.

### DIFF
--- a/neural_lam/create_graph.py
+++ b/neural_lam/create_graph.py
@@ -32,6 +32,7 @@ def plot_graph(graph, title=None):
         # Keep only 1 direction of edge_index
         edge_index = edge_index[:, edge_index[0] < edge_index[1]]  # (2, M/2)
     # TODO: indicate direction of directed edges
+    # For undirected graphs visible as lines; directed graphs need directional markers.
 
     # Move all to cpu and numpy, compute (in)-degrees
     degrees = (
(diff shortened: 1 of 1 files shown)


neural_lam/create_graph.py
```diff
@@ -71,16 +72,52 @@ def plot_graph(graph, title=None):
 
 
 def sort_nodes_internally(nx_graph):
-    # For some reason the networkx .nodes() return list can not be sorted,
-    # but this is the ordering used by pyg when converting.
-    # This function fixes this.
+    """
+    Reconstruct a NetworkX graph with nodes sorted by index.
+
+    NetworkX node ordering may differ from PyG's conversion expectation.
```
kshirajahere, line 78: PyG conversion expectation is a bit vague and sounds stronger than the code proves.

neural_lam/create_graph.py
```diff
@@ -105,6 +142,25 @@ def save_edges_list(graphs, name, base_path):
 
 
 def from_networkx_with_start_index(nx_graph, start_index):
+    """
```
kshirajahere, line 145: This overstates what the function does. It says it “handles that offset during conversion,” but the code converts first and then mutates edge_index. Also, it does not offset node features/positions, only edge indices.

kshirajahere:
This may be overlapping with #211 conceptually, since that PR also addressed boundary/global-forecasting behavior. It doesn’t look like a direct duplicate, but I’d suggest clarifying the relationship and checking that the wording here stays aligned with that earlier work.

sadamov:
hi @Yusufiabichi, since most of the changes here concer docstrings, please contribute them directly in this PR (review, or PR into that branch) #252

neural_lam/create_graph.py
Yusufiabichi: Thanks for the feedback. Could you clarify what wording would better reflect the actual behavior? Should I remove the PyG expectation mention entirely or just soften the wording?

neural_lam/create_graph.py
Yusufiabichi: Good point, thanks. Just to confirm: would you prefer I explicitly state that only edge_index is offset and not node features/positions? I can update the docstring to reflect the exact behavior.

PR #527, opened 2026-03-27
Fix probabilistic variance underflow

Panchadip-128:
## Describe your changes

This PR resolves two critical numerical and logical flaws in the probabilistic forecasting (`--output_std`) engine:

1.  **Fixed `softplus` Underflow (`NaN` crashes)**: In `neural_lam/models/base_graph_model.py`, the standard deviation calculation is now clamped to a minimum of `1e-6`. This prevents the network from producing a machine-zero variance which previously caused division-by-zero errors in NLL and CRPS metrics, leading to irreversible `NaN` training losses.
2.  **Corrected Ensemble Sampling**: The `ARModel._sample_ensemble` method in `neural_lam/models/ar_model.py` was previously discarding the model's predicted uncertainty map in favor of a hardcoded `0.01` noise fallback. This has been refactored to utilize the model's dynamically predicted `pred_std` for physically grounded ensemble generation.
3.  **Regression Testing**: Added `test_base_graph_model_prevents_softplus_underflow_nans` and `test_ar_model_ensemble_samples_from_pred_std` to `tests/test_probabilistic_forecasting.py` to assert that these mathematical stability and logic requirements are met.

**Motivation and Context**: These bugs made probabilistic training unstable and produced deceptively uniform ensemble spreads that ignored the model's internal confidence.

**Dependencies**: No new dependencies.

## Issue Link
solves #526 

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch (`upstream/main` rebased and force-pushed).
- [x] I have performed a self-review of my code.
- [x] For any new/modified functions/classes I have added docstrings.
- [x] I have placed in-line comments to clarify intent.
- [ ] I have updated the README (Not applicable as these are internal bug fixes).
- [x] I have added tests that prove my fix is effective.
- [x] I have given the PR a name that clearly describes the change.

## Author checklist after completed review

- [x] I have added a line to the `CHANGELOG.md` describing this change.
  - *fixes*: softplus variance underflow and ensemble sampling logic.

PR #537, opened 2026-03-29
Fix: prevent variance underflow causing NaN losses

Kapadiaaa:
Problem
Softplus activation can produce zero values for very negative inputs, leading to division-by-zero and NaN losses during training.

 Solution
Added a clamp (min=1e-6) to the predicted standard deviation to ensure numerical stability.

Impact
- Prevents NaN crashes
- Ensures stable probabilistic training

### DIFF
--- a/neural_lam/models/base_graph_model.py
+++ b/neural_lam/models/base_graph_model.py
@@ -351,7 +351,9 @@ class BaseGraphModel(ARModel):
             # NOTE: The predicted std. is not scaled in any way here
             # linter for some reason does not think softplus is callable
             # pylint: disable-next=not-callable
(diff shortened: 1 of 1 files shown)


kshirajahere:
Hey isnt there a draft PR #527 which resolves this. Given that overlap, I’d suggest closing this in favor of #527, unless there’s some unique piece here that should be folded into that branch.

Kapadiaaa:
Thanks for pointing that out. I will review PR #527 and see if there’s anything additional I can contribute or improve there.

sadamov:
Doing some housekeeping: this PR is a duplicate of #527, #253... closing

Issue #384, opened 2026-03-11
Supporting tensor-on-disk-format from weather-model-graphs


leifdenby:
The aim of this issue is two-fold:
- Create an implementation plan for:
   - upstream changes to `weather-model-graphs` to support writing to the tensor-on-disk format that `neural-lam` currently assumes (**need to create issue for this and link here**)
   - changes in `neural-lam` to call `weather-model-graphs` to create graphs and write these to disk in the current tensor-on-disk format in `neural-lam`
- Record what changes will be future work, i.e. what is missing from the agreed implementation plan so that we can address these issues later.

This issue aims to start "bridging the gap" between `weather-model-graphs` and `neural-lam` by building on the discussion on #339 

The implementation being planned in here is reliant on the [tensor-on-disk format to be finalised](https://github.com/mllam/neural-lam/pull/323) I think, so that we can apply the validator both on the `weather-model-graphs` and `neural-lam` side.

leifdenby:
**prior work**

1. existing cli interface in `neural_lam.create_graph` which uses `networkx` directly to create the graph and write tensor-on-disk format: https://github.com/mllam/neural-lam/blob/main/neural_lam/create_graph.py
2. cli interface for creating rectangular mesh graphs from @joeloskarsson and @sadamov's research branch: https://github.com/joeloskarsson/neural-lam-dev/blob/research/neural_lam/build_rectangular_graph.py
3. @prajwal-tech07's [comment on #339](https://github.com/prajwal-tech07) we have the following as a suggestion towards what could be added to `neural-lam`:

```python
import weather_model_graphs as wmg

xy = datastore.get_xy("state")

# Flat graph
components = wmg.create.archetype.create_keisler_graph(
    coords=xy, return_components=True
)
wmg.save.to_neural_lam(components, "graph/keisler", hierarchical=False)

# Hierarchical graph
components = wmg.create.archetype.create_oskarsson_hierarchical_graph(
    coords=xy, return_components=True
)
wmg.save.to_neural_lam(components, "graph/hierarchical", hierarchical=True)
```

leifdenby:
I've spent some time now summarising the most important differences between the three pieces of prior work above:

| Aspect | Current `create_graph` CLI | `build_rectangular_graph.py` CLI | `@prajwal-tech07's code snippet` |
|---|---|---|---|
[...]

PR #591, opened 2026-04-11
test: match edge_index length to feature rows in graph artifacts

### LINKED ISSUE #589: test_graph_creation: assert edge_index and feature tensors agree on edge counts (per artifact)
Problem

tests/test_graph_creation.py currently checks that expected graph files exist and that each tensor has plausible shape (e.g. edge_index is [2, N], features have d_features columns). It does not verify that paired artifacts describe the same number of edges.

There is an explicit maintainer TODO in tests/test_graph_creation.py (around the block that loads required_graph_files):

    # TODO: check that the number of edges is consistent over the files, for
    # now we just check the number of features

If *_edge_index.pt and the matching *_features.pt ever drift (partial write, refactor bug, list-level mismatch in hierarchical mode), tests can still pass while training would later fail or silently mis-index edges.

Scope

For each logical pair below, assert N_edges from edge_index matches the first dimension of the corresponding features tensor:

- g2m_edge_index.pt <-> g2m_features.pt (single tensors)
- m2g_edge_index.pt <-> m2g_features.pt (single tensors)
- m2m_edge_index.pt <-> m2m_features.pt (list, one entry per level; check per list index)
- mesh_up_edge_index.pt <-> mesh_up_features.pt (hierarchical only; list length n_max_levels - 1)
- mesh_down_edge_index.pt <-> mesh_down_features.pt (same)

Where N_edges = edge_index.shape[1] and features.shape[0] should match.

Acceptance criteria

- Extend test_graph_creation (or a small helper) to enforce the pairwise consistency above for all graph_name x datastore_name combinations that are not skipped today.
- Assertion messages should name the failing file pair and print both counts when they differ.
- Do not change graph generation logic in the same PR unless a test exposes a real bug (if so, follow-up issue/PR).

Out of scope (separate issues)

- Saving generation metadata into the graph directory (provenance / config snapshot).
- plot_graph directed-edge visualization (create_graph.py has its own TODO there).

Suggested approach

After loading tensors, either load pairs in a second pass or collect edge_index and features per logical pair, then assert shape[1] vs shape[0] as above for each level in list-valued artifacts.

Shivampal157:
## Describe your changes

Adds assertions in `tests/test_graph_creation.py` so each `*_edge_index.pt` lines up with its `*_features.pt`: same number of edges as feature rows. For the list outputs (`m2m`, and `mesh_up` / `mesh_down` when hierarchical), it checks each level.

This was the old TODO in that test file - we already validated shapes file-by-file, but not that the two sides of a pair still match.

No new runtime deps.

## Issue Link

Closes #589

## Type of change

Pick what fits on your side — I’d call it **tests / maintenance**, not a user-visible feature.

- [x] Branch rebased/updated against `main` (or I can rebase again if needed)
- [x] Self-reviewed
- [x] Small helpers have short docstrings
- [ ] Extra inline comments (only where it gets weird — didn’t need much here)
- [ ] README — not needed for this
- [x] Tests added/extended (this *is* the test change)
- [x] PR title is imperative-ish

Reviewer/assignee: I don’t have write access on `mllam/neural-lam` — happy if someone from the team gets tagged.

## Checklist for reviewers

(leave for maintainers)

- [ ] CHANGELOG line if you want one — I can add under *maintenance* after review.

## Checklist for assignee

(leave for maintainers)

### DIFF
--- a/tests/test_graph_creation.py
+++ b/tests/test_graph_creation.py
@@ -13,28 +13,81 @@ from neural_lam.datastore.base import BaseRegularGridDatastore
 from tests.conftest import init_datastore_example
 
 
-def _assert_edge_rows_match_index(
-    edge_index: torch.Tensor, features: torch.Tensor, label: str
+def _ctx(datastore_name: str, graph_name: str, pair: str) -> str:
+    """Short prefix for assertion messages (CI-friendly)."""
+    return f"datastore={datastore_name!r}, graph={graph_name!r}, pair={pair}"
+
+
+def _assert_tensor_edge_feature_agree(
+    edge_index,
+    features,
+    *,
+    datastore_name: str,
+    graph_name: str,
+    pair: str,
+    d_features: int,
 ) -> None:
-    """One feature row per column in *edge_index* (one edge)."""
+    """One feature row per column in *edge_index* (one edge); validate types/shapes."""
+    c = _ctx(datastore_name, graph_name, pair)
+    assert isinstance(edge_index, torch.Tensor), (
+        f"{c}: edge_index must be torch.Tensor, got {type(edge_index).__name__}"
+    )
+    assert isinstance(features, torch.Tensor), (
+        f"{c}: features must be torch.Tensor, got {type(features).__name__}"
+    )
+    assert edge_index.dim() == 2, (
+        f"{c}: edge_index must be 2D, got dim={edge_index.dim()} "
+        f"shape={tuple(edge_index.shape)}"
+    )
+    assert edge_index.shape[0] == 2, (
+        f"{c}: edge_index must have 2 rows (COO), got shape[0]={edge_index.shape[0]}"
+    )
+    assert features.dim() == 2, (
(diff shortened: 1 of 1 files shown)


Saptami191:
@Shivampal157 fix is correct and covers all required pairs, including per-level checks for list artifacts.

Minor gaps:
Assertion messages lack `graph_name` and `datastore_name`  to  harder to debug in CI
No shape/type validation before comparison
 List inputs assumed without checks

overall its good minimal fix.

Shivampal157:
@Saptami191 Thanks for the review - updated: messages now include datastore + graph name, we validate tensor types/shapes before comparing counts, and list artifacts check list lengths before each level. Pushed to the branch

sadamov (dismissed):
Good tests, but the graph creation and loading will change soon in #384 and it makes more sense to add your tests to th related PRs directly.

Shivampal157:
Thanks for review, I am moving this into the requested related PR/issue

PR #301, opened 2026-03-01
Add edge count consistency check to graph creation tests

osten-antonio:
## Describe your changes

Implements the TODO in `test_graph_creation` that noted edge count consistency between paired graph files was not being checked.

Each edge set in the generated graph consists of two files:

- `*_edge_index.pt` — shape `[2, E]`, connectivity between nodes
- `*_features.pt` — shape `[E, 3]`, attributes for each edge

The previous test validated each file's feature dimension independently but never verified that both files described the same number of edges `E`. A mismatch would pass all existing assertions silently, but cause wrong message passing or an index-out-of-bounds error at training time.

Changes:
- Added `edge_index_feature_pairs` list declaring which index/feature files must match (including `mesh_up`/`mesh_down` for hierarchical graphs)
- Collected edge counts from both files during the existing shape-check loop
- Added a post-loop assertion over all pairs with a descriptive failure message
- Removed the `# TODO` comment as it is now implemented

No new dependencies required.

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

(This is a test improvement - none of the above categories apply exactly)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [ ] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [x] the code is readable
- [x] the code is well tested
- [x] the code is documented (including return types and parameters)
- [x] the code is easy to maintain

[...]

PR #123, opened 2025-02-11
Fix nans in gradient from inverse softplus

SimonKamuk:
## Describe your changes

Fix a bug where the inverse softplus causes gradients to become nans.

I believe the issue was that even though the `non_linear_part` had clamping inside, and the values were later being discarded if they were above the threshold, the input x values could still cause numerical instabilities before we checked if the input was above the threshold.

What I did to fix it was to clamp the input x values. The limits were previously defined as $exp(x\cdot beta)-1 \ge 10^{-6}$ and $x\cdot beta \le threshold$, so I changed it to the equivalent $log(10^{-6}+1)/beta \le x \le threshold/beta$.

## Issue Link

closes #119

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [ ] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [ ] I have performed a self-review of my code
- [ ] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [ ] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented (including return types and parameters)
- [ ] the code is easy to maintain

## Author checklist after completed review

- [ ] I have added a line to the CHANGELOG describing this change, in a section
  reflecting type of change (add section where missing):
  - *added*: when you have added new functionality
  - *changed*: when default behaviour of the code has been changed
  - *fixes*: when your contribution fixes a bug

## Checklist for assignee

[...]

PR #596, opened 2026-04-13
Add create_graph_with_wmg.py CLI using weather-model-graphs

### LINKED ISSUE #384: Supporting tensor-on-disk-format from weather-model-graphs

prajwal-tech07:
## Describe your changes

Add a new CLI script `create_graph_with_wmg.py` that delegates graph creation to [weather-model-graphs (https://github.com/mllam/weather-model-graphs) (wmg) and saves the output using `wmg.save.to_neural_lam()` in the tensor-on-disk format expected by `neural_lam.utils.load_graph()`.

This is the **neural-lam side** of the bridge described in #384. The wmg side is [mllam/weather-model-graphs#123 (https://github.com/mllam/weather-model-graphs/pull/123), which adds `to_neural_lam()` to `wmg.save`.

**What this PR does:**
- Adds `neural_lam/create_graph_with_wmg.py` with support for all three wmg archetypes: `keisler` (flat single-scale),  `graphcast` (flat multiscale), and `hierarchical` (Oskarsson)
- Auto-computes `mesh_node_distance` from grid spacing when not specified
- Reshapes datastore coordinates from `(Nx, Ny, 2)` → `(N, 2)` for wmg
- Adds a deprecation warning to the old `create_graph.py` CLI pointing users to the new script
- Adds `weather-model-graphs[pytorch]>=0.3.0` as a dependency
- Relaxes `torch-geometric` pin from `==2.3.1` to `>=2.5.3` (required by wmg's pytorch extra)
- Registers `create_graph_with_wmg` as a console script entry point

**Dependencies:** Requires [mllam/weather-model-graphs#123](https://github.com/mllam/weather-model graphs/pull/123) to be merged and released first (adds `wmg.save.to_neural_lam()`).

**Files changed (4 files, +305 −1):**
| File | Change |
|------|--------|
| `neural_lam/create_graph_with_wmg.py` | New CLI script (+188 lines) |
| `neural_lam/create_graph.py` | Added deprecation warning (+9 lines) |
| `pyproject.toml` | Added wmg dep, relaxed torch-geometric, added script entry (+5 −1) |
| `tests/test_graph_creation.py` | 12 new tests (9 wmg creation + 3 deprecation) (+103 lines) |

## Issue Link

Solves #384 (neural-lam side)

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


prajwal-tech07:
Merged Leif's `pyproject.toml` fix from #1. The pip, cpu CI failure is expected since `tool.uv.sources` isn't read by pip — only the uv jobs will correctly resolve `weather-model-graphs` from the fork branch. The 3 cancelled checks (pip gpu, uv cpu, uv gpu) were just cascading from the pip,cpu failure. All 21 tests pass locally.

neural_lam/create_graph_with_wmg.py
```diff
@@ -0,0 +1,188 @@
...
+from .datastore.base import BaseRegularGridDatastore
+
+ARCHETYPE_FUNCTIONS = {
+    "keisler": wmg.create.archetype.create_keisler_graph,
+    "graphcast": wmg.create.archetype.create_graphcast_graph,
+    "hierarchical": wmg.create.archetype.create_oskarsson_hierarchical_graph,
+}
+
+
+def _estimate_mesh_node_distance(xy):
```
leifdenby: maybe we could call this `_estimate_grid_node_spacing`? And then expose the grid-mesh-spacing ratio as a CLI arg that defaults to `3.0`?

neural_lam/create_graph_with_wmg.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+
+    if archetype not in ARCHETYPE_FUNCTIONS:
+        raise ValueError(
+            f"Unknown archetype '{archetype}'. "
+            f"Must be one of: {list(ARCHETYPE_FUNCTIONS.keys())}"
+        )
+
+    xy = datastore.get_xy(category="state", stacked=False)
+
```
leifdenby: The `stacked=True` argument is exactly for this situation where you just want a `grid_index` dimension rather than `x,y` Cartesian coordinates. Also, 

```suggestion
    xy = datastore.get_xy(category="state", stacked=True)

```

then you can remove the reshaping below

neural_lam/create_graph_with_wmg.py
```diff
@@ -0,0 +1,188 @@
...
+        xy = xy.reshape(-1, 2)
+
+    if mesh_node_distance is None:
+        mesh_node_distance = _estimate_mesh_node_distance(xy)
+
+    # Build keyword arguments for the archetype function
+    archetype_kwargs = dict(
+        coords=xy,
+        mesh_node_distance=mesh_node_distance,
+        return_components=True,
```
leifdenby, line 114: I think we need to add a comment here about why `return_components` must be `True` here. I thought we had fixed this issue, is it still the case that we must return the graph as these three subgraphs?

tests/test_graph_creation.py
```diff
@@ -117,3 +121,102 @@ def test_graph_creation(datastore_name, graph_name):
                         assert r.shape[0] == 2  # adjacency matrix uses two rows
                     elif file_id.endswith("_features"):
                         assert r.shape[1] == d_features
+
+
+@pytest.mark.parametrize("archetype", ["keisler", "graphcast", "hierarchical"])
+@pytest.mark.parametrize("datastore_name", DATASTORES.keys())
+def test_wmg_graph_creation(datastore_name, archetype):
```
leifdenby, line 352: This is great, but we could actually use the testing that https://github.com/mllam/neural-lam/pull/323 introduces. I should try and get that finished so that we can merge both in together :)

pyproject.toml
```diff
@@ -30,17 +30,21 @@ dependencies = [
     "matplotlib>=3.7.0",
     "plotly>=5.15.0",
     "torch>=2.3.0",
-    "torch-geometric==2.3.1",
+    "torch-geometric>=2.5.3",
```
leifdenby: is this a necessary change?

pyproject.toml
```diff
@@ -30,17 +30,21 @@ dependencies = [
...
     "mlflow>=2.16.2",
     "boto3>=1.35.32",
     "nvidia-ml-py>=13.580.82",
     "pillow>=9.0.0",
+    "weather-model-graphs",
 ]
 requires-python = ">=3.10"
 
+[project.scripts]
+create_graph_with_wmg = "neural_lam.create_graph_with_wmg:cli"
```
leifdenby, line 55: Can you add an entry to the README explaining how this new CLI works and suggest one should use that instead of the existing one?

.github/workflows/install-and-test.yml
leifdenby, line 1: I will make a separate PR for this. I think this is some general CI maintenance that should be merged in before this PR.

leifdenby (changes requested):
Looking good! A few suggestions for changes 🚀

neural_lam/create_graph_with_wmg.py
prajwal-tech07: Renamed _estimate_mesh_node_distance → `_estimate_grid_node_spacing` — it now returns only the average grid spacing. The ×3 multiplier is replaced by a new `grid_mesh_ratio` parameter (default 3.0) exposed as `--grid_mesh_ratio` on the CLI.

neural_lam/create_graph_with_wmg.py
prajwal-tech07: Switched to `stacked=True` and removed the manual reshaping — much cleaner, thanks for catching that.

neural_lam/create_graph_with_wmg.py
prajwal-tech07: Added an explanatory comment: `wmg.save.to_neural_lam()` expects separate g2m/m2g/m2m sub-graph components rather than a single merged graph.

tests/test_graph_creation.py
prajwal-tech07: Thanks, the tests are working well as-is! Happy to adapt them to the testing infrastructure from #323 once that's ready — makes total sense to merge them together. Let me know if there's anything I can do to help with #323!

pyproject.toml
prajwal-tech07: Yes, this is necessary. The weather-model-graphs[pytorch] extra requires `torch-geometric>=2.5.3` in its own dependencies, so we need to allow at least that version here too.

pyproject.toml
prajwal-tech07: Added a new "Graph creation with weather-model-graphs" section in the README, documenting the `create_graph_with_wmg` CLI with all three archetypes and available options. Also added a deprecation note to the old `create_graph` section.

.github/workflows/install-and-test.yml
prajwal-tech07: Noted — happy to revert the CI changes from this branch if you'd prefer to handle them in a separate PR.

pyproject.toml
leifdenby: ah, in that case I think we maybe should reduce required version number on the `weather-model-graphs` side, since all we are doing with `pytorch-geometric` is using it to convert the `networkx.DiGraph` objects to `torch.Tensor` objects, and we already do that in the current `create_graphs.py` code in `neural-lam` with the older version. So how about we change `weather-model-graphs` to require `torch-geometric==2.3.1` so the two are in sync? Then we don't have to change anything on the `neural-lam` side

neural_lam/create_graph_with_wmg.py
leifdenby: looks great, but it would better to name it `--grid_mesh_spacing_ratio` I think - just be clear about what it is the ratio between :) what do you think?

neural_lam/create_graph_with_wmg.py
leifdenby: thanks, I would like to remove the need for returning separate subgraphs in future, but let's leave this as-is for now

tests/test_graph_creation.py
leifdenby: ok, how about we say that depending on whether I finish #323 soon then we either a) merge this in with the tests have you have them implemented already or b) merge #323 in first and then adapting the testing here to use the testing being introduced in #323?

README.md
leifdenby, line 1: looks great!

pyproject.toml
leifdenby: thanks for adding this, looks great!

leifdenby (changes requested):
Thanks for this, I added additional comments on previous review to resolve:

- https://github.com/mllam/neural-lam/pull/596/files/d6da819cb6f27c9faa777da21fbb51791c6989d4#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711
- https://github.com/mllam/neural-lam/pull/596#discussion_r3085149525

neural_lam/create_graph_with_wmg.py
prajwal-tech07: `--grid_mesh_ratio` → `--grid_mesh_spacing_ratio`
Good call — renamed the CLI arg, function parameter, and docstrings throughout to make it clear it's the ratio between mesh-node and grid-node spacing.

pyproject.toml
prajwal-tech07: You're right — wmg only uses `from_networkx` which works fine with 2.3.1. I've also pushed a commit to the wmg PR branch (prajwal-tech07/weather-model-graphs@d0c693d) pinning it to ==2.3.1 there too, so both repos stay in sync.

tests/test_graph_creation.py
prajwal-tech07: Sounds like a great plan — either way works for me. Happy to adapt the tests to #323's infrastructure once it lands, or merge as-is if this PR is ready

leifdenby (changes requested):
Ok, the CLI interface looks great now, as do the readme instructions and install works great (just tried locally on my laptop too).

I think we should also change the tests that use graph creation (i.e. the training examples, they all use the `create_graph_from_datastore()` function in https://github.com/mllam/neural-lam/blob/main/neural_lam/create_graph.py#L540), but I think we should use the new one that you have created that creates a graph with `wmg` instead.

prajwal-tech07:
Thanks for the suggestion, @leifdenby! Done in 76c7ed7.

All test files that used the old `create_graph_from_datastore()` from `neural_lam.create_graph` have been migrated to use the new wmg-based version from `neural_lam.create_graph_with_wmg`:


- test_datasets.py, test_clamping.py, test_plotting.py, test_training.py: Switched to `create_graph_from_datastore(archetype="keisler")` (equivalent to the previous `n_max_levels=1` flat graph).
- `test_plot_graph.py`: Switched to wmg-based graph creation with keisler and `hierarchical` archetypes. Removed the multiscale (graphcast) parametrization from this test since the graphcast archetype produces multi-level m2m edges without up/down edges, which isn't compatible with `utils.load_graph()` (it infers `hierarchical=True` when `n_levels > 1` and then tries to load `mesh_up_edge_index.pt`). Graphcast graph creation is already separately tested in `test_graph_creation.py` where all 3 archetypes pass.

All tests pass locally (55/55 in the modified files, 72/72 in the full suite.

leifdenby:
I just completed a 200 epoch training (on a DGX Spark machine) on the DANRA test dataset using a `keisler` graph created with your new `create_graph_with_wmg` script and the training loss decreases as expected 🥳 

W B Chart 15 4 2026, 15 39 26

I ran the following commands:

```
uv run python -m neural_lam.create_graph_with_wmg --config_path tests/datastore_examples/mdp/danra_100m_winds/config.yaml --archetype keisler
uv run python -m neural_lam.train_model --config_path tests/datastore_examples/mdp/danra_100m_winds/config.yaml
```

So I will give this a final review, but I think this is nearly ready to merge :)

prajwal-tech07:
Hi @leifdenby! Implementation is complete. `create_graph_from_datastore()` in `neural_lam/create_graph_with_wmg.py` is working and tested , all 21 graph creation tests pass across 3 datastores × 3 archetypes (keisler, graphcast, hierarchical), including the deprecation warning test. The full training suite (6/6) also passes. Ready for your review!

prajwal-tech07:
Rebased onto `main` (now includes the merged graph-storage spec/validator from #323). Resolved conflicts, and added the end-to-end validation test from the merge plan: `test_wmg_graph_creation` now builds a graph via create_graph_with_wmg and validates it on disk with `docs/validate_graph.py`, asserting it passes the spec + `metainfo.yaml` version. Locally green (graph creation, datasets, plotting, and full training all pass) and `uv.lock` regenerated for the wmg #123-branch pin. Addressed in 5055f4d.

prajwal-tech07:
Rebased onto latest `main` and switched to the released weather-model-graphs 0.4.0 (now that #123 is merged and v0.4.0 is on PyPI), completing step 3 of the merge plan.

Changes since the last review:

- `pyproject.toml`: `weather-model-graphs>=0.4.0`, and the temporary `[tool.uv.sources]` git pin to the #123 branch is removed
- `uv.lock`: regenerated, resolves 0.4.0 from PyPI (no git references left)
- Added the end-to-end validation test discussed above (option b): `test_wmg_graph_creation` builds a graph through `create_graph_with_wmg` and validates it on disk with `docs/validate_graph.py` from #323, and asserts the metainfo.yaml spec version
- Added a CHANGELOG entry

 **Verification**: the full test suite passes locally against the released 0.4.0 - 259 passed, including graph creation for all three archetypes, the validation test, and a full training run. pre-commit is clean.

**One note on the dependency**: I deliberately did not use the `[pytorch]` extra. wmg 0.4.0's `[pytorch]` extra requires `torch-geometric>=2.5.3`, which is unsatisfiable against neural-lam's `torch-geometric==2.3.1` pin. It isn't needed here - `save/neural_lam/torch_tensors.py` doesn't import torch-geometric (`save/base.py` only imports it optionally via `HAS_PYG`), and neural-lam already provides torch/torch-geometric itself. Let me know if you'd rather pin `==0.4.0` exactly.

Your earlier review is still marked as changes requested - I believe everything from it is addressed now (`_estimate_grid_node_spacing` + `--grid_mesh_spacing_ratio`, `stacked=True`, the `return_components` comment, the README section, and CI left for your separate PR), so this should be ready for a re-review.

prajwal-tech07:
@leifdenby gentle ping on this one. It's green and mergeable - the bump to weather-model-graphs>=0.4.0 and the regenerated uv.lock went in on 28 July, so it now resolves 0.4.0 straight from PyPI with no fork refs.

Going by the merge order we agreed: spec #323 merged, the validation test is on this branch, wmg #123 merged and v0.4.0 released - this PR is the last remaining step. The points from your April review are all addressed. Is there anything else you'd like changed before it can go in?

neural_lam/create_graph_with_wmg.py
```diff
@@ -0,0 +1,201 @@
...
+    datastore : BaseRegularGridDatastore
+        Datastore providing grid coordinates.
+    output_root_path : str
+        Directory where the .pt graph files will be saved.
+    archetype : str
+        Graph archetype to create: ``"keisler"``, ``"graphcast"``, or
+        ``"hierarchical"``.
+    mesh_node_distance : float or None
+        Distance between created mesh nodes (in coordinate units). If None,
+        automatically estimated as ``grid_mesh_spacing_ratio * grid_spacing``.
```
leifdenby: what is `grid_spacing` here? I think it is the grid node spacing, but we should state here that we automatically try to estimate this.

neural_lam/create_graph_with_wmg.py
```diff
@@ -0,0 +1,201 @@
...
+            f"Unknown archetype '{archetype}'. "
+            f"Must be one of: {list(ARCHETYPE_FUNCTIONS.keys())}"
+        )
+
+    xy = datastore.get_xy(category="state", stacked=True)
+    xy = np.array(xy)
+
+    if mesh_node_distance is None:
+        grid_spacing = _estimate_grid_node_spacing(xy)
+        mesh_node_distance = grid_spacing * grid_mesh_spacing_ratio
```
leifdenby: I think we should log here that because `mesh_node_distance` wasn't provided we estimate the grid-node spacing and use that. Maybe something like `mesh_node_distance not provided so estimating grid-node spacing, with value estimated to be xx.xx, resulting mesh-node spacing will be yy.yy` - what do you think?

neural_lam/create_graph_with_wmg.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+    parser.add_argument(
+        "--archetype",
+        type=str,
+        default="keisler",
+        choices=["keisler", "graphcast", "hierarchical"],
+        help="Graph archetype to create",
+    )
+    parser.add_argument(
+        "--mesh_node_distance",
```
leifdenby, line 160: I think it is a bit confusing that we use both term "spacing" and "distance" for the mesh/grid resolution. I appreciate that wmg uses "distance", so maybe should use the same here? "resolution" would probably be an even better term. What do you think?

tests/test_graph_creation.py
leifdenby: we can remove these tests now, no? Maybe we should instead just call the validator from https://github.com/mllam/neural-lam/pull/323 here?

tests/test_plot_graph.py
```diff
@@ -30,23 +35,20 @@ def graph_fixture(request, tmp_path_factory):
     datastore = DummyDatastore()
 
     if graph_name == "hierarchical":
-        hierarchical = True
-        n_max_levels = 3
-    elif graph_name == "multiscale":
```
leifdenby, line 35: why did you remove the multiscale option?

CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add latent encoder/decoder modules and the `GraphEFM` (hierarchical) / `GraphEFMMultiScale` (flat) step predictors for the Graph-EFM ensemble forecasting model. [\#648](https://github.com/mllam/neural-lam/pull/648) @Sir-Sloth-The-Lazy
 
+- Add `neural_lam.create_graph_with_wmg` CLI which builds graphs with
```
leifdenby: this is a bit verbose, can you shorten it to just a single sentence?

leifdenby (changes requested):

leifdenby:
> **One note on the dependency**: I deliberately did not use the `[pytorch]` extra. wmg 0.4.0's `[pytorch]` extra requires `torch-geometric>=2.5.3`, which is unsatisfiable against neural-lam's `torch-geometric==2.3.1` pin

Ah, I hadn't realised this. There won't be anything in `weather-model-graphs` that actually needs a more recent version. But maybe we can just make `neural-lam` require `torch-geometric>=2.3.1` instead? That way we can keep moving without having to release a new revision of `weather-model-graphs` with a more lax requirement

neural_lam/create_graph_with_wmg.py
prajwal-tech07: Yeah, Leif - that's me mixing the two, not wmg. I had a look and there's no convention to protect either way: "spacing" doesn't appear anywhere in the codebase outside this file, and "resolution" only turns up in a comment in `create_graph.py` and a cartopy `resolution="50m"` kwarg. So we can just pick one.

I'd go with "distance". wmg's parameter is `mesh_node_distance` and we pass it straight through, so anything else means the flag and the thing it sets end up with different names for the same quantity. And "resolution" in a weather context usually runs the other way - higher resolution is a smaller number - so `mesh_resolution=5` could be read as 5 units apart or 5x finer, where a distance of 5 can only mean one thing.

That would give `mesh_node_distance` (unchanged), `_estimate_grid_node_distance`, and `grid_spacing` -> `grid_node_distance`.

One thing I'd like your view on: `--grid_mesh_spacing_ratio` reads as `grid:mesh`, but the code is `mesh_node_distance = grid_spacing * ratio`, so it's mesh:grid. At the default of 3.0 the mesh nodes end up 3x further apart, not 3x closer. If we're renaming anyway, `--mesh_grid_distance_ratio` would match, though I don't know if churning a CLI flag is worth it.

tests/test_plot_graph.py
prajwal-tech07: I removed it because at the time it genuinely didn't work. Back then `to_torch_tensors_on_disk` split non-hierarchical m2m by the `level` edge attribute, so graphcast came out as a multi-level list. `load_graph` infers `hierarchical = len(m2m_edge_index) > 1`, so it read a flat graph as hierarchical and went looking for `mesh_up_*.pt` files that aren't there.

Worth being explicit that this is GC-LAM from the README's three paper graphs - it should absolutely be covered here, especially in a PR that deprecates `create_graph.py`.

That's no longer true. When I conformed the function to the storage spec from #323, non-hierarchical graphs started being written as a single merged m2m level, and that's what's in 0.4.0. The note I left in the fixture is just stale - I should have updated it then.

Checked against the released 0.4.0:

`keisler       m2m levels: 1  (8320 edges)   mesh_up: no    load_graph -> hierarchical=False`
`graphcast     m2m levels: 1  (6096 edges)   mesh_up: no    load_graph -> hierarchical=False`
`hierarchical  m2m levels: 3  (5512/544/40)  mesh_up: yes   load_graph -> hierarchical=True`

So it loads fine now, and I've put `multiscale` back as the graphcast archetype. All three are green (6 passed).

CHANGELOG.md
prajwal-tech07: Done, shortened to a single sentence.

neural_lam/create_graph_with_wmg.py
prajwal-tech07: Yes, it's the grid node spacing - roughly the distance between neighbouring grid points, estimated from the datastore coordinates when `mesh_node_distance` isn't given. Agreed the docstring should say that outright instead of leaving it to be inferred.

I'll fold this in together with the rename once we've settled the terminology in the other thread, since the wording depends on whether we land on "distance" or "resolution". Same goes for the log line you suggested.

prajwal-tech07:
>But maybe we can just make neural-lam require torch-geometric>=2.3.1 instead?

I tried it. It works, but it's a bigger change than it looks, so I'd rather not fold it into this PR.
The conflict is real, uv won't resolve the extra against the 2.3.1 pin. Loosening to `>=2.3.1` does resolve, but look at what it resolves to: `torch-geometric 2.3.1 -> 2.8.0.post1`. That isn't optional either. The extra needs `>=2.5.3`, so taking it means leaving 2.3.1 behind.

There's also a trap if we do it separately. Loosening the pin without the extra looks like a no-op, because `uv lock` keeps an already-locked version when it still satisfies the constraint. The lockfile stays on 2.3.1 and CI stays green. Then the next `uv lock --upgrade` jumps to 2.8.0. I'd rather that happened in a PR where someone's watching for it. 2.3.1 is from April 2023, 2.8.0 came out last month, five minor versions apart, and `InteractionNet` subclasses `pyg.nn.MessagePassing`, so it needs testing rather than just a green resolve.

Either way this PR doesn't need the extra. `save/neural_lam/torch_tensors.py` imports only `torch`, and `save/base.py` keeps its `torch_geometric` import behind the `HAS_PYG` guard, so plain `weather-model-graphs>=0.4.0` covers everything `create_graph_with_wmg` uses, and neural-lam brings its own torch and torch-geometric anyway. So I'd leave this as it is and do the bump separately. Can open an issue for it if you want.

neural_lam/create_graph_with_wmg.py
prajwal-tech07: Yes Leif, this should be logged. It's the one number in the command the user didn't pick, so it shouldn't be silent.

One thing on the wording. Your version prints the estimated grid spacing and the resulting mesh spacing, but not the ratio between them, and the ratio is what someone would reach for if the number looks wrong. I'd rather the line explained itself:

`mesh_node_distance not given; estimated grid node spacing 12.50 x grid_mesh_spacing_ratio 3.0 -> mesh_node_distance 37.50`
(parameter name rather than the CLI flag, since `create_graph_from_datastore` gets called directly from the tests too)

The other reason I want it visible: the estimate is rougher than "estimated" suggests. It's `sqrt(x_range * y_range / N)`, so it returns the geometric mean of dx and dy, which on an anisotropic grid lands between the two rather than on either. And since `np.ptp` measures n-1 intervals across n points, it comes out low by `sqrt((nx-1)(ny-1)/(nx*ny))`. I measured -0.2% on 500x500, -1% on 100x100, -10% on 10x10. Doesn't matter on a real domain, very visible on a small test one, which is exactly when someone is squinting at a graph wondering why it looks off.

The file has no logger yet, so I'd add loguru at info to match `create_graph.py`. Happy to make it a warning instead.
Wording depends on the naming thread, by the way. If we go with "distance" this becomes estimated grid node distance.

leifdenby: sounds great! Let's include the ratio as well, good idea

neural_lam/create_graph_with_wmg.py
leifdenby: > I'd go with "distance". wmg's parameter is `mesh_node_distance` and we pass it straight through, so anything else means the flag and the thing it sets end up with different names for the same quantity. And "resolution" in a weather context usually runs the other way - higher resolution is a smaller number - so `mesh_resolution=5` could be read as 5 units apart or 5x finer, where a distance of 5 can only mean one thing.

I agree - good point!

> One thing I'd like your view on: `--grid_mesh_spacing_ratio` reads as `grid:mesh`, but the code is `mesh_node_distance = grid_spacing * ratio`, so it's mesh:grid. At the default of 3.0 the mesh nodes end up 3x further apart, not 3x closer. If we're renaming anyway, `--mesh_grid_distance_ratio` would match, though I don't know if churning a CLI flag is worth it.

Very well caught! Nice, yes rename the arg

neural_lam/create_graph_with_wmg.py
leifdenby: let's go with `distance` here like you suggest :)

tests/test_plot_graph.py
leifdenby: perfect!

leifdenby:
> Either way this PR doesn't need the extra. `save/neural_lam/torch_tensors.py` imports only `torch`, and `save/base.py` keeps its `torch_geometric` import behind the `HAS_PYG` guard, so plain `weather-model-graphs>=0.4.0` covers everything `create_graph_with_wmg` uses, and neural-lam brings its own torch and torch-geometric anyway. So I'd leave this as it is and do the bump separately. Can open an issue for it if you want.

Ah yes, that's because we have `pytorch` and optional dependency in `weather-model-graphs`. Ok, yes if you could make an issue for `weather-model-graphs` to say we should relax the min version of `pytorch` required. And we should put that on the roadmap for the next release of `weather-model-graphs` I think

neural_lam/create_graph_with_wmg.py
prajwal-tech07: Done in 8665d28 - the docstring now says the grid node distance is estimated automatically from the grid coordinates and multiplied by `mesh_grid_distance_ratio`.

neural_lam/create_graph_with_wmg.py
prajwal-tech07: Added in 8665d28, with the ratio in it:

`mesh_node_distance not given; estimated grid node distance 5000.00 x mesh_grid_distance_ratio 3.0 -> mesh_node_distance 15000.00`

loguru at info. Only fires when `mesh_node_distance` isn't given.

neural_lam/create_graph_with_wmg.py
prajwal-tech07: Renamed in 8665d28, both the parameter and the CLI flag: `--grid_mesh_spacing_ratio` -> `--mesh_grid_distance_ratio`. Also `_estimate_grid_node_spacing` -> `_estimate_grid_node_distance` and the local `grid_spacing` -> `grid_node_distance`, so "spacing" is gone from the file. Nothing to deprecate since `create_graph_with_wmg` hasn't been in a release yet, so nobody has typed the old flag.

prajwal-tech07:
Opened mllam/weather-model-graphs#149 for the torch-geometric floor. One thing that came out of writing it up: the `pytorch` extra doesn't look like it's ever installed in CI, so that minimum has never actually been exercised. I flagged it in the issue as a possible follow-up rather than folding it in.

That's everything from the review addressed now, all in 8665d28, and CI is green. Ready for another look!!

tests/test_graph_creation.py
leifdenby: I still think we should remove anything that relates to required files (filenames) and contents of these files from the tests in `tests/` to instead use the functionality in the validator for testing what is there or not. So I would delete anything in the `test_graph_creation()` function that isn't about 1) constructing a graph or 2) validating that stored graph with the validator

sadamov:
@prajwal-tech07 I just merged #673 so it is now required to use static typing in every PR incl. #596. You can use pre-commit hooks to check using ty.

tests/test_graph_creation.py
prajwal-tech07: Done in a256938 and 6f4d06c. `test_graph_creation()` is now just building the graph and running the validator, and I dropped the `mesh_up`/`mesh_down` file check from `test_wmg_graph_creation()` too, so both are down to construct + validate. 137 lines out. The edge count consistency check from #301 goes with it - the validator covers that one.

I kept this to `tests`/`test_graph_creation.py `- of the other files that touch filenames, `test_validate_graph_script.py` has to write and corrupt files because it's testing the validator itself, and `test_graph_version_compatibility.py` is about the on-disk format versions by design.

prajwal-tech07:
Thanks @sadamov  for the heads-up. Added type hints in eb9ba91 - `create_graph_with_wmg.py` is fully annotated now and ty passes clean.

leifdenby (approved):
I have given this a final full read through and just ran graph creation locally on a clean checkout of the code. Everything worked flawlessly for me 🥳 

THANK YOU for the mammoth work you did here @prajwal-tech07 🌟 This a really fantastic contribution that produces the first interface between `neural-lam` and `weather-model-graphs`. And thank for your patience through all this. This was a very tricky one to get right, but I am very happy with the result we ended with. Now to see what people build with it! 🚀

PR #604, opened 2026-04-15
Use CPU/GPU extras with uv sources for torch installation

### LINKED ISSUE #600: Use CPU/GPU extras to simplify torch installation
The CI comment explains the problem well:

> *"We don't use lock files in neural-lam, because that would tie us to a particular variant of pytorch (gpu/cpu)"*

The current workaround is a pip dry-run to detect the torch version, then a separate install from the right index URL. It works but it's fragile and not something contributors can replicate easily locally.

The fix is to declare `cpu` and `gpu` as proper optional extras and route torch to the correct PyTorch index via `[tool.uv.sources]`:

```toml
[project.optional-dependencies]
cpu = ["torch>=2.3.0"]
gpu = ["torch>=2.3.0"]

[tool.uv]
conflicts = [[{ extra = "cpu" }, { extra = "gpu" }]]

[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true

[[tool.uv.index]]
name = "pytorch-gpu"
url = "https://download.pytorch.org/whl/cu128"
explicit = true

[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", extra = "cpu" },
  { index = "pytorch-gpu", extra = "gpu" },
]
```

The CI install steps then collapse to:

```bash
uv sync --extra cpu --group dev   # CPU runners
uv sync --extra gpu --group dev   # GPU runners
```

And contributors get a clear install story:

```bash
uv sync --extra cpu   # local development
uv sync --extra gpu   # GPU machine
```

The pip dry-run hack and the manual `--index-url` steps can be removed entirely. The `conflicts` declaration also prevents accidentally mixing CPU and GPU torch builds in one environment.

RajdeepKushwaha5:
## Describe your changes

Move `torch` from direct `[project] dependencies` to `[project.optional-dependencies]` with mutually exclusive `cpu` and `gpu` extras. Add `[tool.uv.sources]`, `[[tool.uv.index]]`, and `[tool.uv] conflicts` so that `uv sync --extra cpu` (or `--extra gpu`) installs torch from the correct PyTorch index in a single command.

Simplify the uv CI path from manual `uv venv` + `uv pip install torch` + `uv pip install -e .` to `uv sync --extra ${{ matrix.device }} --group dev`. The pip CI path retains the dry-run hack since pip cannot read `[tool.uv.sources]`. README install instructions updated accordingly.

## Issue Link

closes #600

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

  - *maintenance*: added entry under Maintenance in CHANGELOG.md


sadamov (changes requested):
This looks good overall. The extras + `[tool.uv.sources]` approach is the right solution for `uv sync`. A few things to fix:

This PR has a big! side-effect, changing the cuda version from 11.1 to 12.8 during install instructions. This should be made very clear to everyone and also discussed at a future dev-meeting.

pyproject.toml
```diff
@@ -29,7 +29,6 @@ dependencies = [
     "tueplots>=0.0.8",
     "matplotlib>=3.7.0",
     "plotly>=5.15.0",
-    "torch>=2.3.0",
     "torch-geometric==2.3.1",
```
sadamov, line 33: **`torch` must stay in core dependencies (CRITICAL)**

Removing `torch` from `[project.dependencies]` means `pip install neural-lam` (or `uv sync` without an extra) produces a runtime `ImportError`. The extras should control the *variant*, not make torch optional.

```suggestion
    "plotly>=5.15.0",
    "torch>=2.3.0",
    "torch-geometric==2.3.1",
```

pyproject.toml
```diff
@@ -41,9 +40,32 @@ dependencies = [
...
 
+[project.optional-dependencies]
+cpu = ["torch>=2.3.0"]
+gpu = ["torch>=2.3.0"]
+
 [dependency-groups]
 dev = ["pre-commit>=3.8.0", "pytest>=8.3.2", "pooch>=1.8.2"]
 
+[tool.uv]
+conflicts = [[{ extra = "cpu" }, { extra = "gpu" }]]
```
sadamov: Worth noting here that `--torch-backend` is not available for `uv sync`, so readers understand why the extras approach is used rather than a simpler flag.

```suggestion
[tool.uv]
# uv-specific configuration for PyTorch CPU/GPU variant selection.
# NOTE: --torch-backend is only supported in `uv pip`, not `uv sync`.
# Tracking issue for project-interface support: https://github.com/astral-sh/uv/issues/12994
# Until then, we use extras + explicit index routing for `uv sync`.
conflicts = [[{ extra = "cpu" }, { extra = "gpu" }]]
```

.github/workflows/install-and-test.yml
```diff
@@ -1,23 +1,18 @@
...
-# first install a specific torch version (and variant), and then install neural-lam.
-# To ensure that the version of torch installed is compatible with neural-lam we
-# first use `pip install --dry-run .` to determine the version of torch that would
-# be installed.
+# pip does not read [tool.uv.sources], so the pip path still uses a dry-run to
+# detect the torch version and installs from the correct index URL manually.
+#
+# We use two sets of hosts:
+# - ubuntu-latest for CPU (GitHub-hosted)
+# - cirun-aws-runner--${{ github.run_id }} for GPU (AWS via cirun.io)
```
sadamov: The existing header is inaccurate: it says torch is an *optional* dependency (`torch is declared as an optional dependency`). The correct framing is that torch is required and the extras select the variant. Also adds the `--torch-backend` note and pip routing explanation.

```suggestion
# torch is a required dependency in [project.dependencies]. The CPU/GPU variant
# is selected at install time via extras defined in pyproject.toml:
#
#   uv sync --extra cpu --group dev  # CPU runners
#   uv sync --extra gpu --group dev  # GPU runners (CUDA 12.8)
#
# The extras route torch to the correct PyTorch index via [tool.uv.sources]
# and [[tool.uv.index]] in pyproject.toml.
#
# NOTE: `uv sync --torch-backend` is not yet supported (uv pip only).
# Tracking: https://github.com/astral-sh/uv/issues/12994
#
# pip does not support extras-based index routing. The pip path uses a dry-run
# to detect the required torch version and installs from the correct index URL.
#
# To test both CPU and GPU installations with both UV and pip we use two sets
# of hosts to run the training on:
# - ubuntu-latest for CPU: these are on github
# - cirun-aws-runner--${{ github.run_id }} for GPU: these are on AWS,
#   orchestrated by cirun.io
```

.github/workflows/install-and-test.yml
```diff
@@ -62,28 +57,24 @@ jobs:
       - name: Determine PyTorch version for neural-lam (using pip dry-run)
         if: matrix.package_manager == 'pip'
         run: |
-          TORCH_VERSION=$(python -m pip install --dry-run "." | grep "Would install" | grep -o 'torch-[0-9.]*' | awk -F'-' '{print $2}' | tail -n 1)
+          TORCH_VERSION=$(python -m pip install --dry-run ".[${{ matrix.device }}]" | grep "Would install" | grep -o 'torch-[0-9.]*' | awk -F'-' '{print $2}' | tail -n 1)
```
sadamov: `pip` ignores `[tool.uv.sources]` entirely, so `".[cpu]"`/`".[gpu]"` has no effect on which torch variant the dry-run resolves. With `torch` back in core deps the extra is also unnecessary.

```suggestion
          TORCH_VERSION=$(python -m pip install --dry-run "." | grep "Would install" | grep -o 'torch-[0-9.]*' | awk -F'-' '{print $2}' | tail -n 1)
```

README.md
```diff
@@ -70,10 +70,13 @@ keeping versions consistent (it automatically updates the `pyproject.toml`
...
-variants for different CUDA versions and cpu-only support you will need to install
-`torch` separately if you don't want the most recent GPU variant that also
-expects the most recent version of CUDA on your system.
+**regarding `torch` installation**: `torch` ships different package variants for
+CPU-only and different CUDA versions. `neural-lam` requires PyTorch in
+practice; the `cpu` and `gpu` extras are primarily used to select the correct
+PyTorch variant/index rather than to make `torch` unnecessary. When installing
+with `uv` the correct PyTorch index is selected automatically via
+`[tool.uv.sources]`. For `pip` you still need to install `torch` from the
+correct index URL manually before installing `neural-lam`.
```
sadamov: The existing text still implies torch is optional (`primarily used to select the correct PyTorch variant/index rather than to make torch unnecessary`). The paragraph should make clear `torch` is a hard requirement.

```suggestion
**regarding `torch` installation**: `torch` is a required dependency, but it ships separate variants for CPU-only and each CUDA version. With `uv`, the correct PyTorch index is chosen automatically via `[tool.uv.sources]`. With `pip`, you must pass `--index-url` explicitly (see instructions below).
```

README.md
```diff
@@ -90,16 +93,18 @@ python -m pip install neural_lam
...
-3. Create a virtual environment for uv to use with `uv venv --no-project`.
-4. Install a specific version of `torch` with `uv pip install torch --index-url https://download.pytorch.org/whl/cpu` for a CPU-only version or `uv pip install torch --index-url https://download.pytorch.org/whl/cu111` for CUDA 11.1 support (you can find the correct URL for the variant you want on [PyTorch webpage](https://pytorch.org/get-started/locally/)).
-5. Install the dependencies with `uv pip install .`. If you will be developing `neural-lam` we recommend to install the development dependencies with `uv pip install --group dev -e .`. This installs the `neural-lam` package in editable mode, so you can make changes to the code and see the effects immediately.
+3. Install with the CPU or GPU extra:
+   ```bash
+   uv sync --extra cpu --group dev   # CPU-only
+   uv sync --extra gpu --group dev   # GPU (CUDA 12.8)
+   ```
+   This creates a virtual environment, installs `torch` from the correct PyTorch index, and installs all other dependencies (including dev tools) in one step.
```
sadamov: The description is hard to parse (one long line) and there is no guidance for users on a CUDA version other than 12.8. Please expand to:

```
3. Install with the CPU or GPU extra:
   ```bash
   uv sync --extra cpu --group dev  # CPU-only
   uv sync --extra gpu --group dev  # GPU (CUDA 12.8)
   ```
   This creates a virtual environment, installs `torch` from the correct
   PyTorch index, and installs all other dependencies (including dev tools)
   in one step.

   If you need a different CUDA version, swap torch after install:
   ```bash
   uv sync --extra gpu --group dev              # create env
   uv pip install torch --torch-backend auto    # replace torch to match your GPU
   ```
```

(Cannot use a `suggestion` block here due to nested triple-backtick fences.)

sadamov:
**Nit:** `name :` has a spurious space (line 46 of the updated workflow file). Since this line is between diff hunks it cannot be an inline suggestion, but it should be fixed:

```yaml
      - name: Install uv (if applicable)
```

pyproject.toml
sadamov: for some reason i cant get this inline suggestion to display properly, sorry

RajdeepKushwaha5:
Pushed `fb8f4eb` addressing all review comments. Summary of changes:

**`pyproject.toml`**
- Restored `torch>=2.3.0` to `[project.dependencies]` so `pip install neural-lam` (and `uv sync` without an extra) keeps torch as a hard requirement. The `cpu`/`gpu` extras now only select the variant.
- Added the `--torch-backend` / uv issue #12994 note above `[tool.uv]`.

**`.github/workflows/install-and-test.yml`**
- Rewrote the header comment per your suggestion (torch framed as required, extras select variant, `--torch-backend` note, pip routing explanation, hosts block preserved).
- Reverted the pip dry-run argument from `".[${{ matrix.device }}]"` back to `"."` — pip ignores `[tool.uv.sources]` and torch is in core deps now, so the extra is unnecessary.
- Fixed the `name :` whitespace on the `Install uv` step.

**`README.md`**
- Replaced the `**regarding torch installation**` paragraph with your shorter wording that makes the hard requirement explicit.
- Expanded the `uv` install steps with the post-install `uv pip install torch --torch-backend auto` recipe for users on a CUDA version other than 12.8.

**Commit message**
- Updated to reflect that torch stays in core deps (the previous wording said "Move torch from direct dependencies to optional-dependencies" which is no longer accurate).

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add comprehensive type hints to `neural_lam/create_graph.py` [\#618](https://github.com/mllam/neural-lam/pull/618) @GiGiKoneti
 
+- Use CPU/GPU optional-dependency extras with `[tool.uv.sources]` to simplify torch installation in CI and for users [\#600](https://github.com/mllam/neural-lam/issues/600) @RajdeepKushwaha5
```
sadamov: Please link the PR not the issue. Also, the CUDA default jumping from 11.1 to 12.8 should be called out explicitly in the changelog. It is essentially a breaking change for anyone on older drivers, so I think this PR deserves two separate entries.

```suggestion
- Use `cpu`/`gpu` extras with `[tool.uv.sources]` and a committed `uv.lock` to simplify and pin torch installation in CI and for users [\#604](https://github.com/mllam/neural-lam/pull/604) @RajdeepKushwaha5

- Update default CUDA version in install instructions and CI from 11.1 to 12.8. Users on older CUDA versions need to install the matching `torch` variant manually (see README) [\#604](https://github.com/mllam/neural-lam/pull/604) @RajdeepKushwaha5
```

.github/workflows/install-and-test.yml
```diff
@@ -66,24 +68,20 @@ jobs:
...
-      - name: Setup environment with uv (if applicable)
+      - name: Install with uv
         if: matrix.package_manager == 'uv'
         run: |
-          uv venv --no-project
-          source .venv/bin/activate
-          uv pip install torch --index-url ${{ matrix.torch_index }}
-          uv pip install --group dev -e .
-          uv pip list
+          uv sync --extra ${{ matrix.device }} --group dev
```
sadamov: I propose to do the lockfile-commit step in this PR rather than as a follow-up. I made the same point on #606: that PR's caching speedup only fully materialises once `uv.lock` is committed and the cache key is stable. Since this PR introduces the `[tool.uv.sources]` + `conflicts` machinery that makes a single lockfile cover both cpu and gpu, this is the natural place to land it.

**a) Generate and commit `uv.lock`.**

In a clean checkout, with `uv` installed:

```bash
uv lock
git add uv.lock
git commit -m "Add uv.lock for reproducible CPU/GPU installs"
```

`uv lock` only writes the lockfile, no `.venv` side-effects. With `[tool.uv] conflicts = [[{extra="cpu"}, {extra="gpu"}]]` already in place, the resolver forks and pins torch from both `pytorch-cpu` and `pytorch-gpu` indexes in a single file. Quick sanity check after generation: `grep -c 'pytorch-cpu\|pytorch-gpu' uv.lock` should be > 0 for both.

**b) Switch CI to `--locked`.**

```suggestion
          uv sync --extra ${{ matrix.device }} --group dev --locked
```

`--locked` (rather than `--frozen`) asserts that `uv.lock` is in sync with `pyproject.toml` and fails loudly if someone edits a dep without re-running `uv lock`. This is exactly the property that makes the cache key in #606 deterministic.

README.md
```diff
@@ -90,16 +87,26 @@ python -m pip install neural_lam
...
+   ```
+   This creates a virtual environment, installs `torch` from the correct
+   PyTorch index, and installs all other dependencies (including dev tools)
+   in one step.
+
+   If you need a different CUDA version, swap `torch` after install:
+   ```bash
+   uv sync --extra gpu --group dev              # create env
+   uv pip install torch --torch-backend auto    # replace torch to match your GPU
+   ```
```
sadamov: Two changes here. First, swap the `--extra gpu` in step one for `--extra cpu`: the CPU torch wheel is ~200MB vs ~2.5GB for cu128, so if the user already knows they don't want cu128, downloading it just to overwrite it is pointless. Second, once `uv.lock` is committed (see the workflow comment), any subsequent `uv sync` will reset torch to the variant pinned in the lockfile, so a one-liner caveat helps.

Cannot use a `suggestion` block here due to nested triple-backtick fences, please apply manually:

```
   If you need a different CUDA version, install with the `cpu` extra first (smaller download) and then swap `torch` to your CUDA variant:
   ```bash
   uv sync --extra cpu --group dev              # create env (small torch wheel)
   uv pip install torch --torch-backend auto    # replace torch with your CUDA variant
   ```
   Note: this swap is venv-local and will be reverted by the next `uv sync`.
```

sadamov (changes requested):
Thanks for the round of revisions in `fb8f4eb`, all my previous comments are addressed. Three things left, and I'd like all of them in this PR rather than as follow-ups so we land the lockfile + cache-key story coherently with #606:

1. CHANGELOG: link the PR, split into two entries to surface the CUDA bump.
2. Commit `uv.lock` and switch CI to `--locked`. This is the big one.
3. README: lead the CUDA-swap recipe with `--extra cpu` (smaller download) and clarify the swap is reverted by the next `uv sync`.

Details inline.

sadamov (approved):
This looks ready! the bug fixed in #639 remains fixed. Lock-file adds reproducibility layer and also allows for easier caching and faster CICD. 
Thanks a lot @RajdeepKushwaha5!

joeloskarsson (approved):
All looks good to me! Now I have not been involved with the original gh workflow setup, nor am particularly knowledgeable about it, so leaving it to you if you want to get an additional review from someone who knows it more. But I understand all these changes and think they are good.

observingClouds (changes requested):
LGTM, just the [AGENTS.md](https://github.com/mllam/neural-lam/blob/main/AGENTS.md#commands) needs an update about the new install instructions.

leifdenby:
Just had a read through and this looks great to me too 🌟 Thank you for implementing this @RajdeepKushwaha5! 🚀

observingClouds (approved):
LGTM, thanks for addressing my comments.

RajdeepKushwaha5:
thank you everyone for the detailed reviews and patience. I learned a lot through this PR, especially around uv, PyTorch installation flows, lockfiles, and CI reproducibility. I’m already working with CGAL for GSoC this year, but I’d genuinely love to stay involved with MLLAM as well if there’s any way I can continue contributing or help in some role outside GSoC.

Really appreciate the guidance from everyone here!

PR #606, opened 2026-04-15
Speed up CI by caching dependencies and virtual environment

### LINKED ISSUE #605: Speed up CI by caching dependencies and virtual environment
**Summary**
Right now, our CI reinstalls all dependencies from scratch on every run, which adds a lot of unnecessary time. We can speed things up significantly by caching both the downloaded packages (pip/uv) and the virtual environment.
This becomes especially important if we want to run notebook tests on every PR without slowing CI down too much.

**Proposed Solution**
We can introduce a two-layer caching approach:

1. Cache package downloads
Avoid re-downloading dependencies from PyPI on every run.
For pip:
```
- name: Set up Python
  uses: actions/setup-python@v5
  with:
    python-version: '3.13'
    cache: 'pip'
```

For uv:
```
- name: Install uv
  uses: astral-sh/setup-uv@v1
  with:
    enable-cache: true
2. Cache the virtual environment (.venv)
```

Cache the installed environment so we only reinstall when dependencies actually change. The cache key should be based on uv.lock, not pyproject.toml. As discussed on Slack, hashing pyproject.toml isn’t reliable, dependencies can change upstream without any changes to that file. Using a lockfile ensures the cache is tied to exact, reproducible dependency versions.

Config:
```
- name: Cache virtual environment
  uses: actions/cache@v4
  with:
    path: .venv
    key: venv-${{ runner.os }}-${{ hashFiles('uv.lock') }}
```

Blockers

[#600](https://github.com/mllam/neural-lam/issues/600): This depends on resolving the Torch installation workaround. We can’t generate or commit a stable uv.lock until that’s fixed.

Sharkyii:
## Describe your changes

Adds smart two-layer caching to CI by storing both downloaded packages and the full virtual environment, so builds don’t start from scratch every time.
This significantly reduces install time by reusing dependencies when nothing has changed, while still staying reliable with lock-based cache keys.
Overall, it speeds up CI runs dramatically (up to ~90%) without changing any existing functionality or tests.

## Issue Link

#605 

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

closes #605


sadamov:
Thanks @Sharkyii 90% faster sounds fantastic! Currently there seems to be some issue with the clean-up of the .venv? 
From the failing action above:
```
Run uv venv --no-project
Using CPython 3.13.13 interpreter at: /opt/hostedtoolcache/Python/3.13.13/x64/bin/python
Creating virtual environment at: .venv
error: Failed to create virtual environment
  Caused by: A virtual environment already exists at `.venv`. Use `--clear` to replace it
Error: Process completed with exit code 2.
```

archit7-beep:
Hi! I took a look at the current approach and noticed a potential issue around `.venv` handling on cache hits.

Right now, even though the workflow skips venv creation in one place, there are still cases where `uv venv --no-project` may be invoked again (directly or indirectly), which can cause failures when `.venv` is already restored from cache.

A more robust approach would be to ensure that venv creation happens strictly in one place and is fully guarded by the cache-hit condition, so that on cache hits we only reuse the restored environment and avoid any reinitialization.

I experimented with this in a separate PR: #610  — it demonstrates a cache-aware setup where venv creation is skipped entirely on cache hits and the existing environment is reused safely.

Happy to adapt this approach here — if you're okay with it, I can also contribute the changes directly to this branch.
@Sharkyii @sadamov

Sharkyii:
@archit7-beep thanks :) , but currently i am still working on this.
Once i am done you can review it again.

Sharkyii:
@sadamov @archit7-beep 
Please have a look!

archit7-beep:
This is definitely an improvement over the previous version But I had some suggestion 

- Right now, uv pip install still runs even on cache hits, which partially defeats the purpose of caching the .venv, since we lose most of the install-time savings.

- The use of a fully resolved dependency key is more deterministic, but it also adds noticeable overhead (~30–60s on cache miss) and may reintroduce strict resolution issues similar to what we saw earlier.

- Also noticed that source .venv/bin/activate is still present — this might not be necessary if we rely on $GITHUB_PATH for environment setup.

@Sharkyii

.github/workflows/install-and-test.yml
```diff
@@ -79,7 +103,12 @@ jobs:
...
         run: |
-          uv venv --no-project
+          if [ "${{ steps.cache-venv.outputs.cache-hit }}" != "true" ] && [ -d .venv ]; then
+            rm -rf .venv
+          fi
+          if [ ! -d .venv ]; then
+            uv venv --no-project
+          fi
           source .venv/bin/activate
           uv pip install torch --index-url ${{ matrix.torch_index }}
```
kshirajahere: Since .venv is restored from cache, these install steps still run on cache hits and can remove most of the speedup. Can we guard both uv install commands so they run only when the venv cache was not hit?

CHANGELOG.md
```diff
@@ -51,6 +51,10 @@ This release introduces new features including GIF animation support, wandb run
 
 - Initialize `da_forcing_mean` and `da_forcing_std` to `None` when forcing data is absent, fixing `AttributeError` in `WeatherDataset` with `standardize=True` [\#369](https://github.com/mllam/neural-lam/issues/369) @Sir-Sloth-The-Lazy
 
+### Maintenance
+
+- Add two-layer caching to install-and-test CI workflow (pip download cache + uv venv cache) to reduce build times by 80-95% [\#605](https://github.com/mllam/neural-lam/pull/3605) @Sharkyii
```
kshirajahere:
```suggestion
- Add two-layer caching to install-and-test CI workflow (pip download cache + uv venv cache) to reduce build times by 80-95% [\#606](https://github.com/mllam/neural-lam/pull/606) @Sharkyii
```

should be 606 for PR right and not 605 or 3605(which doesn't exist )

kshirajahere (commented):
Nice direction overall, especially adding two-layer caching. Few minor changes and update in /CHANGELOG.md

CHANGELOG.md
Sharkyii: My bad, didn't saw it..

.github/workflows/install-and-test.yml
Sharkyii: Updated it to guard on cache-hit.

Sharkyii:
**I have now guarded installs with cache hit**

> * The use of a fully resolved dependency key is more deterministic, but it also adds noticeable overhead (~30–60s on cache miss) and may reintroduce strict resolution issues similar to what we saw earlier.
> * Also noticed that source .venv/bin/activate is still present — this might not be necessary if we rely on $GITHUB_PATH for environment setup.


The hash is important because, without it, the cache can keep stale packages even after dependencies change. That would mean CI runs against incorrect versions, which defeats the whole point of using a lockfile. The extra 30–60 seconds when dependencies update is a fair trade-off for reliable, deterministic testing.
Also, the source .venv/bin/activate step can’t be removed. The uv pip install commands that follow run in the same shell step and depend on the virtual environment being active. $GITHUB_PATH only impacts later steps in the workflow, not the current one.

Thanks @archit7-beep @kshirajahere

Sharkyii:
@sadamov review :)

.github/workflows/install-and-test.yml
```diff
@@ -44,13 +44,37 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Python 3.13
+        id: setup-python
         uses: actions/setup-python@v6
         with:
           python-version: 3.13
+          cache: 'pip'
```
sadamov: `cache: 'pip'` also runs for uv matrix jobs where it caches a directory uv never reads.

```suggestion
          cache: ${{ matrix.package_manager == 'pip' && 'pip' || '' }}
```

sadamov (commented):
Looks good! And the speedup is impressive @Sharkyii also thanks to @archit7-beep for their review!

I added some inline suggestions below. Noting the seperate PR for uv lockfile here:

**Ephemeral `uv.lock` as cache key:** `uv lock` is generated fresh on every run (no committed lock file, by design). If any upstream dep releases between two CI runs, the hash changes and the venv cache is invalidated. The stated 80-95% speedup may be optimistic during active periods. This will be resolved once #604 lands - it introduces `uv sync` with CPU/GPU extras and `[tool.uv.sources]`, which enables a committed `uv.lock` and makes the cache key fully stable.

.github/workflows/install-and-test.yml
```diff
@@ -74,15 +98,23 @@ jobs:
       - name: Install package (including dev dependencies)
         if : matrix.package_manager != 'uv'
         run: |
-          python -m pip install --group dev .
+          python -m pip install -e .[dev]
```
sadamov: `.[dev]` targets optional extras; `dev` is a `[dependency-groups]` entry. pip silently ignores the unknown extra and skips `pytest`, `pre-commit`, `pooch`. CI still passes because the pip path only runs a smoke test, not pytest.

```suggestion
          python -m pip install --group dev .
```

.github/workflows/install-and-test.yml
```diff
@@ -74,15 +98,23 @@ jobs:
...
+            uv venv --no-project
+          fi
           source .venv/bin/activate
-          uv pip install torch --index-url ${{ matrix.torch_index }}
-          uv pip install --group dev -e .
+          if [ "${{ steps.cache-venv.outputs.cache-hit }}" != "true" ]; then
+            uv pip install torch --index-url ${{ matrix.torch_index }}
+            uv pip install --group dev .
+          fi
+          uv pip install -e . --no-deps
```
sadamov: `source .venv/bin/activate` is redundant (uv auto-detects `.venv`; `$GITHUB_PATH` handles subsequent steps). On cache miss, the non-editable install at line 115 is immediately overridden by the unconditional editable install at line 117. On cache hit, line 117 runs an unnecessary install - the cached editable install already points to the source dir and works as-is.

```suggestion
          if [ "${{ steps.cache-venv.outputs.cache-hit }}" != "true" ]; then
            uv pip install torch --index-url ${{ matrix.torch_index }}
            uv pip install --group dev -e .
          fi
```

.github/workflows/install-and-test.yml
```diff
@@ -44,13 +44,37 @@ jobs:
...
+
+      - name: Cache virtual environment
+        if: matrix.package_manager == 'uv'
+        id: cache-venv
+        uses: actions/cache@v4
+        with:
+          path: .venv
+          key: uv-venv-${{ runner.os }}-${{ matrix.device }}-py${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('uv.lock') }}
+          restore-keys: |
+            uv-venv-${{ runner.os }}-${{ matrix.device }}-py${{ steps.setup-python.outputs.python-version }}-
```
sadamov: When the exact key misses, the `restore-keys` prefix match restores an old `.venv` from cache — which is then immediately deleted because `cache-hit != "true"`. The restore wastes download time with no benefit.

```suggestion
          key: uv-venv-${{ runner.os }}-${{ matrix.device }}-py${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('uv.lock') }}
```

.github/workflows/install-and-test.yml
```diff
@@ -79,10 +103,16 @@ jobs:
...
+          if [ "${{ steps.cache-venv.outputs.cache-hit }}" != "true" ] && [ -d .venv ]; then
+            rm -rf .venv
+          fi
+          if [ ! -d .venv ]; then
+            uv venv --no-project
+          fi
+          if [ "${{ steps.cache-venv.outputs.cache-hit }}" != "true" ]; then
+            uv pip install torch --index-url ${{ matrix.torch_index }}
+            uv pip install --group dev -e .
+          fi
```
sadamov: The three guards on `cache-hit` and `[ -d .venv ]` are redundant. The middle condition (`[ ! -d .venv ]`) is only reachable when `cache-hit != "true"` since the first block already deleted `.venv`. Collapse into one:

```suggestion
          if [ "${{ steps.cache-venv.outputs.cache-hit }}" != "true" ]; then
            rm -rf .venv
            uv venv --no-project
            uv pip install torch --index-url ${{ matrix.torch_index }}
            uv pip install --group dev -e .
          fi
```

sadamov (commented):
The two-layer approach (uv download cache via `enable-cache: true` + explicit `.venv` cache) is sound. Two issues below.

Sharkyii:
@sadamov updated the code based on the recommended changes..

sadamov (approved):
marking this one as ready. I only rebased on main because of many recent changes there. thanks @Sharkyii! especially with #604 this will be sweet

sadamov:
going through the changes here for tomorrows dev-meeting. Is the changelog slightly outdated? it still refers to caching the pip download, but I think we completely moved to UV in #604, no?

Sharkyii:
yeah, that line was stale from before #604, just fixed the wording, pushed.

sadamov:
We are a bit slow with our version releases. So my initial assigment to v0.8.0 seems a bit defensive now. But I guess its not urgent so we can wait a bit :)

sadamov:
just a note from the dev-meeting: this PR here caches the deps and venv but not the downloaded datasets (pooch)(danra,meps,cosmo,era,...). a separate isse will handle he caching of pooch datasets.

PR #616, opened 2026-04-21
tests: add hypothesis property-based tests for time slicing invariants

### LINKED ISSUE #602: Add property-based testing with hypothesis
Most tests currently use handpicked input values and assert specific outputs. Hypothesis generates inputs automatically and finds edge cases you wouldn't think to write. E.g. https://github.com/ecmwf/anemoi already uses it for testing attention and MLP layers.

### Candidates

- **Standardization**: output should always have mean ~0 and std ~1 regardless of input shape or value range
- **Clamping**: model output should always stay within declared bounds for any input tensor. `test_clamping.py` already tests specific cases, hypothesis would test the invariant more thoroughly
- **Dataset slicing**: `test_time_slicing.py` hardcodes specific input/output pairs - these could be expressed as properties that hold for any valid combination of `ar_steps`, `num_past_forcing_steps`, `num_future_forcing_steps`
- **Datastore interface contracts**: `get_dataarray` output shape should always match `num_grid_points` for any valid split and category

### Setup

`hypothesis` would be a `dev` dependency only and integrates directly with pytest via `@given` decorators, so no CI changes are needed.

PARTHSH1:
Closes #602

Adds `hypothesis` as a dev dependency and introduces property-based tests
for the time slicing invariants in `WeatherDataset`.

The existing `test_time_slicing.py` hardcodes 4 specific combinations of
`ar_steps / num_past_forcing_steps / num_future_forcing_steps`, all with
`ar_steps=3` and `num_future_forcing_steps=0`. These new tests express the
same logic as invariants that hold for any valid parameter combination,
including `num_future_forcing_steps > 0` which was never covered before.

Tests added:
- `test_time_slicing_output_shapes_invariant` — shape of init_states,
  target_states, forcing, target_times always matches the parameters
- `test_time_slicing_sample_index_invariant` — invariants hold for any
  valid sample index, not just index 0
- `test_step_length_invariant` — any positive step_length produces a
  usable dataset

Reuses `SinglePointDummyDatastore` from `test_time_slicing.py` with no
new test infrastructure. `hypothesis` is a dev-only dependency and
integrates with pytest via `@given` decorators — no CI changes needed.

### DIFF
--- a/tests/test_property_based.py
+++ b/tests/test_property_based.py
@@ -1,0 +1,146 @@
+# Standard library
+from datetime import timedelta
+from pathlib import Path
+
+# Third-party
+import numpy as np
+import pytest
+from hypothesis import assume, given, settings
+from hypothesis import strategies as st
+
+# First-party
+from neural_lam.weather_dataset import WeatherDataset
+from tests.test_time_slicing import SinglePointDummyDatastore
+
+ANALYSIS_STATE_VALUES = list(range(30))  # bigger pool than original
+FORCING_VALUES = list(range(100, 130))
+
+
+def make_datastore(num_steps=30):
+    time_values = np.datetime64("2020-01-01") + np.arange(num_steps)
+    return SinglePointDummyDatastore(
+        state_data=ANALYSIS_STATE_VALUES[:num_steps],
+        forcing_data=FORCING_VALUES[:num_steps],
+        time_values=time_values,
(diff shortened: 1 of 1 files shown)


sadamov:
I have to close this one again, as the issue was recently assigned to https://github.com/snakefood3232

PR #629, opened 2026-04-29
Fix/create dataarray category feature

Pran-jal25:
WeatherDataset.create_dataarray_from_tensor had 2 bugs:

Wrong coordinate access — the feature coordinate was hardcoded as .state_feature regardless of the category argument. Calling the method with category="forcing" or category="static" raised AttributeError: 'DataArray' object has no attribute 'state_feature'. Fixed by using da_datastore_state[f"{category}_feature"] so the correct coordinate (forcing_feature, static_feature, etc.) is accessed dynamically.

Missing f-string prefix — the error message for the 2D tensor validation case was a plain string "dimensions (grid_index, {category}_feature), but got ", so {category} appeared literally in the message instead of being interpolated. Added the f prefix.

No behaviour change for the existing category="state" path.


sadamov:
closing this because the PR template is missing and no issues linked. Also, I am pretty sure these two bugs are already handled in separate PRs.

Issue #138, opened 2025-04-02
Enabling training on interior and boundary datastore


leifdenby:
@sadamov and @joeloskarsson have been doing great work on https://github.com/mllam/neural-lam/pull/93 and https://github.com/mllam/neural-lam/pull/90 to enable training and inference where the domain interior and boundary come from different datastores.

I am making this issue to start trawling through changes on https://github.com/joeloskarsson/neural-lam-dev/tree/research that relate to this but are not currently included in the branches that the above PRs currently contain. The aim here is to have a complete list of changes and built a plan to get #93 and #90 merged.

- bugfix for boundary forcing time sampling: a2bcf239713d9db3bd4a809e4251f11169d6b3b3
- feat: dynamic time deltas for boundary forcing: e58e334c64b6a75fd9da83791666827d7490a917
- feat: allow for using forecast data on boundary: fb820e2a2acede39b8d9ea8a454444faccf1cd10
- feat: writing eval predictions to zarr: 39eb6e2aa5ce20ff1ce311c92988caee176bea43 (appears to use implementation discussed on https://github.com/mllam/neural-lam/pull/104)
- bugfix restoration of trainer state -  e7d11c9a9b10d0165314d00f230243addd79d9fb
- bugfix: ensure that eval actually uses defined eval_split - 819b47e64aecf4f21ebd5dce6c3a3091b98c0b5c
- bugfix for clamping of predictions during finetuning - 7985f2cb7b03936af50bdbc0a2fa069ab48b4f06

sadamov:
@joeloskarsson can you add your items here please

joeloskarsson:
# Crucial things to merge
These are the things that are crucial to merge to have a "realistic boundary setup", as in our paper. In some categories:

## Graph

### In WMG

- [ ] Separate g2m connectivity arguments for boundary and interior grid nodes (https://github.com/joeloskarsson/weather-model-graphs/commit/974ad07e0afe6aa84312acafe61844041f07c061)
  - This is absolutely necessary to be able to use wmg to create graphs without having disconnected nodes.
- [x] Decoding mask, to only include a subset of the grid nodes (the interior grid nodes) in m2g (https://github.com/mllam/weather-model-graphs/pull/34)
- [ ] Allow for filtering created mesh nodes to convex hull of all grid nodes. 
  - Crucial, as without this mesh nodes in g2m will easily end up with degree 0, creating all kinds of problems.
[...]

Issue #108, opened 2025-01-28
Plotting Boundaries in `train_model.py --eval` not just interior


sadamov:
With milestone v0.4.0 we are releasing realistic boundaries. For that purpose the plotting functionalities need to be updated. Currently they only plot the interior domain, with a semi-transparent boundary around it. In the new approach the boundary and interior data comes from two separate datastores. 

Some considerations:

- Using domain_cropping in the config will keep only the points at the given distance from the convex hull of the referenced interior data. This set of points will in general not lay on a regular grid.
- There is no boundary mask in neural-lam right now. All points in the boundary datastore are considered boundary input points.
- As plots anyhow are in the crs of the interior, boundary points would still not lay on a regular grid (if they were originally in e.g. a regular lat-lon grid).

I started working on this but got lost in the x,y,lat,lon,projections,coordinates. When are they 1D, when are they 2D, stacked unstacked. Anyways :D here is my working branch: https://github.com/sadamov/neural-lam/tree/plot_boundaries But unfortunately I don't have the bandwidth right now to continue working on it. A typical example plot looks like this:

![Image](https://github.com/user-attachments/assets/a1ea40ae-2229-4904-8d68-95cf8743cea8)

sadamov:
The plotting based on datastores is pretty much solved in https://github.com/joeloskarsson/neural-lam-dev/blob/research/verification/verification_gridded_cosmo.ipynb. 
Because of large size I had to remove the maps from the notebook, but the code is there and the figures are in the paper.

Jayant-kernel:
Thanks for the context here — I read through the issue and the follow-up note about the plotting code being mostly solved in the `verification_gridded_cosmo.ipynb` notebook.

I’m interested in helping with this, but I want to align with the current direction before starting:

* Would you prefer a PR that **ports/cleans up the plotting logic from the research notebook into the main codebase** (for `train_model.py --eval`)?
* Or is this better postponed until more of the interior/boundary datastore work from #138 is merged?

I’m asking because this looks less like a small plotting tweak and more like an integration task across datastore/coords/projection handling, so I want to avoid doing work that conflicts with the current plan.

If helpful, I can first make a small draft PR that only focuses on:

1. plotting boundary + interior from separate datastores, and
2. keeping the existing eval plotting API unchanged.

sadamov:
@Jayant-kernel I'm glad that you asked:
[...]

PR #635, opened 2026-05-11
feat: add optional boundary datastore support

### LINKED ISSUE #652: [RFC/Design]: Support for multiple datastores
I have been thinking about how to generalise #635 to allow for using sources of data that would describe say multiple point-wise observations and/or regular gridded data. Although I am not suggesting that we alter #635 as-is, with which the will be support for a `datastore` (which used both for input and output of the model) and `datastore_boundary` (only used for input), I thought it might be useful to think about how that work could be evolved on down the line. With #625 Neural-LAM configuration will assume a fixed set of datastore roles, such as a main datastore and optionally a boundary datastore. This does not scale well to configurations that combine multiple data sources, for example regional reanalysis, global boundary data, synoptic observations, or satellite-derived inputs.

The aim is to replace role-specific datastore config keys with a single `datastores` mapping. Each datastore should be named by the user and explicitly declare how it is used by the model.

## Proposal

Introduce a `datastores` dictionary in `NeuralLAMConfig`, keyed by datastore name. Each entry should contain the datastore selection fields plus a new `use` section defining which data categories are used as model inputs and outputs.

Suggested config shape:

```yaml
datastores:
  danra:
    use:
      input: [state, static, forcing]
      output: [state]
    kind: mdp
    config_path: danra.datastore.yaml

  era5:
    use:
      input: [state, forcing]
      output: []
    kind: mdp
    config_path: era5.datastore.yaml
```

Suggested dataclasses:

```python
@dataclasses.dataclass
class DatastoreUse:
    input: list[str]
    output: list[str]
```

```python
@dataclasses.dataclass
class DatastoreSelection:
    use: DatastoreUse
    kind: str
    config_path: str
```

```python
@dataclasses.dataclass
class NeuralLAMConfig(dataclass_wizard.JSONWizard, dataclass_wizard.YAMLWizard):
    datastores: Dict[str, DatastoreSelection]
    training: TrainingConfig = dataclasses.field(default_factory=TrainingConfig)
```

## Motivation

This makes datastore roles explicit and extensible. The datastore name describes the data source, while `use` describes how the data source contributes to the model, and the graph connectivity that will be required to be made.

This supports configurations such as:

```yaml
# LAM trained on regional reanalysis with global boundary data
datastores:
  danra:
    use:
      input: [state, static, forcing]
      output: [state]
    kind: mdp
    config_path: danra.datastore.yaml

  era5:
    use:
      input: [state, forcing]
      output: []
    kind: mdp
    config_path: era5.datastore.yaml
```

```yaml
# LAM trained with regional reanalysis, global reanalysis, and synoptic observations
datastores:
  danra:
    use:
      input: [state, static, forcing]
      output: [state]
    kind: mdp
    config_path: danra.datastore.yaml

  synop:
    use:
      input: [forcing]
      output: []
    kind: mdp
    config_path: synop.datastore.yaml

  era5:
    use:
      input: [state, forcing]
      output: []
    kind: mdp
    config_path: era5.datastore.yaml
```

```yaml
# Global reanalysis with satellite-derived inputs
datastores:
  era5:
    use:
      input: [state, static, forcing]
      output: [state]
    kind: mdp
    config_path: era5.datastore.yaml

  amsua:
    use:
      input: [forcing]
      output: []
    kind: mdp
    config_path: amsua.datastore.yaml
```

sadamov:
## Describe your changes

Add optional boundary datastore support to `WeatherDataset` and `WeatherDataModule`, enabling LAM models to ingest boundary forcing from a separate domain (e.g. ERA5 for a COSMO/DANRA interior).

- `NeuralLAMConfig` takes a named `datastores` mapping; a datastore's role follows from the categories it provides, so the one without `state` data is the boundary. The old top-level `datastore:` key raises `InvalidConfigError` pointing at the new format
- `load_config_and_datastore` returns a 3-tuple `(config, datastore, datastore_boundary)`
- `WeatherDataset.__getitem__` returns a 5-tuple `(init_states, target_states, forcing, boundary, target_times)` where the boundary tensor is empty (last dim 0) when no boundary datastore is configured
- New CLI args `--num_past_boundary_steps` / `--num_future_boundary_steps` control the boundary forcing window size
- `ForecasterModule.common_step` unpacks the boundary tensor but does not yet wire it into the forward pass (model-side integration is planned as a separate PR)
- Boundary datastore can be any registered datastore type (mdp, npyfilesmeps, etc.)
- `MDPDatastore` and `NpyFilesDatastoreMEPS` both handle boundary-only configs (forcing + static, no state variables) gracefully
- New `is_on_regular_spatial_grid` property on `BaseDatastore`: a domain-cropped `MDPDatastore` reports `False`, and graph creation and `datastore.plot_example` refuse it rather than silently padding the missing cells
- ERA5 boundary test configs added at `tests/datastore_examples/mdp/era5_1000hPa_danra_100m_winds/` (WeatherBench2 64x32 equiangular grid as boundary for DANRA interior), with `init_datastore_boundary_example()` fixture in `conftest.py`

This is PR A in the boundary datastore plan outlined in #138. PR B (model-side boundary handling) and PR C (#636, boundary plotting) will follow.

## Issue Link

refs #108

## Type of change

Note on breaking change: every existing config must move its datastore under a named entry in `datastores:`. `WeatherDataset.__getitem__` now returns a 5-tuple instead of 4-tuple, and `load_config_and_datastore` returns a 3-tuple instead of 2-tuple. All callers in the repo have been updated.

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

refs #138


neural_lam/datastore/mdp.py
```diff
@@ -453,9 +472,16 @@ def grid_shape_state(self):
...
         """
-        ds_state = self.unstack_grid_coords(self._ds["state"])
+        # Use state if available, otherwise fall back to forcing (e.g.
+        # boundary-only datastores that have no state variables).
+        for category in ("state", "forcing", "static"):
+            if category in self._ds:
+                ds_cat = self.unstack_grid_coords(self._ds[category])
+                break
+        else:
+            raise ValueError("Dataset has no state, forcing, or static data")
```
joeloskarsson: Is this really possible now, then we have a fully empty datastore? I feel like we should raise an error way before this then (in the constructor).

neural_lam/models/module.py
```diff
@@ -243,9 +243,10 @@ def on_after_batch_transfer(self, batch, dataloader_idx):
 
         Lightning calls this for every train/val/test/predict batch.
         WeatherDataset returns unstandardized state and forcing; both are
-        normalized here so the work runs on the accelerator.
+        normalized here so the work runs on the accelerator. The boundary
+        forcing is passed through unchanged.
```
joeloskarsson: Is there a good reason to not do the standardization of boundary data here? This feels confusing and inconsistent to me. Best to do all standardization for the batch in the same place.

neural_lam/models/module.py
```diff
@@ -268,10 +269,16 @@ def on_after_batch_transfer(self, batch, dataloader_idx):
...
 
     def common_step(self, batch):
-        init_states, target_states, forcing_features, batch_times = batch
+        (
+            init_states,
+            target_states,
+            forcing_features,
+            boundary_features,
+            batch_times,
+        ) = batch
```
joeloskarsson, line 413:
```suggestion
        (
            init_states,
            target_states,
            forcing_features,
            boundary_features,
            batch_times,
        ) = batch
        # NOTE: For now we do not use the boundary features from here. This is yet 
        # to be implemented on the model side.
```

neural_lam/weather_dataset.py
```diff
@@ -457,13 +545,15 @@ def __getitem__(self, idx):
         )
 
         forcing = torch.tensor(da_forcing_windowed.values, dtype=tensor_dtype)
+        boundary = torch.tensor(da_boundary_windowed.values, dtype=tensor_dtype)
 
         # init_states: (2, N_grid, d_features)
         # target_states: (ar_steps, N_grid, d_features)
         # forcing: (ar_steps, N_grid, d_windowed_forcing)
+        # boundary: (ar_steps, N_boundary_grid, d_windowed_boundary)
```
joeloskarsson: I am wondering if it is wise to return an empty tensor here, or if it would just be better for this to be None? I suppose that with None we would need a custom (but simple) collate function for the batching. 
My reasoning is that if you for example forget to specify a boundary datastore then many things will still work and this will be a silent bug, potentially even the forward pass would work (in some scenario)? I can not see a case where we would not want to explicitly do things differently in a model depending on if the boundary forcing is present or not, and it being None would be a good signifier of this. Otherwise everything processing boundary would need to have the boundary datastore, to check if that is None.
I am not sure about this, and could probably be convinced either way. But would be happy to hear your thoughts.

neural_lam/weather_dataset.py
```diff
@@ -375,10 +415,24 @@ def _build_item_dataarrays(self, idx):
                 da_forcing=da_forcing, idx=sample_idx, n_steps=self.ar_steps
             )
 
+        # Slice boundary forcing if available
+        if self.da_boundary_forcing is not None:
+            da_boundary_windowed = self._slice_forcing_time(
+                da_forcing=self.da_boundary_forcing,
+                idx=sample_idx,
```
joeloskarsson: I don't think this is the correct idx for the boundary. What about when the boundary has a different time step? What about when the boundary is a forecast? I feel like we used to have a lot of logic to find this alignment, that I can't see now, which makes me fear that we dropped something important. 

But this might also be a matter of the scope of this PR, if you were intending this to restrict to reanalysis boundary with the same timestep? I would though prefer to make what we merge in as similar as possible to our contribution in the paper :)

tests/test_datasets.py
joeloskarsson, line 1: Could we also port over a version of https://github.com/joeloskarsson/neural-lam-dev/blob/research/tests/test_time_slicing.py? I remember this being very useful to figure out all the alignment between interior and boundary data.

joeloskarsson (changes requested):
It's so great that you started the work on this 😄 I had a look through everything except the tests now. Had one major comment about the interior-boundary data alignment in the dataset, otherwise mostly small things.

neural_lam/datastore/mdp.py
sadamov: Fair point, fully empty was never meant to be valid. Added a check in `__init__` that requires `state` or `forcing`, and simplified `grid_shape_state` to just pick between those two (dropped the `static` fallback and the defensive `else`). fixed in cf323a8.

neural_lam/models/module.py
sadamov: As we discussed over lunch, I would argue that this is now part of the model side PR B, since we are using on_batch_transfer_end methods now and relieved the WeatherDataset of its standardization duties. But, we said to implement it here nonetheless so: `ForecasterModule.__init__` now takes an optional `datastore_boundary` and registers `boundary_mean`/`boundary_std` buffers, and `on_after_batch_transfer` standardizes the boundary exactly like the interior forcing. With no boundary datastore the buffers stay `None` and the tensor passes through unchanged. Covered by the two new tests in `tests/test_gpu_normalization.py`. done in 9a9af31.

neural_lam/models/module.py
sadamov: Applied with a small reword (0ec56f5). Boundary is now standardized in `on_after_batch_transfer` (comment #2) but still not consumed by the forward pass, so in 9a9af31 I extended the NOTE to point at the standardization and reference #108 for the model-side wiring.

neural_lam/weather_dataset.py
sadamov: I went with an empty tensor to avoid a custom collate and keep `__getitem__` shape-stable for code that just unpacks the 5-tuple. The silent-bug risk is real though, and I think the model wiring (PR B) is the right place to catch it: the model knows whether it expects boundary, so we can assert there that `boundary.shape[-1] > 0` matches `datastore_boundary is not None`. Keeps the loader simple but still fails loudly when they get out of sync.
what do you think?

neural_lam/weather_dataset.py
sadamov: You were right, I messed up the partial port from the research branch, trying to reduce the scope of the PR.
Ported the time-based alignment from the research branch, so all four combinations work now: analysis/forecast interior crossed with analysis/forecast boundary. See 140caf5.

A few intentional deviations from the research branch:

- No `interior_subsample_step` / `boundary_subsample_step` (buggy, orthogonal)
- No `window_time_deltas`, `dynamic_time_deltas`, `time_slice` concat in the boundary tensor (didn't help, can be added later)
- Per-step the window still centers on the target time (the boundary condition for the interior at the predicted time), but only after the launch is fixed correctly on init. see a543ce9

tests/test_datasets.py
sadamov: Good idea, that file is really useful. Extended the existing `tests/test_time_slicing.py` rather than adding a new one. `SinglePointDummyDatastore` now supports forecast mode, plus a new `BoundaryOnlyDummyDatastore` that mirrors a real boundary store (forcing-only, state access raises `KeyError`).
Added in cbc1c14.

observingClouds:
Hi @sadamov,
Thanks for starting and advancing this feature integration. As this is something that we at DMI need to make our operational AI model, let me and @SimonKamuk know if we can help to bring this over the finish line. We might have some comments too 😛

sadamov:
@observingClouds yes please! Feel free to leave your reviews. If there are major pieces missing, you should have push access to my PR-branch here and after a quick ping you can also push directly.

leifdenby:
@sadamov I added https://github.com/mllam/neural-lam/issues/652 to say how I would generalise the config. I am not saying that we need to use that config layout with this PR, but you could maybe have a look at that PR and see what you think? If you do like to use it then we could maybe consider using a structure in this PR that could remain relatively unchanged down the line.

sadamov:
> @sadamov I added https://github.com/mllam/neural-lam/issues/652 to say how I would generalise the config. I am not saying that we need to use that config layout with this PR, but you could maybe have a look at that PR and see what you think? If you do like to use it then we could maybe consider using a structure in this PR that could remain relatively unchanged down the line.

https://github.com/mllam/neural-lam/issues/652#issuecomment-4657364109

observingClouds (commented):
LGTM. I just wonder if we need to raise a deprecation notice for the new config format, so that users now how to change it.

observingClouds (approved):

sadamov:
@joeloskarsson this is ready for another look when you have time. Summary of what changed since your review:

- reworked to the named `datastores:` mapping we landed on in #652 (state/forcing meaning kept, just named), and added `plotting.boundary_datastore` to name the boundary overlay source - that's your option 3 from #652
- migration guard in `load_config_and_datastore` raising `InvalidConfigError` on the legacy top-level `datastore:` key, so users get pointed at the new format (@observingClouds )
- the interior-boundary time alignment is ported from the research branch (analysis/forecast interior crossed with analysis/forecast boundary, all four combinations), with coverage in `test_time_slicing.py` among other new tests
- @leifdenby documented the state/forcing/static input/output meaning in the README Data section as you asked in #652 - have a look.

README.md
```diff
@@ -584,13 +611,18 @@ Canonical dimension names used in tensor shape annotations throughout the codeba
...
+- `num_boundary_grid_nodes` - number of nodes in the flattened boundary spatial grid
 - `num_mesh_nodes` - number of mesh nodes; indexed as `num_mesh_nodes[l]` for hierarchical level `l`
 - `num_state_vars` - number of atmospheric state variables
 - `num_forcing_vars` - number of forcing input variables
+- `num_windowed_forcing_vars` - forcing variables stacked over the past/future forcing window
+- `num_windowed_boundary_vars` - boundary forcing variables stacked over the past/future boundary window
 - `num_variables` - generic variable dimension used in metric functions
 - `hidden_dim` - internal hidden representation size in GNN layers and MLPs
 - `input_dim` - input feature dimensionality to a layer before transformation
+- `d_mesh_static` - number of static features per mesh node
```
joeloskarsson, line 626: This feels like a different pattern, and would be better to name similar to others. Maybe either `num_mesh_vars` or `mesh_static_dim`.

README.md
```diff
@@ -215,6 +226,21 @@ the input-data representation is split into two parts:
...
+| `state`   | ✓           | ✓            | Prognostic variables the model both reads and predicts (autoregressed forward in time). |
+| `forcing` | ✓           |              | Time-varying inputs known in advance (e.g. solar radiation, boundary forcing). |
+| `static`  | ✓           |              | Time-invariant inputs (e.g. orography, land-sea mask). |
+
+(A `diagnostic` category, predicted but not fed back as input, is not used at
+present but fits naturally as an output-only category.) These categories are
+also what determine a datastore's role: a datastore that provides `state` data
+is the interior domain (model input and output), while one without `state` data
+is used for input only, e.g. boundary forcing from a separate domain.
+
```
joeloskarsson: I don't think we should mention the potential diagnostic category here unless we add it. If we want this documented somewhere as a potential future thing, it would be better to just add a github issue about it.

README.md
```diff
@@ -175,7 +179,14 @@ training:
...
 
-1. The kind of datastore and the path to its config
+1. A named mapping of datastores (`datastores`), each giving the kind of
+   datastore and the path to its config. The role of each datastore is implied
+   by the categories of data it provides: a datastore that contains `state`
+   data is used for both model input and output (the interior domain), while a
+   datastore without `state` data is used for input only (e.g. boundary forcing
+   from a separate domain such as ERA5 for a LAM domain). Exactly one datastore
+   must provide `state` data. When a boundary datastore is present its forcing
+   is windowed and included as an additional tensor in each training sample.
```
joeloskarsson: Can there be multiple boundary datastores, providing only forcing data? Would be good to mention explicitly, since we say that there can only be one interior, but never discuss this for boundary.

neural_lam/models/module.py
joeloskarsson: Thanks! Yes this is really on the model side (when I look at it now, I see that this is even more so the case than I originally though 😅). Still, I think it is very good that we add this here, as it is clearer to see how it ties in with the rest of the data loading in this PR. And if needed I can just modify this in the model-side follow up PR. But the standardization looks all correct and good now.

neural_lam/weather_dataset.py
joeloskarsson: That sounds like a very good solution, yes. Then the `ForecasterModule` can do this check (it will need to know about the datastore_boundary anyhow), and make sure that everything is consistent before passing anything on to a `Forecaster`.
Just need to remember to add this on the model side. I agree that this belongs in the model-side PR. Keep me to this ;)

joeloskarsson: I'll resolve this now, as nothing more should be changed here.

neural_lam/weather_dataset.py
joeloskarsson: Will go over the additions in my current review, and resolve this comment.

neural_lam/train_model.py
```diff
@@ -346,6 +346,19 @@ def main(input_args=None):
...
+        default=1,
+        help="Number of past time steps to use as input for boundary forcing",
+    )
+    data_group.add_argument(
+        "--num_future_boundary_steps",
+        type=int,
+        default=1,
+        help="Number of future time steps to use as input for boundary "
+        "forcing",
+    )
```
joeloskarsson, line 362:
```suggestion
    data_group.add_argument(
        "--num_past_boundary_steps",
        type=int,
        default=1,
        help="Number of past time steps to use as input for boundary forcing, "
        "when present",
    )
    data_group.add_argument(
        "--num_future_boundary_steps",
        type=int,
        default=1,
        help="Number of future time steps to use as input for boundary "
        "forcing, when present",
    )
```

neural_lam/utils.py
joeloskarsson, line 1: In #648 we noted that the utils.py file is really growing long, and it is similar here with these additions. The plan is to change this into a separate module, with multiple utils files. @Sir-Sloth-The-Lazy kindly agreed to look into this in a separate PR. Once that is done, the utils additions here are probably suitable in their own file in ther. Not sure which PR will land first, but should be no problem to adapt any one to the other. 

This is just a comment, something that can be good to know. Nothing that needs change in this PR.

neural_lam/weather_dataset.py
```diff
@@ -201,183 +294,215 @@ def __len__(self) -> int:
...
-                da_sliced = da_forcing.isel(
-                    analysis_time=idx,
-                    elapsed_forecast_duration=slice(start_idx, end_idx + 1),
+            # Choose a single analysis_time (launch) for this sample. We
+            # anchor on the model init time (the last input state), not the
+            # first target, so we never select a boundary forecast launched
+            # after init - that forecast would be unavailable operationally.
+            # A launch exactly at init is also rejected (strictly before),
+            # then shifted further back if a larger num_past_steps requires
+            # more lead headroom.
```
joeloskarsson, line 381: Hmm, I think this might require some explicit discussion. Maybe this is good for a first "simple" solution (none of this time step handling is simple really 😅), but there seems to be more refined approaches. 
What I am thinking is that: Could not a forecast launched a exactly init be considered operationally available? If we consider a workflow where 
1. I have available to me (interior analysis, global analysis)
2. I run a global AI model from the global analysis and potentially crop its output forecast to match my boundary region.
3. I now run my LAM model from interior analysis and boundary data from the forecast in point 2.

joeloskarsson, line 381: A separate question here (therefore separate comment 😄) is if we should do this shift back when `num_past_steps` requires more lead room. This is necessary as long as we require all boundary forcing to come from the same forecast, but there exists an alternative where we would use the best possible data before the current init time. What I mean concretely is that we would for all future boundary forcing take states from the forecast started at the last time <= init time. But for the past boundary forcing steps, these could come from earlier analyses or forecasted states from a forecast started earlier. So for init_time 4, boundary forcing forecast data that is only initialized at times 0, 2, 4 , ... and with `num_past_steps = num_future_steps = 2` we could end up with a windowed forcing like 
```
[
  (init_time = 2, fc_time_delta = 0),
  (init_time = 2, fc_time_delta = 1),
  (init_time = 4, fc_time_delta = 0),
  (init_time = 4, fc_time_delta = 1),
  (init_time = 4, fc_time_delta = 2),
]
```
I hope this makes sense, let me know otherwise. The downside of something like this is that the boundary forcing will not be internally consistent throughout the window. The upside is that you always get the latest information. I think we had this exact discussion during the research work, and chose to prioritize consistency. But perhaps we should now make this an explicit option.

I am in general leaning towards just making this choice of prioritizing consistency in this PR, to not overcomplicate things. Making this an option could be a follow up. But would then be nice to document this idea in an issue.

neural_lam/utils.py
```diff
@@ -825,3 +827,185 @@ def get_integer_time(tdelta: datetime.timedelta) -> tuple[int, str]:
...
+            f"Found different time steps: {np.unique(time_diffs)}"
+        )
+    return time_diffs[0]
+
+
+def check_time_overlap(
+    da1: xr.DataArray,
+    da2: xr.DataArray,
+    da1_is_forecast: bool = False,
+    da2_is_forecast: bool = False,
```
joeloskarsson: Since these dataarrays have a different role I think we should name the variables something more descriptive rather than da1 and da2. That should make this code easier to follow. Similar for crop_time_as_needed.

Maybe da1 can be driving_da, or da_to_cover. da2 could be da_to_validate or covering_da. Or if you have some other good idea :)

neural_lam/utils.py
```diff
@@ -825,3 +827,185 @@ def get_integer_time(tdelta: datetime.timedelta) -> tuple[int, str]:
...
+            da2_is_forecast,
+            num_past_steps,
+            num_future_steps,
+        )
+        return da1
+    except ValueError:
+        if da1_is_forecast:
+            # Cropping a forecast da1 by analysis time would change sample
+            # cardinality silently; leave alignment errors to surface later.
+            return da1
```
joeloskarsson: I am a bit confused why we could not crop da1 if it is a forecast da? Maybe there is some good reason for this that I just forgot 😅

It seems to me that all cropping here would change sample cardinality (assuming this means the number of samples in the dataset?). Not sure if it is quietly, as we log any cropping done.

neural_lam/utils.py
```diff
@@ -825,3 +827,185 @@ def get_integer_time(tdelta: datetime.timedelta) -> tuple[int, str]:
...
+            required_max = da2_tvals[-1] + analysis_offset
+        else:
+            required_min = da2_tvals[0] + num_past_steps * da2_dt
+            required_max = da2_tvals[-1] - num_future_steps * da2_dt
+
+        begin_mask = da1_tvals >= required_min
+        if not begin_mask.any():
+            raise ValueError(
+                "Boundary forcing ends before any interior time satisfies "
+                f"required_min={required_min}; cannot align."
```
joeloskarsson: Do we want to describe this as boundary forcing and interior? The rest of the method is written more agnostic to the meaning of the two da:s.

neural_lam/utils.py
```diff
@@ -825,3 +827,185 @@ def get_integer_time(tdelta: datetime.timedelta) -> tuple[int, str]:
...
+            da2_tvals = da2.analysis_time.values
+        else:
+            da2_tvals = da2.time.values
+
+        da2_dt = get_time_step(da2_tvals)
+        if da2_is_forecast:
+            da1_dt = get_time_step(da1_tvals)
+            analysis_offset = max(da1_dt, num_past_steps * da2_dt)
+            required_min = da2_tvals[0] + analysis_offset
+            required_max = da2_tvals[-1] + analysis_offset
```
joeloskarsson: Hmm, I feel like this should depend on the maximum elapsed_forecast_time of forecasts in da2? And probably then also how long rollouts we are making, i.e. how long target timeseries we are trying to get from the dataset. E.g.
* assume analysis offset = 1 day
* the last forecast of da2 is started on day 5, 
* da2 contains forecasts that are 10 days long
* I am trying to roll out 10 days long forecasts with my ML model, so getting 10 day target timeseries from the dataset class

Then my last da1 time could be at day 5 + 10 = 15, rather than (current code) 5 + 1 = 6. (Might be some off by one error in there, just an example to try to get across my idea).

neural_lam/utils.py
```diff
@@ -825,3 +827,185 @@ def get_integer_time(tdelta: datetime.timedelta) -> tuple[int, str]:
...
+        if not begin_mask.any():
+            raise ValueError(
+                "Boundary forcing ends before any interior time satisfies "
+                f"required_min={required_min}; cannot align."
+            )
+        first_valid_idx = int(begin_mask.argmax())
+        n_removed_begin = first_valid_idx
+        if da1_tvals[-1] > required_max:
+            end_mask = da1_tvals > required_max
+            last_valid_idx_plus_one = int(end_mask.argmax())
```
joeloskarsson: This is very fast right? It's a slightly inefficient way to do this, taking O(T). Could make it O(log(T)) with a binary search, or (maybe even better) just start iterating from the beginning/end, since in most cases we will only slice a little. But I don't know if this is even worth optimizing right now.

neural_lam/weather_dataset.py
```diff
@@ -81,6 +113,67 @@ def __init__(
...
+        # Within-sample time step for the state series: this is the step
+        # between consecutive state times that __getitem__ exposes, used
+        # below to advance the forcing/boundary window across AR steps.
+        if self.datastore.is_forecast:
+            self._time_step_state = get_time_step(
+                self.da_state.elapsed_forecast_duration.values
+            )
+        else:
+            self._time_step_state = get_time_step(self.da_state.time.values)
+
```
joeloskarsson: Is this property actually used anywhere? Did not find where when search now.

neural_lam/weather_dataset.py
```diff
@@ -81,6 +113,67 @@ def __init__(
...
+                    num_future_steps=self.num_future_boundary_steps,
+                )
+                check_time_overlap(
+                    self.da_state,
+                    self.da_boundary_forcing,
+                    da1_is_forecast=self.datastore.is_forecast,
+                    da2_is_forecast=datastore_boundary.is_forecast,
+                    num_past_steps=self.num_past_boundary_steps,
+                    num_future_steps=self.num_future_boundary_steps,
+                )
```
joeloskarsson, line 146: I guess this relates to the comment about cropping also a forecast state da, but this feels a bit like a double-check because we don't trust our cropping. It seems to me that once we have cropped the overlap check should always pass (otherwise the cropping function is wrong). And if a valid cropping is not possible it should throw an error internally, not return

neural_lam/weather_dataset.py
```diff
@@ -201,183 +294,215 @@ def __len__(self) -> int:
...
+                    np.floor((target_time - forcing_at.values) / forecast_step)
                 )
-
-                da_sliced = da_sliced.expand_dims(
-                    dim={"time": [current_time.values]}
+                center_time = forcing_at.values + lead * forecast_step
+                assert center_time <= target_time, (
+                    "Boundary forecast valid time runs ahead of the interior "
+                    f"target time ({center_time} > {target_time})."
                 )
```
joeloskarsson: It seems odd to mix in an assert here. In some sense I guess these could all be asserts, as if we do everything correct with the slicing we should never run into any issues here (right? 🤔). But might be good to keep to catch bugs. Just making it consistent between assert and ValueErrors.

neural_lam/weather_dataset.py
```diff
@@ -456,19 +617,54 @@ def _build_item_dataarrays(
...
+                data=np.empty(
+                    (self.ar_steps, boundary_grid_index.size, 0),
+                ),
+                dims=("time", "grid_index", "forcing_feature"),
+                coords={
+                    "time": da_target_times,
+                    "grid_index": boundary_grid_index,
+                    "forcing_feature": [],
+                },
+            )
```
joeloskarsson: Could we add a helper func to make such empty dataarrays, to not have to repeat code for interior and boundary forcing.

joeloskarsson (changes requested):
Great work with all of this! All my previous comments are fixed, but as you know the time slicing changes meant that there was a lot of new code added. So let me now be annoying and come with more suggestions from looking over also these ;) I still think these are mostly small things, but a couple that maybe need some discussions. Partly this is because I can't remember why we did certain decisions earlier, so don't know if those decisions should be carried forward as we now merge this work to main.

README.md
sadamov: `d_mesh_static` is already the name used in `create_graph.py`, `utils.py` and `test_graph_creation.py` - that's why I didn't use the new naming convention here. Better to align `d_mesh_static` in a small follow-up PR (maybe there is other variables too?)

README.md
sadamov: fair point, dropped the "diagnostic" in 059033d and opened #684 to track it as a future output-only category.

README.md
sadamov: The mapping is generic but `load_config_and_datastore` currently enforces exactly one interior datastore and at most one boundary. So it's a single boundary for now - added that explicitly in 8b08abd.

neural_lam/utils.py
sadamov: that is a great idea, I'll review #682 soon and since it's only refactoring we can merge it quick. Leaving this open and fixing once merged.

neural_lam/weather_dataset.py
sadamov: I agree here, we made excplicitely strict for the building lam paper, to be on the safe side. I think `main` should accept forecasts initialised before OR at the same time as  interior
In operations both the interior initial analysis and the boundary forcing need some time (~30mins) to be ready for forecasting. So it is indeed unclear which one is available sooner (at least to me). So same time is fair. I did not change the code yet and leave this open for discussion.

sadamov: I remember this discussion too, and from that and also from operational Met Services perspective the consistency is usually the preferred choice. BUT it is also not a typical NWP problem to have multiple past boundary window steps. So in some sense we are free to choose either. I think having consistent "within-forecast" windowing here is fine. We can create an issue for the "cascading forecast" approach separately. OK? (I have not done anything yet)

neural_lam/utils.py
sadamov: oh yeah, @leifdenby had similar feedback 1-2 years ago. renamed `da1`/`da2` to `da_requested`/`da_available` in both `check_time_overlap` and `crop_time_if_needed` in dc0585d.

neural_lam/utils.py
sadamov: Done in ec71402: the crop now trims a forecast `da_requested` along `analysis_time` (whole launches dropped, logged like the analysis crop) instead of returning it unchanged, plus a `test_forecast_interior_cropped_along_analysis_time`.
I honestly forgot why we had this in research. :shrug:

neural_lam/utils.py
sadamov: true, phrased the error and log messages in terms of `da_requested`/`da_available` to match the agnostic signature (the docstrings keep a "typically interior/boundary" hint). fixed in dccb7a6.

neural_lam/utils.py
sadamov: oh yeah that was pretty dumb. fixed in ddf882c

neural_lam/utils.py
sadamov: yeah this is so fast it doesn't matter. but your suggestion is faster and clearer - so here it is: 441ce61

neural_lam/weather_dataset.py
sadamov: Nope, must be a left-over from the step-subsampling or the time-delta embed; and so is `_forecast_step_forcing` right next to it. Only `_forecast_step_boundary` is used. Removed both in a75a1ad.

neural_lam/weather_dataset.py
sadamov: agreed. Made `crop_time_if_needed` the single source of truth: it now raises whenever the valid window is empty (`first_valid_idx >= last_valid_idx_plus_one`), not just when nothing clears the lower bound. see 8a2dc12

neural_lam/weather_dataset.py
sadamov: true, the lone `assert center_time <= target_time` among the ValueErrors was inconsistent. Converted it to a ValueError so `_window_forcing_in_time` raises uniformly, fixed in 812d25a.

sadamov:
@joeloskarsson thanks for the review, there were definitely some hickups in the code, still. I left a few unresolved where you can give your input. All others I agreed with and they are implemented with individual commits.

neural_lam/weather_dataset.py
sadamov: good idea, factored both into `_empty_windowed_dataarray(grid_index, target_times)`; in c096dc4

neural_lam/datastore/mdp.py
joeloskarsson, line 25: Something struck me when I was thinking about this change: Is it maybe a problem that `MDPDatastore` is a `BaseRegularGridDatastore` now? Since if we want to use this type of datastore for the boundary data, this will generally not be on a regular grid? Seems like something that might cause silence bugs if one calls the Regular grid methods on the boundary datastore.

Not sure how much the MDP datastore really relies on the regular grid construction? This might not be such a tricky thing to change, but need to look deeper into it.

README.md
joeloskarsson: Ah, I see. Yes, then better to stick to it here -> open issue to align it with other naming in a separate PR.

neural_lam/weather_dataset.py
joeloskarsson: Yes, agree! Let's stick to the consistency here and make the cascading setup a follow up. We also don't have any experimental results (from research) about how the cascading setup would work, and how crucial the consistency is for ML LAM models. I am in general cautious to add anything not proven in research here.

README.md
joeloskarsson: Issue: https://github.com/mllam/neural-lam/issues/687

neural_lam/utils.py
joeloskarsson: Can we just swap the time axis between `time` and `analysis_time` here without any other adjustments. I was thinking that `requested_tvals` here might need to not just include the analysis times but also the valid times for the full last forecast (i.e. cover all the way up to the last analysis time + maximum forecast lead time). Let me know if you already thought about this and I am just overcomplicating things.

neural_lam/utils.py
joeloskarsson: Haha, well the code even got easier to read, so then it's worth it ;)

neural_lam/weather_dataset.py
joeloskarsson: I think same time is fair, for the reasons you outline. But open to discussion from my side as well. However I think we might need to explicitly ask if we want input from anyone else here :)

joeloskarsson (changes requested):
I have gone over all changes, and resolved as suitable. The only things I still think need some consideration are the few open comment threads, in particular my latest comment about the MDP datastore always being the regular grid kind.

neural_lam/datastore/mdp.py
leifdenby: yes I agree. I have since designing this changed my mind about using the class inheritance to denote whether the underlying spatial locations of the datastore are in a regular grid layout. I think instead we should go for having a property on the base datastore abstract base class, maybe `is_on_regular_spatial_grid` or something like that, which can then be used downstream (e.g. in plotting)

I haven't got time to work on this right now unfortunately, but I could write up an issue with ideas for how this could be implemented

sadamov: I see, that is indeed an issue. Let's discuss this tomorrow in the dev-meeting briefly!

joeloskarsson: Yes, either making it a property or a Mixin class would probably simplify things. Could we not do that change here as well? Seems related.

sadamov:
@joeloskarsson this is ready for you:

From your review:
- `is_on_regular_spatial_grid` as a property instead of `BaseRegularGridDatastore` inheritance, following @leifdenby's suggestion: 2bc6425, with ca4a7d3, ab5ade5 and e80a9fd as follow-ups. Graph creation and `plot_example` now refuse a domain-cropped boundary datastore instead of silently unstacking onto a padded rectangle.
- a boundary forecast launched exactly at model init is used rather than the previous launch: 15e77e0.
- `requested_tvals` covering the last analysis time plus the max lead: 406a108.

What CLAUDE complained while reviewing:
- 406a108 when the interior is cropped to stay inside the boundary's coverage, the crop was applied to the state but not to the forcing. Shapes still matched and the times still looked right, so nothing failed, the model just trained on mismatched pairs.
- 0b26e6c picking which boundary launch a sample uses was an estimate from step lengths rather than the actual launch times. The launch is computed directly now. f028798 the same check also counted more samples than the dataset serves, so it asked the boundary to reach further than any real sample needs.
- bde65b0 `datastore_boundary` was dropped on checkpoint reload, and boundary data passed through unstandardized when stats were missing.
- clear failures instead of opaque ones: 824b52f (window off its axis), b0725c4 (non-unique `analysis_time`), 2fbc873, 2163f22, f27f0f5, d0d1d54, 2944bb0, edf7239.
- dead code and docs: a385d36, 29b781e, 078ea21, c31e404, 0913735, 8a4fb69.
- tests: 2387d0c, 07580ea, f2f9fc4, 8560a77, 0c793e4, 8582f17, eb8517b, mostly pinning values where the assertions were vacuous.

Some things I realized while reviewing:

- 3b1fdfa is worth a look: the first ERA5 example build took ~10mins because gcsfs probes the GCE metadata server 40 times. with `NO_GCE_CHECK=true`, its much faster. Is this a good solution, I wonder...
- I just realized that in the research branch we removed the legacy behavior of `n_boundary_points` that would mask the loss for the outermost gridcells of the interior. But both for compatibility and because someone might want to discard the outermost interior cells due to artifacts, we should keep this logic. I have introduced the double logic here: https://github.com/mllam/neural-lam/pull/635/commits/b3bd1879cf1ae4801a34218def066d242c78bd5f

neural_lam/datastore/mdp.py
joeloskarsson: Continuing this discussion here: I like the new introduced property, but I think this might only take us half-way, to a state that is actually quite confusing (but I think also easy to fix).

My idea here (and I think @leifdenby's also, let me know otherwise) was that if we introduce this property then we don't actually need the separate `BaseDatastore` and `BaseRegularGridDatastore`, but could just keep the base one with the property. As is right now, the property basically just tells us which class an object inherits from. 

Consider the scenario that we have some method that should work with both regular grid and non-regular-grid datastores. This is where the property should be useful. Such a method would now have to take a `BaseDatastore` (in order to work with both), and then check the `is_on_regular_spatial_grid` property before doing anything requiring a regular grid. But if `is_on_regular_spatial_grid == True`, and we then want to e.g. call `BaseRegularGridDatastore.unstack_grid_coords(...)`, then we are calling `unstack_grid_coords` on a `BaseDatastore`, which does not have this method. This could be worked around by having the method take a `BaseDatastore` or `BaseRegularGridDatastore`, but then the property just signifies the class anyway. 

The same issue arises here with the subclasses. We should note that we have actually not solved the original issue: `MDPDatastore` still inherits from `BaseRegularGridDatastore`, meaning that we should not put irregular gridded boundary data in it. We can not have `MDPDatastore` inherit from different classes depending on the data is is filled with. 

Overall I would vote for just removing the `BaseRegularGridDatastore`, and letting this new property keep track of if the grid is regular or not. That would require collapsing the two base classes, including the stacking/unstacking in the `BaseDatastore`. It would then have to be shielded behind a check for the regular grid property. I think this would solve our problems, and `MDPDatastore` could set `is_on_regular_spatial_grid` based on the data it actually loads (maybe not trivial to figure out, but worst case we could just check the coordinates if they make up a regular grid 😅).

neural_lam/weather_dataset.py
```diff
@@ -199,186 +396,478 @@ def __len__(self) -> int:
...
+        n_samples -= self.num_future_forcing_steps
+        if self.da_forcing is not None:
+            n_samples = min(
+                n_samples,
+                len(self.da_forcing.time)
+                - offset
+                - n_total
+                + 1
+                - self.num_future_forcing_steps,
+            )
```
joeloskarsson: This code looks a bit unstructured and could be more concise. Suggestion below, from just moving things around step-by-step. Please double-check that it is equivalent.
```suggestion
        n_samples_base = len(times)
        n_samples_offset = - offset - n_total + 1 - self.num_future_forcing_steps
        if self.da_forcing is not None:
            n_samples_base = min(n_samples_base, len(self.da_forcing.time))
        n_samples = n_samples_base + n_samples_offset
```

README.md
```diff
@@ -215,6 +226,28 @@ the input-data representation is split into two parts:
...
    returning `torch.Tensor`-objects.
 
+Each interior `MDPDatastore` also has its own boundary mask - a rim of
+`n_boundary_points` grid cells (default 30) excluded from the loss and
+overwritten in the prediction, independent of pairing it with a separate
+boundary datastore (below). Set `datastores.<name>.n_boundary_points` in
+`config.yaml` to narrow it (e.g. to 0) once boundary forcing comes from
+elsewhere, or leave a small margin to avoid boundary artifacts.
+`NpyFilesDatastoreMEPS`'s boundary mask is fixed at data-preparation time and
+does not support this option.
```
joeloskarsson: Yes, it's good to leave the boundary mask (induced by `n_boundary_points`) in place here. That will be my headache to figure out, once we integrate the boundary-input on the model side 😅

Suggesting some tweaks to these comments, to emphasize that with boundary mask there is actually ground-truth inputs for the interior being fed to the model, so leaving a small boundary is a more substantial decision than just "adding a margin to avoid artefacts".

```suggestion
Each interior `MDPDatastore` still has its own boundary mask - a rim of
`n_boundary_points` grid cells (default 30) excluded from the loss and
overwritten with ground truth in the prediction. This is independent of any 
pairing with a separate boundary datastore (below). 
Set `datastores.<name>.n_boundary_points` in `config.yaml` to 0 once boundary 
forcing comes from elsewhere. If there is no separate datastore with boundary 
forcing, `n_boundary_points > 0` can still be used to take boundary input from the 
interior state. Using both `n_boundary_points > 0` and a separate boundary datastore 
corresponds to taking two kinds of boundary inputs (from interior state and boundary 
datastore), and is not recommended unless you have a very specific setup.
`NpyFilesDatastoreMEPS`'s boundary mask is fixed at data-preparation time and
does not support this option.
```

neural_lam/config.py
```diff
@@ -30,22 +31,45 @@ class DatastoreSelection:
...
+    n_boundary_points : int, optional
+        Overrides :class:`MDPDatastore`'s default boundary-mask width (only
+        valid for `kind: mdp`). Left unset, the datastore's own default (30)
+        applies. This is independent of whether this `datastores:` mapping
+        also has a separate boundary datastore entry: pairing one with an
+        interior datastore does not by itself change how much of the
+        interior's own edge is masked out of the loss, so set this
+        explicitly (e.g. to a small value, or 0) if the interior's boundary
+        mask should shrink now that boundary forcing is supplied
+        externally.
```
joeloskarsson:
```suggestion
        Overrides :class:`MDPDatastore`'s default boundary-mask width (only
        valid for `kind: mdp`). Left unset, the datastore's own default (30)
        applies. This is independent of whether this `datastores:` mapping
        also has a separate boundary datastore entry: pairing one with an
        interior datastore does not by itself change how much of the
        interior's own edge is considered boundary forcing, so set this
        explicitly to 0 if using external boundary forcing. 
```

joeloskarsson (changes requested):
All a bunch of good changes! My earlier comments are fixed. 

There is still some discussion needed about the regular grid datastore. I tried to summarize my view, about getting rid of the subclass, in the thread. But happy to hear if you have any other better idea, as long as it fixes the current inconsistencies with MDPDatastore not always being regular/irregular.

The rest are just small polishing comments.

neural_lam/weather_dataset.py
sadamov: this is a result of my brain after calculating offsets for hours straight :D

sadamov:
@joeloskarsson okay you wanted to go deeper: `BaseRegularGridDatastore` is gone in 05303bc. `grid_shape_state`, `stack_grid_coords`, `unstack_grid_coords`, `spatial_coordinates` and `num_grid_points` all sit on `BaseDatastore` now, behind `is_on_regular_spatial_grid`, and they raise `NotImplementedError` instead of quietly padding the result. So anything that has to deal with both just takes a `BaseDatastore` and checks the property first, no more reading it off the class. That also kills the `MDPDatastore` weirdness you flagged: it works this out from the data it actually loaded, counting its grid points against the distinct `x`/`y` values, i.e. the coordinate check you half joked about. A cropped one says `False` and refuses to unstack, the same class uncropped says `True`, both pinned in tests. Your other three are in too: e72397f the `n_samples` restructure, e5bd5e4 the README boundary mask wording, 79abdcc the `n_boundary_points` docstring. Ready for another look.

neural_lam/datastore/base.py
```diff
@@ -365,173 +390,83 @@ def get_lat_lon(self, category: str) -> np.ndarray:
...
+        Datastores that are not on a regular spatial grid must override
+        this.
+
         Returns
         -------
         int
             The number of grid points in the dataset.
 
         """
+        return self.grid_shape_state.x * self.grid_shape_state.y
```
joeloskarsson, line 407: Just for clarity (I understand that `grid_shape_stat` would throw an error if irregular grid), to make this more direct, can we guard this with a regular grid check?

```suggestion
       """
        if self.is_on_regular_spatial_grid:
            return self.grid_shape_state.x * self.grid_shape_state.y
        else:
            raise NotImplementedError(
                "`num_grid_points` is not implemented for irregularly spatial grids in the `BaseDatastore` class."
                "Please override this method in the concrete datastore implementation."
            )
```

neural_lam/datastore/base.py
```diff
@@ -365,173 +390,83 @@ def get_lat_lon(self, category: str) -> np.ndarray:
...
-    unstack the spatial grid coordinates of the data-arrays (called
-    `stack_grid_coords` and `unstack_grid_coords` respectively).
-    """
-
-    spatial_coordinates: tuple[str, str] = ("x", "y")
+        return False
 
     @cached_property
-    @abc.abstractmethod
     def grid_shape_state(self) -> CartesianGridShape:
```
joeloskarsson, line 426: Now that this property exists for both regular and irregular stores, I think we should maybe name it something more specific. Since I would intuitively expect it to return the same as `num_grid_points` for an irregular grid, but that is not the intention.
Maybe we can call it `regular_grid_shape` or `cartesian_grid_shape`. Does it need to include `state` in the name?  Do we allow for state/forcing/static to be on different grids in the same datastore? E.g. `num_grid_nodes` does not have state in the name.

neural_lam/datastore/base.py
```diff
@@ -365,173 +390,83 @@ def get_lat_lon(self, category: str) -> np.ndarray:
...
+            If the datastore's grid points do not form a complete 2D grid.
+        """
+        if not self.is_on_regular_spatial_grid:
+            raise NotImplementedError(
+                f"`{operation}` requires the grid points to form a complete "
+                "2D grid (`is_on_regular_spatial_grid`); "
+                f"{type(self).__name__} reports that they do not, so the "
+                "result would be padded with cells the datastore has no "
+                "data for."
+            )
```
joeloskarsson, line 469: Seems overly specific, we should not assume that check this is only used for reshaping
```suggestion
                f"{type(self).__name__} reports that they do not."
            )
```

neural_lam/datastore/mdp.py
```diff
@@ -474,17 +545,24 @@ def coords_projection(self) -> ccrs.Projection:
...
-        ds_state = self.unstack_grid_coords(self._ds["state"])
+        category = "state" if "state" in self._ds else "forcing"
+        da_cat = self._ds[category]
         xdim, ydim = self.spatial_coordinates
-        da_x, da_y = ds_state[xdim], ds_state[ydim]
-        assert da_x.ndim == da_y.ndim == 1
-        return CartesianGridShape(x=da_x.size, y=da_y.size)
+        n_x = np.unique(da_cat[xdim].values).size
+        n_y = np.unique(da_cat[ydim].values).size
+        return CartesianGridShape(x=n_x, y=n_y)
```
joeloskarsson, line 565: I don't think we should have this dual meaning of this method, returning the number of unique coordinates in x and y also for irregular grid stores. This also seems like changing the meaning of this method from the base class.

I think we should just say that this raises an error if the datastore is not on regular grid, and that then `num_grid_nodes` should be used.

neural_lam/datastore/mdp.py
```diff
@@ -570,3 +648,30 @@ def get_lat_lon(self, category: str) -> np.ndarray:
...
+        a cropped datastore holds fewer than ``grid_shape_state.x * .y`` and
+        unstacking it would pad the missing cells.
+
+        Returns
+        -------
+        bool
+            ``True`` when every cell of the ``x``/``y`` grid is present.
+        """
+        grid_shape = self.grid_shape_state
+        return self.num_grid_points == grid_shape.x * grid_shape.y
```
joeloskarsson, line 666: Related to my comment on `grid_shape_state`, I am not sure if this is a robust enough check. If we follow my suggestion to not allow for `self.grid_shape_state` for irregular grid stores, then we could not use it here. And more broadly, it is possible to have `num_grid_points = unique(x-coords) * unique(y-coords)` without a regular grid, e.g. if the x- and y-coordinates are not equidistantly spaced.

joeloskarsson (changes requested):
Thanks for the change! I think this would work. Now the main questions is just if we should make this Datastore refactor part of this PR or a separate one (also discussed on slack). Hopefully it would not be to much work to separate it out into a separate PR if needed.

I left some small comments about the design of the new combined base Datastore class, and how it is instantiated for MDP. But if this is moved to another PR, then these all really belong there. These do not relate to the boundary datastore part of this PR (I am happy with all the changes related to that). Added these comments here now, so we at least have them noted down.

sadamov:
@leifdenby do you think we should put the removal of the BaseRegularGridDatastore into a separate PR or no?

PR #636, opened 2026-05-11
Add boundary data overlay to evaluation plots

### LINKED ISSUE #108: Plotting Boundaries in `train_model.py --eval` not just interior

sadamov:
## Describe your changes

Wire the boundary datastore from `config.yaml` into `ForecasterModule` and the plotting pipeline so that evaluation plots (`--eval`) overlay boundary forcing data from a separate domain underneath the interior prediction and ground truth panels.

Changes:
- `vis.plot_on_axis` and `vis.plot_prediction` accept optional `boundary_da` and `boundary_datastore` parameters. When provided, boundary data is plotted underneath the interior field at alpha=0.5 and the map extent is expanded by a configurable `boundary_margin_degrees` margin (default 1°) around the interior domain.
- Boundary grid coordinates are read via `get_xy("forcing")` / `get_lat_lon("forcing")` so that boundary datastores without state variables (e.g. ERA5 forcing-only) work correctly.
- New `plotting` section in `config.yaml` (`NeuralLAMConfig.plotting`, a `PlottingConfig`): `boundary_datastore` names which entry in `datastores` supplies the overlay (defaults to the sole boundary datastore), `boundary_margin_degrees` controls the map extent margin, and `boundary_var_mapping` maps interior state variable names to boundary forcing feature names (falling back to same-name matching).
- `ForecasterModule.__init__` accepts an optional `datastore_boundary`. During `plot_examples`, raw (unstandardized) boundary forcing is loaded from the boundary datastore and passed through to `vis.plot_prediction`, matched per state variable via `boundary_var_mapping`.
- `train_model.py` passes `datastore_boundary` from config into `ForecasterModule`.
- Deprecate (rather than remove) the `crop_to_interior` keyword on `plot_prediction`/`plot_spatial_error`, emitting a warning and ignoring it, to avoid a breaking change for existing callers.
- Integration test with real DANRA + ERA5 boundary data (marked `@pytest.mark.slow`); fixed `mllam-data-prep` resolving a referenced datastore config relative to the process CWD instead of the config file's directory.
- README updated to document the boundary overlay behavior and the new `plotting` config options during evaluation.

**Depends on #635** (boundary datastore data loading). Please merge #635 first; after that the base branch of this PR can be updated to `main`.

## Issue Link

refs #108

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

closes #108


sadamov:
here are some examples for dummy data with and without boundary each:

boundary_plot_interior_only
boundary_plot_with_overlay

neural_lam/vis.py
```diff
@@ -240,6 +287,8 @@ def plot_prediction(
     boundary_alpha=0.7,
     crop_to_interior=True,
```
kshirajahere: the default eval path still crops to the interior. plot_examples() calls plot_prediction() without overriding crop_to_interior, so boundary data is drawn and then set_extent crops back to the interior; with a 1000 km boundary dummy, I saw the called extent stay at [5.755, 14.251, 53.693, 58.245] while the boundary extent is [0.964, 19.036, 51.277, 60.49]. When boundary_da is provided, either default to the boundary/union extent or explicitly pass crop_to_interior=False from plot_examples().

neural_lam/vis.py
```diff
@@ -84,6 +93,37 @@ def plot_on_axis(
     gl.xlabel_style = {"size": _TICK_SIZE}
     gl.ylabel_style = {"size": _TICK_SIZE}
 
+    # Plot boundary data underneath interior data when provided
+    if boundary_da is not None and boundary_datastore is not None:
+        b_lats_lons = boundary_datastore.get_lat_lon("state")
```
kshirajahere: this uses boundary_datastore.get_lat_lon("state"), but the PR description says boundary-only datastores without state variables are supported via forcing coordinates. Current #636 head also lacks #635’s later boundary-only MDP fixes. Use the forcing category here and add a test where the boundary datastore genuinely has no state

neural_lam/models/module.py
```diff
@@ -443,6 +465,19 @@ def plot_examples(self, batch, n_examples, split, prediction):
...
+                # Select boundary field for this timestep (if available)
+                boundary_da_t = None
+                if (
+                    da_boundary_forcing is not None
+                    and "time" in da_boundary_forcing.dims
+                ):
+                    target_time = np.array(
+                        time_slice[t_i - 1].cpu(), dtype="datetime64[ns]"
+                    )
+                    boundary_da_t = da_boundary_forcing.sel(
```
kshirajahere, line 782: method="nearest" can silently overlay the wrong boundary timestep when the boundary datastore has missing times or a different cadence. For verification plots this should be exact, or at least use an explicit tolerance tied to the datastore step length and fail/skip when no valid boundary time exists.

sadamov:
thanks @kshirajahere for the review, I'll have a look soon.
In the meantime I also implemented a "real-world" test-case using DANRA and ERA5 data across #635 and #636. The resulting plot looks like this: 
image

PS: Tests fail also on main right now because of missing torch wheel for 2.12.0.

kshirajahere:
The plots look amazing @sadamov :D

joeloskarsson:
I see the extent of the plot is now determined based on the full boundary size (please clarify if I missunderstood). I wonder if we want to give the user some more options here, for example if the boundary is very large one might want to only zoom in on the interior, with this covering most of the plot. Not sure where such an option best sits though.

sadamov:
> I see the extent of the plot is now determined based on the full boundary size (please clarify if I missunderstood). I wonder if we want to give the user some more options here, for example if the boundary is very large one might want to only zoom in on the interior, with this covering most of the plot. Not sure where such an option best sits though.

not quite. first ,there is a simple check (boundary data present yes/no), if no -> plot interior only. If yes -> then plot a 1° margin (in projection) around the interior. this 1° is independent of the full boundary size and can be set via `boundary_margin_degrees`. I did however not expose this to the user yet and it would have to be wired through ForecasterModule.plot_examples, if you wanted that.

neural_lam/vis.py
sadamov: i changed the logic here to a clearly defined margin in degrees (projection) around the interior box. I hope that is clearer to use

neural_lam/vis.py
sadamov: you are right! I did a bit of cheating with the inheritance of DummyBoundaryDatastore. Now I properly removed the state (3e6e246, part of #635) and added a plotting test for the no-state path (a1789b2).

sadamov: oh yeah and I also added the full era5 reanalysis dataset for testing with only forcing vars selected in the yaml config for mllam-data-prep. so now we should be fully covered on the forcing-only aspect

neural_lam/models/module.py
sadamov: I actually like this to be nearest. Often the boundary will have coarser temporal resolution (e.g. 6h) and than the interior e.g. 1h. Now, neural-lam allows to select any number of future and past timesteps from the buondary that are nearest to that interior timestep creating a window. And I think it makes sense to display the nearest boundary in the verification plot because that is the boundary forcing closest to the truth that neural-lam can use in the inference. So unless you have strong objections, i'd rather leave this as is.

sadamov: I added a comment in the code about this rationale: 7b82b07

joeloskarsson:
Aha, I see 😄 Yea I guess wiring through`boundary_margin_degrees` to the user is exactly what I was thinking about then. To follow all other plotting config, this should be an argparse flag. But this also feels like we are getting a ridiculous amount of flags 😮‍💨 Not sure if it is worth it. Maybe an option is to group this with the configuration of interior-boundary variable-mappings somehow? That is also plotting config really.

sadamov:
> Aha, I see 😄 Yea I guess wiring through`boundary_margin_degrees` to the user is exactly what I was thinking about then. To follow all other plotting config, this should be an argparse flag. But this also feels like we are getting a ridiculous amount of flags 😮‍💨 Not sure if it is worth it. Maybe an option is to group this with the configuration of interior-boundary variable-mappings somehow? That is also plotting config really.

I introduce a new plotting config here: https://github.com/mllam/neural-lam/pull/636/commits/d6acd0f992cf8a9b3b4d4a2d37cda1ef8e1df957 for both vars matching (similar to research branch I think) and boundary width.

tests/datastore_examples/mdp/era5_1000hPa_danra_100m_winds/era5.datastore.yaml
```diff
@@ -0,0 +1,108 @@
...
+        start: 2022-03-30T00:00
+        end: 2022-04-12T00:00
+  # Second-pass crop: keep only ERA5 points within `margin_width_degrees`
+  # of the interior dataset's convex hull. Follows the documented mllam
+  # pattern in `example.era5_cropped.yaml` -- the dim_mapping below stacks
+  # `[longitude, latitude]` directly into `grid_index` (no rename) so that
+  # convex-hull cropping can find lat/lon coords by name.
+  domain_cropping:
+    margin_width_degrees: 10
+    interior_dataset_config_path: danra.datastore.yaml
```
kshirajahere, line 37: the slow “real-world” plotting test currently fails before it reaches plotting because `mllam-data-prep` resolves this as `danra.datastore.yaml` from the process cwd, not from this config file’s directory. Running `python -m pytest tests/test_plotting.py::test_plot_prediction_with_era5_boundary --run-slow -q` from the repo root raises `FileNotFoundError: [Errno 2] No such file or directory: 'danra.datastore.yaml'`, so the integration test cited for DANRA + ERA5 is not actually runnable.

neural_lam/vis.py
```diff
@@ -483,12 +635,19 @@ def plot_prediction(
     title=None,
     vrange=None,
     boundary_alpha=0.7,
-    crop_to_interior=True,
     colorbar_label: str = "",
+    boundary_da=None,
+    boundary_datastore=None,
+    boundary_margin_degrees: float = 1.0,
```
kshirajahere, line 769: removing the existing `crop_to_interior` keyword from `plot_prediction`/`plot_spatial_error` is a breaking API change for user scripts. On current `main`, callers can pass `crop_to_interior=False`; on this branch that raises `TypeError: got an unexpected keyword argument 'crop_to_interior'`.

README.md
```diff
@@ -514,6 +526,23 @@ Some options specifically important for evaluation are:
 * `--n_example_pred`: Number of example predictions to plot during evaluation.
 * `--ar_steps_eval`: Number of time steps to unroll for during evaluation
 
+When a `datastore_boundary` is configured in `config.yaml`, evaluation plots
+will overlay the boundary forcing data underneath the interior prediction and
+ground truth panels, allowing visual inspection of the boundary conditions used
```
kshirajahere: this says the plotted boundary conditions are “used during the forecast”, but the current model path still ignores `boundary_features` in `ForecasterModule.common_step`; this PR loads boundary forcing separately for visualization in `plot_examples`.

kshirajahere (changes requested):
Fascinating work. Thanks for giving me the opportunity to review it @sadamov really grateful :) left some 3 comments which I think might be worth addressing

tests/datastore_examples/mdp/era5_1000hPa_danra_100m_winds/era5.datastore.yaml
sadamov: ah true I didn't consider how mdp determines the output path. now it's fixed and documented in some comments

neural_lam/vis.py
sadamov: ok, it's probably a bad idea to remove this. I added a deprecation warning instead.

README.md
sadamov: yes, here I was foreshadowing the upcoming PR that introduces boundary handling on the model side. I changed the comment to reflect the current state and this PR only.

sadamov:
This is now rebased onto #635 and is ready for a review, too. @joeloskarsson

PR #639, opened 2026-05-14
fix: resolve torch version from target wheel index in CI

### LINKED ISSUE #640: FIX : GPU CI fails when PyPI carries a torch version not yet in the cu128 wheel index
## Describe the bug

The GPU CI job fails when PyPI publishes a new torch version that the cu128 wheel index has not yet packaged. The "Determine PyTorch version" step resolves the version via:

```bash
TORCH_VERSION=$(python -m pip install --dry-run "." | grep "Would install" | grep -o 'torch-[0-9.]*' | awk -F'-' '{print $2}' | tail -n 1)
```

This queries **PyPI** and returns the latest available version (currently `2.12.0`). The subsequent step then tries to install that exact version from the cu128 index:

```bash
python -m pip install "torch==2.12.0" --index-url https://download.pytorch.org/whl/cu128
```

Which fails because cu128 only carries up to `2.11.0+cu128`:

```
ERROR: Could not find a version that satisfies the requirement torch==2.12.0
(from versions: 2.7.0+cu128, 2.7.1+cu128, 2.8.0+cu128, 2.9.0+cu128,
2.9.1+cu128, 2.10.0+cu128, 2.11.0+cu128)
```

## Steps to reproduce

Open any PR against `main` and observe the `pip + gpu` CI job failing at the "Install PyTorch" step.

## Expected behaviour

The CI should install the latest torch version available in the **target wheel index**, not the latest on PyPI.

## Context

This was not an issue before `torch 2.12.0` was released. PR #208 merged on 2026-05-11 when `2.11.0` was the latest on both PyPI and cu128, the existing logic worked. The problem surfaced on PR #507 a few days later after `2.12.0` appeared on PyPI.

This will affect every PR running GPU CI until the cu128 index catches up to PyPI.

## Proposed fix

PR #639 fixes this by reading the lower-bound constraint from `pyproject.toml` and querying the target wheel index directly to resolve the version to install, ensuring the pinned version always exists in the target index.

Sir-Sloth-The-Lazy:
## Describe your changes

The GPU CI job fails when PyPI carries a torch release that the cu128 wheel index does not yet have. The existing "Determine PyTorch version" step runs `pip install --dry-run "."` which queries **PyPI** and resolves the latest version (e.g. `2.12.0`). The subsequent install step then tries to fetch that exact version from the cu128 index, which only carries up to `2.11.0+cu128`, causing the error:

```
ERROR: Could not find a version that satisfies the requirement torch==2.12.0
(from versions: 2.7.0+cu128 ... 2.11.0+cu128)
```

This fix changes the version detection to:
1. Read the **lower-bound** torch constraint directly from `pyproject.toml` (`torch>=2.3.0`), avoiding PyPI resolution entirely.
2. Run a dry-run against the **target matrix wheel index** to find the highest version available there (e.g. `2.11.0` from cu128, or the current latest from the CPU index).

The resolved version is guaranteed to exist in the target index because it was discovered from that same index.

This was identified during GPU CI on PR #507. PR #208 was not affected because it merged before `torch 2.12.0` was published to PyPI, at that point `2.11.0` was the latest on both PyPI and cu128, so the existing logic worked. Any PR running CI now will hit this failure until the cu128 index catches up.

## Issue Link

closes #640 

## Type of change

- [x] I have placed in-line comments to clarify the intent of the hard-to-understand passages of my code

### DIFF
--- a/.github/workflows/install-and-test.yml
+++ b/.github/workflows/install-and-test.yml
@@ -62,9 +62,20 @@ jobs:
       - name: Determine PyTorch version for neural-lam (using pip dry-run)
         if: matrix.package_manager == 'pip'
         run: |
-          TORCH_VERSION=$(python -m pip install --dry-run "." | grep "Would install" | grep -o 'torch-[0-9.]*' | awk -F'-' '{print $2}' | tail -n 1)
-          echo "Torch version detected: $TORCH_VERSION"
-          echo "TORCH_VERSION=$TORCH_VERSION" >> $GITHUB_ENV
+          # Read the lower-bound torch version directly from pyproject.toml so we
+          # are not affected by PyPI resolving a version that the target wheel index
+          # (e.g. cu128) does not yet carry.
+          MIN_VERSION=$(grep -o 'torch>=[0-9.]*' pyproject.toml | grep -o '[0-9.]*$')
+          echo "Minimum torch version required: ${MIN_VERSION}"
+          # Find the latest version satisfying that minimum in the target index.
+          # This returns the highest version present in the wheel index (e.g. cu128)
+          # rather than the highest version on PyPI.
+          TORCH_VERSION=$(python -m pip install --dry-run "torch>=${MIN_VERSION}" \
+            --index-url ${{ matrix.torch_index }} 2>&1 \
+            | grep "Would install" | grep -o 'torch-[0-9.]*' \
+            | awk -F'-' '{print $2}' | tail -n 1)
+          echo "Torch version to install: ${TORCH_VERSION}"
+          echo "TORCH_VERSION=${TORCH_VERSION}" >> $GITHUB_ENV
 
       - name: Install PyTorch
         if : matrix.package_manager != 'uv'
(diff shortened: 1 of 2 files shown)


Sir-Sloth-The-Lazy:
@joeloskarsson this is the PR to solve the issue of CI Job failing. Hope this would solve the error discussed in PR #507  :)

joeloskarsson:
Thanks for putting this up so quickly! This should indeed solve the error, and it was really a misconfiguration before to not look at the correct index url for deciding the version.

There are two outstanding question to me for solving this:
1. This would mean that we will run the cpu and gpu tests on different versions of pytorch, right? Maybe not the optimal setup I fell, but also I have no good idea for how to 
2. Could we use a newer CUDA version than 12.8 for the testing? and that way be able to run the tests with the latest pytorch version and not have this issue at all. Need input from others with more experience on the GPU setup for this.

To avoid tests failing, I would not mind merging this right now, and sorting out these points in a (potential) follow-up PR. But will wait on input from someone else as well.

joeloskarsson (approved):
Let's go with this solution for now, to have tests working. Then we can refine in a follow-up PR if needed.

joeloskarsson:
I added the changelog entry just so that this could be merged quickly now :)

PR #489, opened 2026-03-22
feat: Add batched tensor expansion utilities for O(S) ensemble unroll…

Debadri-das:
## Describe your changes

This PR introduces a standalone batch-expanding PyTorch utility (`expand_ensemble_batch` and `fold_ensemble_batch`) to optimize tensor manipulation for Ensemble Lateral Boundary Conditions (LBCs) and probabilistic forecasting variables.

**Motivation and Context:** Currently, probabilistic processing of $S$ ensemble members often relies on sequential `for` loops. To support the upcoming model refactoring and probabilistic capabilities without introducing $O(S)$ sequential bottlenecks, these functions leverage `torch.repeat_interleave()` and `.view()` to dynamically flatten state tensors `(B, ...) -> (B * S, ...)`. This enables optimized parallelized operations over the GPU without permanently mutating standard AR model dimensions.

Dependencies: None. Built strictly with standard PyTorch tensor primitives.

## Issue Link

 #49 and #62 

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch .
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented (including return types and parameters)
- [ ] the code is easy to maintain

## Author checklist after completed review

- [ ] I have added a line to the CHANGELOG describing this change, in a section reflecting type of change (add section where missing):
[...]

PR #648, opened 2026-05-27
feat: add latent encoder/decoder infrastructure for Graph-EFM port

### LINKED ISSUE #62: Merge Graph-EFM model from `prob_model_lam` branch

Sir-Sloth-The-Lazy:
## Describe your changes

Adds `neural_lam/models/latent/` the encoder and decoder submodules that the probabilistic Graph-EFM model needs. This is **infrastructure-only**: no model uses these classes yet. They are consumed by the upcoming `GraphEFMPredictor` (`StepPredictor` subclass) which will close #62.

**New modules in `neural_lam/models/latent/`:**

| File | What it adds |
|---|---|
| `base_encoder.py` | `BaseLatentEncoder` : abstract base; handles isotropic / diagonal Gaussian output |
| `base_decoder.py` | `BaseGraphLatentDecoder` : abstract base; residual grid MLP + latent embedder + param map |
| `constant_encoder.py` | `ConstantLatentEncoder` : input-independent prior (used when `learn_prior=False`) |
| `graph_encoder.py` | `GraphLatentEncoder` : flat graph: grid → mesh via PropagationNet + InteractionNet stack |
| `graph_decoder.py` | `GraphLatentDecoder` : flat graph: grid + latent → grid via g2m / processor / m2g |
| `hi_graph_encoder.py` | `HiGraphLatentEncoder` : hierarchical mesh: propagates up to top level, reads out latent dist |
| `hi_graph_decoder.py` | `HiGraphLatentDecoder` : hierarchical mesh: up + latent fusion + down pass back to grid |

**Adaptations from `prob_model_lam` for the current `main` architecture:**

- `constants.GRID_STATE_DIM` (removed) → `num_state_vars` constructor arg on all decoders
- `from neural_lam.interaction_net import ...` → `from neural_lam.gnn_layers import ...`
- `GraphLatentDecoder.processor` unified with the other four GNN-seq constructions to use `utils.make_gnn_seq`, which also handles `processor_layers=0` gracefully
- `HiGraph{Encoder,Decoder}` raise `ValueError` for single-level meshes (where the latent would be silently ignored); points users to the flat variants

**Also adds to `neural_lam/utils.py`:**

- `IdentityModule`  pass-through `nn.Module` for multi-arg `pyg.nn.Sequential` pipelines
- `make_gnn_seq`  builds a `pyg.nn.Sequential` of `InteractionNet` layers, or `IdentityModule` when `num_gnn_layers=0`; lazy-imports `gnn_layers` to avoid the existing `gnn_layers → utils` circular dependency

**Open question flagged in `constant_encoder.py` docstring:** the static prior returns `Normal(mean=1, std=1)`  faithful to `prob_model_lam`  but the `--learn_prior` CLI help on that branch describes it as "mean 0". One of the two is wrong; will raise it separately with @joeloskarsson.

**Dependencies:**
- `PropagationNet` and `InteractionNet` are already on `main` (#507 merged).
- #226 which brings the ensemble metrics on the current main.
- #489 add ensemble branch fold utility

## Issue Link

Partially addresses #62 (prerequisite for the `GraphEFMPredictor` PR). 

## Type of change

- [x] I have requested a reviewer and an assignee

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

- [x] I have added a line to the CHANGELOG describing this change

## Checklist for assignee

- [x] the tests pass
- [x] the PR is assigned to the next milestone
- [x] author has added an entry to the changelog


Sir-Sloth-The-Lazy:
Pinging @joeloskarsson !

neural_lam/models/forecaster_module.py
joeloskarsson, line 1: Is this (and also forecaster, ar_forecaster and step_predictor) now added as duplicate files here? A bit confused about these additions? These files are placed differently on main?

neural_lam/models/latent/__init__.py
joeloskarsson, line 1: I think there is a discussion to be had if we want to use the encoder/decoder terminology here, since this means something different than the encoder/decoder used in the encode-process-decode framework of the current deterministic models. I think this should be fine, since these are explicitly Latent{Encoder, Decoder}, and we don't have any other classes called encoder/decoder right now. But wanted to make note of this, and open to other ideas.

Sir-Sloth-The-Lazy: I have this noted and will bring this up in future discussions , if ambiguity comes up !

neural_lam/models/forecaster_module.py
Sir-Sloth-The-Lazy: This is branched of the latest main , I don't think these are duplicates , however will tell you concertely in this weeks meeting.

joeloskarsson: Hmm, if I go to the "files changed" tab then I see these files as added in this PR? This is what confuses me.

neural_lam/models/latent/graph_encoder.py
```diff
@@ -0,0 +1,61 @@
...
+        g2m_edge_index,
+        m2m_edge_index,
+        hidden_dim,
+        processor_layers,
+        hidden_layers=1,
+        output_dist="isotropic",
+    ):
+        super().__init__(latent_dim, output_dist)
+
+        self.g2m_gnn = PropagationNet(
```
joeloskarsson: I think also here the gnn type should be set with the corresponding argparse flag.

neural_lam/models/latent/graph_encoder.py
```diff
@@ -0,0 +1,61 @@
...
+    on-mesh InteractionNet processor layers.
+    """
+
+    def __init__(
+        self,
+        latent_dim,
+        g2m_edge_index,
+        m2m_edge_index,
+        hidden_dim,
+        processor_layers,
```
joeloskarsson: This feels like a strange argument name to me. We don't really have a "processor" here, as this is all an encoder.

neural_lam/models/latent/constant_encoder.py
```diff
@@ -0,0 +1,35 @@
...
+    Latent encoder that returns a constant (input-independent) distribution.
+
+    Used as a non-learned prior in ``GraphEFM`` when ``learn_prior`` is
+    disabled. ``compute_dist_params`` returns a tensor of ones, so the
+    resulting Normal is ``Normal(mean=1, std=1)`` for ``output_dist=
+    "isotropic"`` and ``Normal(mean=1, std=softplus(1)+eps)`` for
+    ``output_dist="diagonal"``. (Note: the ``train_model.py`` CLI help on
+    ``prob_model_lam`` describes this prior as "mean 0"; the code itself
+    has always produced mean 1. Preserved as-is during the port for
+    behavioral parity — open question for upstream.)
```
joeloskarsson: Should be mean 0, the mean 1 is probably just a bug. Does not really change anything, just a constant offset, but makes more sense to use 0 mean.

neural_lam/models/latent/constant_encoder.py
```diff
@@ -0,0 +1,35 @@
...
+    def compute_dist_params(self, grid_rep, **kwargs):
+        """
+        Return constant parameters of shape (B, num_mesh_nodes, output_dim).
+        """
+        return torch.ones(
+            grid_rep.shape[0],
+            self.num_mesh_nodes,
+            self.output_dim,
+            device=grid_rep.device,
+        )
```
joeloskarsson:
```suggestion
        return torch.zeros(
            grid_rep.shape[0],
            self.num_mesh_nodes,
            self.output_dim,
            device=grid_rep.device,
        )
```

neural_lam/utils.py
```diff
@@ -389,6 +390,40 @@ def make_mlp(blueprint: list[int], layer_norm: bool = True) -> nn.Sequential:
...
+
+
+def make_gnn_seq(edge_index, num_gnn_layers, hidden_layers, hidden_dim):
+    """
+    Build a sequential stack of InteractionNet layers that propagates both
+    node and edge representations. Returns an IdentityModule if
+    num_gnn_layers is 0.
+    """
+    # First-party
+    from neural_lam.gnn_layers import InteractionNet
```
joeloskarsson: Can we make this general enough to not just support Inets?

neural_lam/utils.py
```diff
@@ -389,6 +390,40 @@ def make_mlp(blueprint: list[int], layer_norm: bool = True) -> nn.Sequential:
...
+
+    if num_gnn_layers == 0:
+        return IdentityModule()
```
joeloskarsson: This feels like not the best place to handle this. If we call this to build a gnn sequence, then we should at least request one layer. Otherwise there should be a check that makes this not being called at all.

neural_lam/models/latent/hi_graph_encoder.py
```diff
@@ -0,0 +1,120 @@
...
+            [
+                PropagationNet(
+                    edge_index,
+                    hidden_dim,
+                    hidden_layers=hidden_layers,
+                    update_edges=False,
+                )
+                for edge_index in mesh_up_edge_index
+            ]
+        )
```
joeloskarsson: Can we also here use argparse options for choosing the gnn types.

joeloskarsson (commented):
Had a look over the encoders now, decoders still TODO :)

Sir-Sloth-The-Lazy:
On it !

neural_lam/models/latent/graph_encoder.py
Sir-Sloth-The-Lazy: I've switched it to resolve the GNN via `get_gnn_class()` and added a `g2m_gnn_type` parameter wired to the existing `--g2m_gnn_type flag`, matching the deterministic models. Two things I'd like your opinion on before I finalize:

Default value. The original prob_model_lam encoder used `PropagationNet` for the g2m step, but `--g2m_gnn_type` defaults to `InteractionNet`. I've defaulted the new parameter to `InteractionNet` for consistency with the flag and the rest of the codebase, which does change the encoder's default behavior vs. the original port. Are you happy with that, or would you prefer keeping `PropagationNet` as the default here?

Shared vs. separate flag. Wiring this to `--g2m_gnn_type` means the latent encoder's g2m GNN and the step predictor's g2m GNN share a single flag and can't be configured independently. In the original Graph-EFM these differed (predictor g2m = InteractionNet, encoder g2m = PropagationNet). Do you think sharing the one flag is fine, or should the encoder get its own flag (e.g. `--latent_g2m_gnn_type`) to preserve that independence?

Sir-Sloth-The-Lazy:
I have pushed the latest changes , please have a look whenever you have time

neural_lam/models/forecaster_module.py
joeloskarsson: This is still a problem, there are all these extra files here.

neural_lam/models/latent/graph_encoder.py
joeloskarsson: Hmm, these are all very good questions. I feel like we are now hitting the point where the way we configure the GNN types is quite limited. But I am reluctant to here introduce some more complex system for configuring this.
In general all configurations are fine, the only constraint is that the mesh up-edges in the hierarchical encoders should really use propnets (they need to push information into the latent Z) and the up-edges in the decoder should really use inet (otherwise there is no use of Z at initialization).

I think the best solution might be to hard-code the GNN types mentioned above (this only affects the mesh-up edges), and then reuse the existing flags for all choices of which type to use for g2m/m2g. We can keep the InteractionNet default also for these encoders/decoder, to be consistent with the rest of the codebase.

neural_lam/models/latent/constant_encoder.py
```diff
@@ -0,0 +1,36 @@
...
+    Used as a non-learned prior in ``GraphEFM`` when ``learn_prior`` is
+    disabled. ``compute_dist_params`` returns a tensor of zeros, so the
+    resulting Normal is ``Normal(mean=0, std=1)`` for ``output_dist=
+    "isotropic"`` and ``Normal(mean=0, std=softplus(0)+eps)`` for
+    ``output_dist="diagonal"``. (Note: ``prob_model_lam`` returned a tensor
+    of ones here, giving mean 1, while its ``train_model.py`` CLI help
+    described the prior as "mean 0". The mean 1 was a bug -- it is only a
+    constant offset, but a mean-0 prior is what is intended, so the port
+    uses zeros.)
+    """
```
joeloskarsson: This comment needs some changes:
* Better to not say that this is used for a specific purpose in Graph-EFM. The docstring should describe what this class does, it is up to other parts of the code to use it as suitable. Can be in Graph-EFM, but coudl also be for any other purpose.
* Remove the note here. Do not leave meta-information from the work on this, that always ages poorly. Comments in the code should only describe the current code.

neural_lam/utils.py
```diff
@@ -389,6 +390,55 @@ def make_mlp(blueprint: list[int], layer_norm: bool = True) -> nn.Sequential:
...
+    """
+    Build a sequential stack of GNN layers that propagates both node and
+    edge representations. The layer type is set by ``gnn_type`` (any key in
+    ``gnn_layers.GNN_TYPES``, default ``InteractionNet``); all such layers
+    share the ``(send, rec, edge) -> (rec, edge)`` interface.
+
+    ``num_gnn_layers`` must be at least 1. Callers that want a no-op stage
+    (e.g. zero intra-level layers) should substitute an ``IdentityModule``
+    themselves rather than calling this with 0.
+    """
```
joeloskarsson: 1. This needs a more properly written docstring, following numpy style convention
2. In general I think most docstrings miss the "Parameters" header (https://numpydoc.readthedocs.io/en/latest/format.html#parameters). This is something we are trying to be stricter with going forward.

neural_lam/models/latent/graph_encoder.py
```diff
@@ -0,0 +1,66 @@
...
+            hidden_layers=hidden_layers,
+            update_edges=False,
+        )
+
+        self.m2m_gnns = (
+            utils.make_gnn_seq(
+                m2m_edge_index, m2m_layers, hidden_layers, hidden_dim
+            )
+            if m2m_layers > 0
+            else utils.IdentityModule()
```
joeloskarsson: I think it would be better to just do a check in the forward pass if we actually have any m2m GNNs, and avoid updating representations if not. We should be able to get rid of the IdentityModule all together. That would lead to more clear code, as it now looks like we always have some m2m GNNs updating representation even when we don't.

neural_lam/models/latent/hi_graph_encoder.py
joeloskarsson: See https://github.com/mllam/neural-lam/pull/648#discussion_r3358346704 for all discussion about this.

neural_lam/models/latent/constant_encoder.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __init__(self, latent_dim, num_mesh_nodes, output_dist="isotropic"):
+        super().__init__(latent_dim, output_dist)
+        self.num_mesh_nodes = num_mesh_nodes
+
+    def compute_dist_params(self, grid_rep, **kwargs):
+        """
+        Return constant parameters of shape (B, num_mesh_nodes, output_dim).
```
joeloskarsson: Improve docstring. Should also match (and potentially expand on) base class docstring.

neural_lam/models/latent/base_decoder.py
```diff
@@ -0,0 +1,96 @@
...
+
+class BaseGraphLatentDecoder(nn.Module):
+    """
+    Abstract decoder mapping a grid representation plus a latent sample on
+    mesh to the parameters of the next-state distribution on the grid.
+
+    Subclasses implement :meth:`combine_with_latent`, which fuses the latent
+    representation with the grid representation. The resulting features are
+    mapped to either ``num_state_vars`` outputs (mean only) or
+    ``2 * num_state_vars`` outputs (mean + softplus std) depending on
```
joeloskarsson:
```suggestion
    ``2 * num_state_vars`` outputs (mean, std) depending on
```

neural_lam/models/latent/base_decoder.py
joeloskarsson, line 1: It would be good to expand on the different parameters in the docstrings in this file, to make it easier to follow. Also applies to the other decoders inheriting from this.

neural_lam/models/latent/base_decoder.py
```diff
@@ -0,0 +1,96 @@
...
+        last_state: (B, num_grid_nodes, num_state_vars)
+        graph_emb: dict with at least ``g2m``, ``m2m``, ``m2g`` entries
+
+        Returns:
+        pred_mean: (B, num_grid_nodes, num_state_vars)
+        pred_std: (B, num_grid_nodes, num_state_vars) or ``None``
+        """
+        latent_emb = self.latent_embedder(latent_samples)
+
+        residual_grid_rep = grid_rep + self.grid_update_mlp(grid_rep)
```
joeloskarsson, line 151: Some comment clarifying the role of this grid update MLP would be useful.

neural_lam/models/latent/base_decoder.py
```diff
@@ -0,0 +1,96 @@
...
+
+        combined_grid_rep = self.combine_with_latent(
+            grid_rep, latent_emb, residual_grid_rep, graph_emb
+        )
+
+        state_params = self.param_map(combined_grid_rep)
+
+        if self.output_std:
+            mean_delta, std_raw = state_params.chunk(2, dim=-1)
+            pred_std = nn.functional.softplus(std_raw)
```
joeloskarsson, line 166: Add some comment explaining this chunking (why, how?)

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,680 @@
...
+)
+from ..base import StepPredictor
+
+
+class GraphEFM(StepPredictor):
+    """
+    Graph-based Ensemble Forecasting Model -- single-step predictor.
+
+    Port of ``prob_model_lam``'s ``GraphEFM`` (``forward`` is the source's
+    ``predict_step``) onto the ``StepPredictor`` interface. The predictor owns
```
joeloskarsson: Remove such mentions of this porting process. Instead reference the paper where people can read about the model.

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,680 @@
...
+    which carries its own g2m/processor/m2g GNNs, so the encode-process-decode
+    backbone of ``BaseGraphModel`` does not apply -- this extends
+    ``StepPredictor`` directly. It is self-contained: besides ``forward`` it
+    exposes the per-step ELBO pieces (``compute_step_loss`` ->
+    ``(likelihood_term, kl_term, pred_mean, pred_std)``) and the sampling
+    helpers used by a future rollout/ensemble module. Rollout, ELBO assembly,
+    ensemble logic and logging live outside the predictor.
+
+    One class handles both flat and hierarchical meshes, resolved at
+    construction from ``self.hierarchical`` (set by ``utils.load_graph``).
```
joeloskarsson: Unless you have a good argument for keeping this as one class, that instantiates encoders/decoders as needed, I think we should actually go with a different route here. In particular, either of:
1. Make Graph-EFM take the prior-model, encoder and decoder that it should use as input arguments. This feels a bit more programatic, but also requires quite some checking that the encoders/decoders given are actually compatible.
2. Have two subclasses for Graph-EFM (hierarchical) and Graph-EFM-MS (flat graph). Then the prior model (constant or learned encoder) could still be a boolean option to the constructor. In this option the model instantiates its own encoders/decoders, as now.

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,680 @@
...
+        self,
+        config: NeuralLAMConfig,
+        datastore: BaseDatastore,
+        graph_name: str = "hierarchical",
+        hidden_dim: int = 64,
+        hidden_layers: int = 1,
+        latent_dim: Optional[int] = None,
+        prior_processor_layers: int = 2,
+        encoder_processor_layers: int = 2,
+        processor_layers: int = 4,
```
joeloskarsson: Again the processor terminology is confusing here and we should avoid.

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,680 @@
...
+        learn_prior: bool = True,
+        prior_dist: str = "isotropic",
+        num_past_forcing_steps: int = 1,
+        num_future_forcing_steps: int = 1,
+        output_std: bool = False,
+        sample_obs_noise: bool = False,
+        output_clamping_lower: Optional[Dict[str, float]] = None,
+        output_clamping_upper: Optional[Dict[str, float]] = None,
+    ):
```
joeloskarsson, line 70: This really needs a docstring explaining all these args.

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,680 @@
...
+        latent_dim: Optional[int] = None,
+        prior_processor_layers: int = 2,
+        encoder_processor_layers: int = 2,
+        processor_layers: int = 4,
+        learn_prior: bool = True,
+        prior_dist: str = "isotropic",
+        num_past_forcing_steps: int = 1,
+        num_future_forcing_steps: int = 1,
+        output_std: bool = False,
+        sample_obs_noise: bool = False,
```
joeloskarsson: I think we should get rid of this option and just never do this. Without any form of correlation model sampling this noise is a terrible idea in practice, and having the option here might tempt users to try this (and waste time :))

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,680 @@
...
+        self.mlp_blueprint_end = [hidden_dim] * (hidden_layers + 1)
+        self.grid_prev_embedder = utils.make_mlp(
+            [self.grid_dim] + self.mlp_blueprint_end
+        )  # For states up to t-1
+        self.grid_current_embedder = utils.make_mlp(
+            [grid_current_dim] + self.mlp_blueprint_end
+        )  # For states including t
+        # Embedders for mesh edges
+        self.g2m_embedder = utils.make_mlp([g2m_dim] + self.mlp_blueprint_end)
+        self.m2g_embedder = utils.make_mlp([m2g_dim] + self.mlp_blueprint_end)
```
joeloskarsson: I believe that all of this is actually duplicate code (or very close to) from `BaseGraphModel`. This is the sort of thing that is good to keep in mind when porting this over, what overlap we have with the current codebase.

We should definitely get rid of such duplicate code, as this is bad for maintainability and overall structure. However exactly how to do this is not obvious. I'll leave it to you to come up with some ideas here, and am happy to discuss and bounce ideas. Directions to consider could be util functions, some shared class to inherit from, or even letting graph-efm be a BaseGraphModel (which might require re-defining exactly what this interface means).

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,680 @@
...
+        if not self.output_std:
+            pred_std = self.per_var_std  # (d_f,)
+
+        if self.sample_obs_noise:
+            return torch.distributions.Normal(pred_mean, pred_std).rsample()
+            # (B, num_grid_nodes, d_state)
+
+        return pred_mean  # (B, num_grid_nodes, d_state)
+
+    def embedd_current(
```
joeloskarsson: Would be good with more descriptive names for `embedd_current` and `embedd_all`.

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,680 @@
...
+        return likelihood_term, kl_term, pred_mean, pred_std
+
+    def forward(
+        self,
+        prev_state: torch.Tensor,
+        prev_prev_state: torch.Tensor,
+        forcing: torch.Tensor,
+    ) -> tuple[torch.Tensor, Optional[torch.Tensor]]:
+        """
+        Sample one time step prediction (source's ``predict_step``):
```
joeloskarsson: What does "(source's ``predict_step``)" mean? clarify

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,680 @@
...
+        prior_dist = self.prior_model(
+            grid_prev_emb, graph_emb=graph_emb
+        )  # (B, num_mesh_nodes, d_latent)
+
+        # Sample from prior
+        latent_samples = prior_dist.rsample()
+        # (B, num_mesh_nodes, d_latent)
+
+        # Compute reconstruction (decoder)
+        last_state = prev_state
```
joeloskarsson: why rename?

neural_lam/models/step_predictors/graph/graph_efm.py
joeloskarsson, line 1: There is some parts that I am unsure if they should belong to this class, and need discussion, including the `per_var_std` (in case it is not predicted), the KL-calculation and the likelihood calculation. I find it very hard to know if the methods using/computing these makes sense without seeing how these fit into the larger structure, i.e. how will the final loss that we optimize be derived. In general it feels strange to me that the step predictor would compute the likelihood, and we should only feed the target to this level if really strictly needed (computing the likelihood does need this). To get some clarity in this I think we need to either:
1. Wait with these specific parts until we have the other PR clearly defining the interfaces for loss computations in probabilistic models.
2. If you have a really clear idea, that I am just not seeing, you can outline it here or on slack 😄 and I can ask until I get it.

joeloskarsson (changes requested):
Good work with fixing the earlier things! I now did a _complete_ readthrough of all of this, which resulted in a number of comments, some small, some more substantial. There are also a few that require more discussion, so do expect us to have some back and forth around these. 

As you are fixing these things, please leave me some comments very briefly explaining the fix and link to the commit that fixes it (a good reason to fix one thing in each commit :)).

neural_lam/models/latent/graph_encoder.py
Sir-Sloth-The-Lazy: I've implemented this in 1396357:

**Hard-coded GNN types (no longer parameters):**

- `HiGraphLatentEncoder` mesh-up edges: `PropagationNet` as you said, these need to push grid information up into the latent readout.
- `HiGraphLatentDecoder` mesh-up edges: `InteractionNet` with a PropagationNet the residual path would carry the aggregated messages instead of the top-level receiver rep (= Z), leaving Z unused at initialization.
- `HiGraphLatentDecoder` mesh-down edges: `PropagationNet`  I hard-coded these too, by the symmetric argument: the downward pass has to push Z from the top level down through the hierarchy so it reaches the grid output. This matches the original Graph-EFM. Let me know if you'd rather keep these configurable since your comment only strictly required fixing the up-edges.
I added short comments at each hard-coded site explaining the constraint, so future readers don't "helpfully" parameterize them again.

**Everything else reuses the existing flags with the codebase-standard defaults:**

- g2m in both encoders and the decoders, and m2g in the decoders, remain configurable via `g2m_gnn_type/m2g_gnn_type`, now all defaulting to `InteractionNet` for consistency with the deterministic models (this changes the flat encoder's g2m and the decoders' m2g defaults vs. the original port, per your suggestion).
- `GraphEFM` accepts `g2m_gnn_type/m2g_gnn_type` and passes them through to the prior, encoder and decoder, so they'll wire directly to the existing `--g2m_gnn_type/--m2g_gnn_type` flags once the model is registered. There's no `mesh_up_gnn_type/mesh_down_gnn_type` on `GraphEFM` since those are now fixed; the existing flags keep affecting Hi-LAM only.

neural_lam/models/latent/constant_encoder.py
Sir-Sloth-The-Lazy: pushed as afe56d0

Sir-Sloth-The-Lazy: added it so that we can track the port , would have removed it later, just thought it would make it easier to review

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: Done , I will keep this in mind

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: Good catch  this was indeed near-duplicate of `BaseGraphModel`'s constructor. I dug into exactly how much overlaps before picking an approach, and went with extracted util functions rather than a shared base class or making Graph-EFM a `BaseGraphModel`. Reasoning below, happy to push further in either direction if you disagree.

What actually overlaps. Comparing the two constructors line by line, the genuinely shared logic is all mechanical setup, not behaviour:

- the graph load + register-buffers loop (identical)
- the `grid_input_dim` formula (identical arithmetic; just named `grid_dim` in EFM)
- `mlp_blueprint_end` and the `g2m/m2g` edge embedders (identical)

Everything that follows diverges: `BaseGraphModel` registers `diff_mean/diff_std` and builds the encode-process-decode backbone (`g2m_gnn`, `encoding_grid_mlp`, `m2g_gnn`, `output_map`) consumed by its concrete `forward`; Graph-EFM instead registers `per_var_std`, builds two grid embedders, and hands the `g2m/m2g` GNNs off to its prior/encoder/decoder. The two `forward` passes have no shared structure.

**Why util functions over the other two directions?** Since the only thing in common is constructor boilerplate (no shared behaviour to inherit), I didn't think either heavier option earned its cost here:

- _Making Graph-EFM a `BaseGraphModel`_  its forward is a concrete encode→process→decode pipeline with `process_step/embedd_mesh_nodes` hooks. Graph-EFM's computation (embed grid → prior samples latent on mesh → decode latent+grid) can't be expressed as that pipeline with a different `process_step`; it has an extra variational encoder and prior. Fitting it would mean making `BaseGraphModel.forward` abstract, which breaks the contract GraphLAM/HiLAM/HiLAMParallel rely on.

- **A shared intermediate base class** would add an inheritance layer to the MRO of all four existing deterministic models to host ~20 lines of setup that aren't behaviour. Restructuring working code to share a constructor felt like the wrong trade.
So I pulled the drift-prone bits into two helpers in `utils.py` and called them from both families:

- `load_and_register_graph(module, datastore, graph_name) -> bool` loads the graph, registers its tensors/`BufferLists` on the module, returns whether it's hierarchical.
- `grid_input_dim(datastore, grid_static_dim, num_past_forcing_steps, num_future_forcing_steps) -> int`  the shared grid-input-dim formula.
This removes the actual duplication (the parts that could silently diverge), and the win lands on BaseGraphModel too e.g. when the boundary-forcing input from #138 changes the grid input dim, both model families update in one place, which I think is the maintainability point you were getting at. I left the embedders inline (one-liners, and EFM's grid embedder genuinely differs) and didn't touch the class hierarchy. Graph-EFM keeps only its own `grid_current_dim = grid_dim + num_state_vars` locally, since the extra target-state input is EFM-specific.

Done in 963b3d9; the deterministic and EFM model tests both still pass.

That said  if you'd rather go all the way to a shared BaseGraphStepPredictor for the structure, these helpers compose cleanly into one and I'm happy to take it there. Let me know which you prefer.

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: This is the meta information from the port, was keeping this to keep a track of the port. Have removed it now.

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: Agreed, that was a pointless alias. It was more of a confusion on my end, cleanup was need sorry for this. In `forward`, `prev_state` is already X_t, the state the decoder adds its residual onto, so I now pass it straight to the decoder and dropped the `last_state = prev_state line` (0a53788). Note the similar-looking `last_state = prev_states[:, -1]` in `compute_step_loss` stays, since there it actually extracts `X_t` from the stacked `(B, 2, …)` tensor rather than renaming.

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: Rename embedd_all -> embedd_grid_and_graph (embeds the grid for states
up to t-1 plus the full graph) and embedd_current ->
embedd_grid_with_target (embeds the grid including the target state, for
the encoder), per review. Done in [6cd6a29](https://github.com/mllam/neural-lam/pull/648/commits/6cd6a297d661bb709726d1270f16aa7506d32b30)

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: Sampling uncorrelated Gaussian observation noise per grid node is not
useful in practice, so the option is removed entirely rather than left
to tempt users, per review. forward now returns the decoder mean
directly (the prediction is stochastic only through the latent sample)
and the trivial sample_next_state helper is dropped. [83dbe92](https://github.com/mllam/neural-lam/pull/648/commits/83dbe924074212bfaa1cbd8b68650f1831ae765a)

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: [5a2befc](https://github.com/mllam/neural-lam/pull/648/commits/5a2befc637b8c8ca973acfe5a549eff809b4bee3) Changed in this commit

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: Replace the single GraphEFM class that resolved flat vs hierarchical at
construction with an explicit class per graph type, per review:

- BaseGraphEFM: graph-type independent setup (graph loading, grid and
  grid-mesh edge embedders, per-variable std) and all shared behavior
  (forward, compute_step_loss, estimate_likelihood, sampling helpers).
  Validates the loaded graph against the subclass's
  requires_hierarchical and exposes an embedd_mesh hook used by
  embedd_all.
- GraphEFM: hierarchical mesh graphs; builds per-level mesh embedders
  and HiGraphLatentEncoder/HiGraphLatentDecoder modules.
- GraphEFMMS: flat (e.g. multi-scale) mesh graphs; builds the flat mesh
  embedders and GraphLatentEncoder/GraphLatentDecoder modules.

learn_prior remains a constructor flag on both subclasses. Tests select
the class per graph type and cover the graph-type mismatch error.

done in [`f5790f2`](https://github.com/mllam/neural-lam/pull/648/commits/f5790f20d0981280b1d6636f6305660b8cb814de)

neural_lam/models/latent/base_decoder.py
Sir-Sloth-The-Lazy: [`43beef9`](https://github.com/mllam/neural-lam/pull/648/commits/43beef91a4cc3c247d11c3b13e8242d8b2d04798) I have added the explaination in this commit. I will scan the codebase to find out more of such ambiguities

neural_lam/models/latent/base_decoder.py
Sir-Sloth-The-Lazy: I have tried to do this again. Please review this again

neural_lam/models/latent/constant_encoder.py
Sir-Sloth-The-Lazy: I have tried to bring numpy style docstrings

neural_lam/models/latent/graph_encoder.py
Sir-Sloth-The-Lazy: When m2m_layers / intra_level_layers is 0, the latent modules now set
the corresponding GNN attribute to None and skip the update in the
forward pass, instead of routing representations through a no-op
IdentityModule. This makes it clear from the forward code that no
processing happens in that case. IdentityModule is removed from utils.

The hierarchical up/down loops index levels explicitly to accommodate
the conditional; outputs are unchanged (verified bit-identical against
the previous implementation). done in commit [`9926779`](https://github.com/mllam/neural-lam/pull/648/commits/99267799355051d35fadeb3fa2a04eb409183904)

neural_lam/utils.py
Sir-Sloth-The-Lazy: Add proper Parameters/Returns sections following the numpydoc
convention, per review. [`70c2b22`](https://github.com/mllam/neural-lam/pull/648/commits/70c2b228dbf92b71981157b2df1a16dd795fdd46)

neural_lam/models/forecaster_module.py
joeloskarsson: Fixed now :)

neural_lam/models/latent/graph_encoder.py
joeloskarsson: Great, that's perfect! Can we also expand on the description for these flags in `train_model.py` so that the user understand what the different flags impacts and not. Sufficient to add a note for mesh_up_gnn_type/mesh_down_gnn_type that these do not apply to the probabilistic Graph-EFM model.

neural_lam/models/latent/constant_encoder.py
joeloskarsson: For comments that are only relevant for the review process, but should not be in the final code, it is better to put it as a comment on github. Then I will see it as I am reviewing, but do not get the idea that it should stay in the codebase.

neural_lam/utils.py
joeloskarsson: Nice! Since we now recently also have the [canonical dimension names](https://github.com/mllam/neural-lam/tree/main#dimension-glossary), do take a look through that those are followed everywhere here. There might be a need to introduce additional ones here as well (e.g. `num_members`), that can then be added to the list in the readme.

neural_lam/models/latent/graph_encoder.py
joeloskarsson: I see, that created some major re-writing of the iterating through the levels of the hierarchy in the hi-encoder/decoder. I think this is good though, looks a bit cleaner than the original setup.

neural_lam/models/step_predictors/graph/graph_efm.py
joeloskarsson: Good, this looks like a nice solution. I have some comments on specific parts, but will add those as new comments where relevant and resolve this thread.

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,960 @@
...
+    graph type. Concrete subclasses are specific to a graph type (declared
+    by ``requires_hierarchical``): their constructors build the mesh
+    embedders and the prior/encoder/decoder latent modules, and they
+    implement :meth:`embedd_mesh`. See :class:`GraphEFM` (hierarchical
+    graph) and :class:`GraphEFMMS` (flat graph).
+    """
+
+    # Whether the concrete subclass requires a hierarchical mesh graph
+    requires_hierarchical: bool
+
```
joeloskarsson: I don't think this fits within the base class. The checking of graph compatability should be done in the sub-classes (they know which type of graph they expect). A good separation of concerns here would be that this base class is not aware of the notion of hierarchical/flat graphs.

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,960 @@
...
+        super().__init__(
+            datastore=datastore,
+            output_std=output_std,
+            output_clamping_lower=output_clamping_lower,
+            output_clamping_upper=output_clamping_upper,
+        )
+
+        # Load graph with static features.
+        # NOTE: (IMPORTANT!) mesh nodes MUST have the first
+        # num_mesh_nodes indices.
```
joeloskarsson: This does not seem like the right place for this note (and it will no longer be relevant soon, so we can just remove it).

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,960 @@
...
+                hidden_layers=hidden_layers,
+                g2m_gnn_type=g2m_gnn_type,
+                output_dist=prior_dist,
+            )
+        else:
+            self.prior_model = ConstantLatentEncoder(
+                latent_dim=latent_dim,
+                num_mesh_nodes=self.num_mesh_nodes,
+                output_dist=prior_dist,
+            )
```
joeloskarsson: To avoid duplicate code, could we put the prior construction and related arguments in the base class? For the constant prior it should be the same for both subclasses, and then we can just delegate the instantiation of the graph-dependent learnable prior to a method that the subclasses have to implement.

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: I am working on a basic interface that would consume the work we have done on this PR but I am not sure I should put it here or as a seperate PR ?😅

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,960 @@
...
+                for emb, edge_feat in zip(self.m2m_embedders, self.m2m_features)
+            ]
+        else:
+            # Need a placeholder otherwise, just use raw features
+            mesh_emb["m2m"] = list(self.m2m_features)
+
+        return mesh_emb
+
+
+class GraphEFMMS(BaseGraphEFM):
```
joeloskarsson: I think we need a better class name for this 😅 EFMMS is to much abbreviation. Maybe `GraphEFMMultiScale` or `MultiScaleGraphEFM`.

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: I think, it would be easier to discuss on the design I have in mind with some basic code. If you feel I should put the plan here first before any code, then please say the word. I will do that first

joeloskarsson: Could you start with describing the idea (which parts go where, which methods sit where) either as a comment here or on slack? I think the most important things is to discuss the core idea just with very concise descriptions of what classes and methods are involved.
Then for actually implementing this interface, I believe it would be best as a separate PR featuring just the interface at first, without the instantiation of Graph-EFM using it.

joeloskarsson: I think having methods with inputs + what they return can be good, just no need for writing out the bodies of any functions.

joeloskarsson (changes requested):
Good job with all the nice fixes! I went over some of the previous comments and resolved most of them. Still have some left, but submitting these comments for now just so you can see them.

neural_lam/utils.py
```diff
@@ -436,6 +442,83 @@ def loads_file(fn: str) -> Any:
...
+    for name, attr_value in graph_ldict.items():
+        # Make BufferLists module members and register tensors as buffers
+        if isinstance(attr_value, torch.Tensor):
+            module.register_buffer(name, attr_value, persistent=False)
+        else:
+            setattr(module, name, attr_value)
+    return hierarchical
+
+
+def grid_input_dim(
```
joeloskarsson:
```suggestion
def compute_grid_input_dim(
```
Suggestion for more specific name

neural_lam/utils.py
```diff
@@ -436,6 +442,83 @@ def loads_file(fn: str) -> Any:
...
+    datastore: "BaseDatastore",
+    grid_static_dim: int,
```
joeloskarsson: Can't we get the grid static dim from the datastore? Then there should be no need to feed it as a separate argument

neural_lam/models/step_predictors/graph/graph_efm.py
joeloskarsson: Thanks for this nice overview, that outlines the options well I think. I agree with the choice to make this util functions.
Something that I do see now thought is that with the additions here the `utils.py` file really is growing to an unreasonable length, and has all kinds of different things in it. I think we would be better of making `utils` a separate module, like `models`, and having util functions related to different things in different files within there. That might also fix the annoying circular import requiring the `TYPE_CHECKING` conditional. I would not want to include such a utils module refactor in this PR (it's a separate issue really), but would be happy if you could open an issue about this and either open a separate PR for that refactoring or work with someone else in the community interested to take it on.

joeloskarsson (changes requested):
Now I have gone over everything, and resolved or commented on things needing further attention 😄This is coming along super nicely I think, and especially nice that we are finding some shortcomings of both the earlier graph-efm code and in the repo, and able to resolve much of that.

Most things left are smaller fixes. The bigger question is how this will fit with the larger Module/Forecaster framework. This should be the priority to hammer out now. I have also not yet looked much at the methods in GraphEFM that deal with loss computation, simply because I expect these might change a bit once we figure out the details of the probabilistic model training interface.

Sir-Sloth-The-Lazy:
Thank you @joeloskarsson  for the review 🤩. I would be happy to take the work of utils refactor. I will be opening another PR soon solving that issue

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: fixed this in the commit 76c0dab

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: removed it in 3bec96c commit

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: Fixed the duplicate code problem in the commit ab005c3

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: 00ec230 (the GraphEFMMS → GraphEFMMultiScale rename)

neural_lam/utils.py
Sir-Sloth-The-Lazy: This is a mistake, sorry fixed it dc6600e

neural_lam/models/step_predictors/graph/graph_efm.py
joeloskarsson: Now this is ongoing in #682

neural_lam/models/step_predictors/graph/graph_efm.py
joeloskarsson: Continuing discussion in #685

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,1115 @@
...
+            )
+        else:
+            self.per_var_std = None
+
+        # Compute indices and define clamping functions. GraphEFM's forward
+        # never clamps (the decoder outputs the full next state), so these are
+        # inert -- accepted for interface parity with other StepPredictors.
+        self.prepare_clamping_params(datastore)
+
+    def build_prior(
```
joeloskarsson: Do we actually need this function? It seems to be called identically by both subclasses, so could we not just do the prior setup immediately in the base class constructor, calling `self.build_learnable_prior` just as here, which will be the part handled by the subclasses.

neural_lam/models/latent/graph_encoder.py
joeloskarsson: Don't forget this thread, still think we should add this to train_model.py

neural_lam/utils.py
joeloskarsson: Also here, do go over the dimension names (or let me know if already done). I found for example some `d_h` that should probably be `hidden_dim`.

joeloskarsson (changes requested):
Nice work on this! Only one small thing with the build_prior setup, and a reminders about the two unresolved threads from above. 

The thing left now is just the loss/likelihood part. But I think that will have to wait until we have started to settle #685. I will look at that next.

neural_lam/models/latent/graph_encoder.py
Sir-Sloth-The-Lazy: Did this in commit f5c5bc0

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: I've implemented this in 961d1f2: prior construction now lives entirely in `BaseGraphEFM.__init__`, not in each subclass.

`BaseGraphEFM.__init__` now takes `latent_dim/learn_prior/prior_dist/prior_layers/g2m_gnn_type` directly, and:

- Validates the loaded graph against a new `expects_hierarchical` class attribute (`GraphEFM.expects_hierarchical = True`, `GraphEFMMultiScale.expects_hierarchical = False`), replacing the two near-identical `raise ValueError` blocks that used to live in each subclass.
- Derives `self.num_mesh_nodes` generically from `self.hierarchical` (top mesh level for hierarchical graphs, all mesh nodes for flat graphs) previously duplicated per subclass.
- Resolves the `latent_dim` default and stores it as `self.latent_dim`.
- Calls `self.build_prior(...)` itself at the end of `__init__`, which still delegates to the subclass's `build_learnable_prior` for the learned case.

`GraphEFM` and `GraphEFMMultiScale` just forward these params to `super().__init__()` and reuse `self.latent_dim` when building their encoder/decoder. Each subclass now only builds its mesh embedders and encoder/decoder, and implements `embedd_mesh`/`build_learnable_prior` no more constructor-level duplication.

This was safe to reorder because `load_and_register_graph` (called at the top of the base constructor) already registers all edge indices as buffers before `build_learnable_prior` runs, so nothing the prior needs depends on code that used to run later in the subclass constructors.

Verified: all 14 `test_graph_efm_predictor.py` tests pass (including the graph-type-mismatch `ValueError` test), the broader latent/step-predictor suite passes, and pre-commit (black/isort/flake8/mypy/interrogate) is clean.

Sir-Sloth-The-Lazy:
@observingClouds, please spare sometime and review this work as well

neural_lam/models/step_predictors/graph/graph_efm.py
joeloskarsson: That makes more sense to me, that the base class can build the constant/learnable prior network.

1. I still think we should now get rid of this method, and just add the code directly in the constructor. It is very little code and the method is just called once I think.
2. We earlier removed a property in the base class signaling if a hierarchical/non-hierarchical graph is expected. This seems re-introduced here. I still think that the hierarchical/non-hierarchical graph should only be a concern for the subclasses, and prefer the ValueErrors as before, in the subclass constructors. 
I understand that we need to know `self.num_mesh_nodes` in order to build the constant prior. But a more general solution would be to have an abstract property/method like `latent_spatial_dim` that the subclasses then have to implement. They can then use the knowledge of their corresponding graph constructions to give the correct value there, which will correspond to the number of mesh nodes in the top/only mesh graph level.

neural_lam/utils.py
```diff
@@ -24,6 +25,10 @@
 # Local
 from .custom_loggers import CustomMLFlowLogger
 
+if TYPE_CHECKING:
+    # Imported only for type checking to avoid a runtime import cycle
+    # Local
+    from .datastore import BaseDatastore
```
joeloskarsson: Remember to check if we can remove this now that #682 is merged.

joeloskarsson:
I am pretty happy with this now, just 
1. figuring out some details around the prior network instantiation above.
2. need to get this back up to date with main and look over tests and linting.

@observingClouds do you want to give this a look over as well before we merge it in? Recall that this will not just be the model components that make up the Graph-EFM model as a step-predictor, not any of the mechanisms to actually train it and compute the related losses. That will be a follow up building on #700, instantiating the interface there.

observingClouds:
I'm good with this once your concerns/comments are addressed

neural_lam/utils.py
Sir-Sloth-The-Lazy: Done ! in 780941f

Sir-Sloth-The-Lazy:
Completed the requested changes and resolved the conflict ! @joeloskarsson @observingClouds

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,1000 @@
...
+            [grid_current_dim] + self.mlp_blueprint_end
+        )  # For states including t
+        # Embedders for mesh edges
+        self.g2m_embedder = utils.make_mlp([g2m_dim] + self.mlp_blueprint_end)
+        self.m2g_embedder = utils.make_mlp([m2g_dim] + self.mlp_blueprint_end)
+
+        # Compute indices and define clamping functions. GraphEFM's forward
+        # never clamps (the decoder outputs the full next state), so these are
+        # inert -- accepted for interface parity with other StepPredictors.
+        self.prepare_clamping_params(datastore)
```
joeloskarsson: I realized first now that we don't actually do any clamping here. I think that we actually should (for the predicted mean), so that clamping can be used similarly to for deterministic models.

This will require some thought, since the addition with the previous state now sits in the decoder. Might have to move that residual connection to this class, and then call `get_clamped_new_state`.

Sorry for not spotting this earlier 😅

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -0,0 +1,1000 @@
...
+        }
+
+        if self.embedd_m2m:
+            mesh_emb["m2m"] = [
+                self.expand_to_batch(emb(edge_feat), batch_size)
+                for emb, edge_feat in zip(self.m2m_embedders, self.m2m_features)
+            ]
+        else:
+            # Need a placeholder otherwise, just use raw features
+            mesh_emb["m2m"] = list(self.m2m_features)
```
joeloskarsson: This seems dangerous, can we use just an empty list instead?

joeloskarsson:
Gave this a complete read through before merging. Unfortunately I found something related to the clamping that I had missed earlier :/ (see above). But otherwise this all looks good. So if we can just sort out clamping for this model then I will go ahead and merge.

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: Thank you for the detailed review ! I should've been more discrete with this sorry for that fixed it in [fdc333d](https://github.com/mllam/neural-lam/pull/648/commits/fdc333d06f262581d690d6d0d40003efbb86479b)

neural_lam/models/step_predictors/graph/graph_efm.py
Sir-Sloth-The-Lazy: Sorry for this fixed it in [`79371f5`](https://github.com/mllam/neural-lam/pull/648/commits/79371f518be63641f630403bfa6ecf69c1fcbaf6)

Sir-Sloth-The-Lazy:
@joeloskarsson I have completed the changes !

neural_lam/models/step_predictors/graph/graph_efm.py
joeloskarsson: That's a good fix. Smaller of a change than I expected, just needed to extract the residual connection from the decoder.

I think we should also do the same rescaling as the deterministic model: https://github.com/mllam/neural-lam/blob/94bdb2e0ec65a2a814337d2e477a281af3a854a2/neural_lam/models/step_predictors/graph/base.py#L338-L339

This was not done originally in Graph-EFM, but I think it is a good idea. This ensures that all decoder-outputs can stay at a similar scale, even across variables that vary at very different timescales.

Apart from this I am happy with everything, so if you agree with the point above and change that then I can go ahead and merge.

joeloskarsson (approved):
Thanks for sorting out the clamping as well! This all looks good now, so I will go ahead and merge.

PR #673, opened 2026-06-17
Enforce 100% type-hint coverage and align typing to PEP 585/604

### LINKED ISSUE #448: Add type hints to methods in interaction_net.py
## Summary
The methods in `neural_lam/interaction_net.py` currently lack type annotations.
mypy is already configured in pre-commit but cannot fully validate this module
without proper type hints.

## Methods affected
- `InteractionNet.forward`
- `InteractionNet.message`
- `InteractionNet.aggregate`
- `SplitMLPs.forward`

## Proposed Fix
Add `torch.Tensor`, `Optional[torch.Tensor]`, `Tuple` and `Union` type hints
to all method signatures and return types.

## Why This Matters
- mypy is already set up in `.pre-commit-config.yaml`
- Type hints improve contributor experience and catch bugs early
- Consistent with type hints already added in `metrics.py` (see #446)

### LINKED ISSUE #670: [Maintenance] Enforce 100% type-hint coverage and align typing to PEP 585/604
We have been adding type hints module-by-module through a long series of issues
(#617, #619, #621, #624, #630, #646, #446, #454, and the four currently in flight:
#662 / #664 / #666 / #668). That piecemeal approach got us most of the way, time to complete the journey :) 

**1. Reach 100% annotation coverage.** Every function and method in `neural_lam/` gets full
parameter and return annotations. `mypy --disallow-untyped-defs --disallow-incomplete-defs`
currently flags 43 gaps across 15 files (run it for the live list):

```
12  models/module.py
 6  datastore/npyfilesmeps/compute_standardization_stats.py
 4  datastore/npyfilesmeps/store.py
 3  train_model.py
 3  models/step_predictors/graph/hi_lam.py
 2  weather_dataset.py
 2  models/step_predictors/graph/hierarchical.py
 2  models/step_predictors/graph/base.py        # fixed by #667
 2  models/step_predictors/base.py              # fixed by #665
 2  datastore/plot_example.py
 1  vis.py
 1  utils.py
 1  models/step_predictors/graph/hi_lam_parallel.py
 1  models/step_predictors/graph/graph_lam.py   # fixed by #669
 1  datastore/__init__.py
```

**2. Align to modern conventions (PEP 585 / 604).** Repo-wide rewrite:
- `Optional[X]` -> `X | None`
- `Union[A, B]` -> `A | B`
- `Dict` / `List` / `Tuple` -> `dict` / `list` / `tuple`
- drop the now-unused `from typing import Dict, List, Tuple, Union, Optional` imports

This is mechanical and can be auto-applied with `pyupgrade --py310-plus` (or ruff `UP` rules
if #614 lands first); review the diff afterwards.

**3. Enforce going forward.** Add a `[tool.mypy]` block and turn on the coverage gate so an
untyped def fails pre-commit/CI, the type analog of `interrogate`:

```toml
[tool.mypy]
python_version = "3.10"
disallow_untyped_defs = true
disallow_incomplete_defs = true
```

_Note_:

- Tests and docs dirs stay out of scope, consistent with the `interrogate` exclude list.

GiGiKoneti:
## Describe your changes

- Enforce 100% type-hint coverage across the `neural_lam/` codebase by annotating all remaining parameters and return values.
- Align all type annotations with PEP 585 (using native generics like `list`, `dict`, and `tuple`) and PEP 604 (using the union operator `|`).
- Clean up unused imports and resolve all `flake8` line-length issues.
- Enforce strict checks via a mypy gate (`disallow_untyped_defs`/`disallow_incomplete_defs`) in `pyproject.toml`.

## Issue Link

closes #670

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass


GiGiKoneti:
| File Path | Description of Edits |
| :--- | :--- |
| CHANGELOG.md | Added a changelog entry to the Maintenance section detailing codebase-wide type hinting and PEP 585/604 compliance. |
| neural_lam/config.py | Annotated parsing methods and fields, and aligned standard containers to native generic types. |
| neural_lam/create_graph.py | Type-hinted entrypoint functions and CLI setup helpers. |
| neural_lam/datastore/__init__.py | Type-annotated store initialization factory helpers. |
| neural_lam/datastore/base.py | Fully annotated all interface and abstract methods on `BaseDatastore`. |
| neural_lam/datastore/mdp.py | Type-hinted the `MDPDatastore` class methods. |
| neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py | Type-hinted statistical aggregator methods and removed unused imports. |
| neural_lam/datastore/npyfilesmeps/store.py | Added parameter and return type annotations to file-based store methods. |
| neural_lam/datastore/plot_example.py | Annotated dataset plotting functions. |
| neural_lam/gnn_layers.py | Aligned existing types to PEP 585/604. |
| neural_lam/models/forecasters/autoregressive.py | Aligned typing signatures to PEP 585/604. |
| neural_lam/models/module.py | Extensively annotated Lightning components and wrapped long expressions to resolve flake8 line-length errors. |
| neural_lam/models/step_predictors/base.py | Fully annotated parameter and return types for base predictor interfaces. |
| neural_lam/models/step_predictors/graph/base.py | Annotated base graph GNN encoding, processing, and decoding methods. |
| neural_lam/models/step_predictors/graph/graph_lam.py | Annotated predictor forward passes and step configurations. |
| neural_lam/models/step_predictors/graph/hi_lam.py | Added full typing annotations to hierarchical graph model methods. |
| neural_lam/models/step_predictors/graph/hi_lam_parallel.py | Added full typing annotations to parallel hierarchical graph model methods. |
| neural_lam/models/step_predictors/graph/hierarchical.py | Fully type-hinted the hierarchical processing operations. |
| neural_lam/train_model.py | Type-hinted main execution flow functions. |
| neural_lam/utils.py | Type-hinted helper functions and introduced typing guards to avoid circular import errors. |
| neural_lam/vis.py | Fully type-hinted standard matplotlib visualization helpers. |
| neural_lam/weather_dataset.py | Type-hinted datamodule and dataset structures. |

GiGiKoneti:
@sadamov 
i think it is ready for review .... Take your time... keep me updated if you need any corrections !

sadamov (commented):
Thanks @GiGiKoneti. Since #670 is type-hints only: could we keep executed code identical to `main` and lean on `cast` / `# type: ignore` (and the narrowing `assert`s you already added) where the checker needs a hint? Inline notes below, this looks good, I just invested a few hours to get this review right. Moving towards fully typed (and enforced) is not a small step!

On the class-level attribute blocks: most attrs are plain `self.x = ...` in `__init__` that mypy already infers, so I've suggested keeping only the ones it can't (`register_buffer` tensors, graph-dict-loaded features, and the subclass narrowings) and dropping the rest. 

Non-blocking: the `T | None` we own is narrowed two ways, `assert ... is not None` (`pred_std`, `self.da_state`) and nine `# type: ignore[union-attr]` on `self.logger`; I'd pick one for the cases we control and keep the coded ignores for the genuinely dynamic ones (`self.hparams.*`, `log_image`).

neural_lam/datastore/mdp.py
```diff
@@ -567,3 +571,11 @@ def get_lat_lon(self, category: str) -> np.ndarray:
...
         coords = np.stack((lon.values, lat.values), axis=1)
         return coords
+
+    @cached_property
+    def state_feature_weights_values(self) -> List[float]:
+        """
+        Return the weights for each state feature as a list of floats.
+        """
+        n_features = self.get_num_data_vars("state")
+        return [1.0 / n_features] * n_features
```
sadamov: New implementation, out of scope for #670 and unused (`get_state_feature_weighting` doesn't call it), so I suggest to drop it.

```suggestion
```

neural_lam/datastore/npyfilesmeps/store.py
```diff
@@ -842,3 +866,13 @@ def coords_projection(self) -> ccrs.Projection:
...
+
+    @functools.cached_property
+    def state_feature_weights_values(self) -> list[float]:
+        """Return the weights for each state feature as a list of floats."""
+        weights = np.load(self.root_path / "static" / "parameter_weights.npy")
+        if self._remove_state_features_with_index:
+            mask = np.ones_like(weights, dtype=bool)
+            mask[self._remove_state_features_with_index] = False
+            weights = weights[mask]
+        return weights.tolist()
```
sadamov: Same here, drop it and let's do datastore-owned weights in their own issue.

```suggestion
```

neural_lam/datastore/plot_example.py
```diff
@@ -73,7 +79,7 @@ def plot_example_from_datastore(
 
     crs = datastore.coords_projection
     col_wrap = min(4, int(da[col].count()))
-    g = da.plot(
+    g = da.plot.pcolormesh(
```
sadamov: `da.plot.pcolormesh` is a behaviour change, revert to `main`'s dispatch.

```suggestion
    g = da.plot(
```

neural_lam/vis.py
```diff
@@ -767,7 +772,8 @@ def plot_spatial_error(
         pad=0.02,
     )
     cbar.ax.tick_params(labelsize=_TICK_SIZE)
-    cbar.formatter.set_powerlimits((-3, 3))
+    if hasattr(cbar.formatter, "set_powerlimits"):
+        getattr(cbar.formatter, "set_powerlimits")((-3, 3))
```
sadamov: The `hasattr` guard silently skips the powerlimits, keep the direct call.

```suggestion
    cbar.formatter.set_powerlimits((-3, 3))  # type: ignore[attr-defined]
```

neural_lam/utils.py
```diff
@@ -732,9 +745,9 @@ def inverse_softplus(
     ``torch.clamp`` will zero the gradients near the bounds, but values this
     close to zero or ``threshold / beta`` already have negligible gradients.
     """
-    x_clamped = torch.clamp(
-        x, min=torch.log(torch.tensor(1e-6 + 1)) / beta, max=threshold / beta
-    )
+    min_val = torch.tensor(math.log(1e-6 + 1) / beta, device=x.device)
+    max_val = torch.tensor(threshold / beta, device=x.device)
+    x_clamped = torch.clamp(x, min=min_val, max=max_val)
```
sadamov: Revert `inverse_softplus` to `main` (the `device=` change alters behaviour).

```suggestion
    x_clamped = torch.clamp(
        x, min=torch.log(torch.tensor(1e-6 + 1)) / beta, max=threshold / beta
    )
```

neural_lam/utils.py
```diff
@@ -1,15 +1,20 @@
 """Utility helpers shared across Neural-LAM training and evaluation."""
 
 # Standard library
+import argparse
 import datetime
+import math
```
sadamov: Drop the now-unused `import math`.

```suggestion
```

neural_lam/utils.py
```diff
@@ -582,7 +592,7 @@ def log_on_rank_zero(
     **kwargs : Any
         Keyword arguments passed to the logger.
     """
-    if rank_zero_only.rank == 0:
+    if getattr(rank_zero_only, "rank", 0) == 0:
```
sadamov: `getattr` changes the missing-attribute behaviour, revert.

```suggestion
    if rank_zero_only.rank == 0:
```

neural_lam/datastore/npyfilesmeps/store.py
```diff
@@ -473,9 +487,13 @@ def _get_single_timeseries_dataarray(
         x = xs[:, 0]  # Unique x-coordinates (changes along the first axis)
         y = ys[0, :]  # Unique y-coordinates (changes along the second axis)
         for d in dims:
+            coord_values: Any
             if d == "elapsed_forecast_duration":
-                coord_values = self.step_length * np.arange(self._num_timesteps)
+                coord_values = np.array(
+                    [self.step_length * i for i in range(self._num_timesteps)]
+                )
```
sadamov: Revert the coord build to `main` (the list-comp changes the array dtype).

```suggestion
                coord_values = self.step_length * np.arange(self._num_timesteps)
```

neural_lam/create_graph.py
```diff
@@ -65,7 +64,7 @@ def plot_graph(
     edge_lines = np.stack((from_pos, to_pos), axis=1)
     axis.add_collection(
         matplotlib.collections.LineCollection(
-            edge_lines, lw=0.4, colors="black", zorder=1
+            list(edge_lines), lw=0.4, colors="black", zorder=1
```
sadamov: `LineCollection` takes the array directly, drop the `list(...)`.

```suggestion
            edge_lines, lw=0.4, colors="black", zorder=1
```

neural_lam/datastore/npyfilesmeps/store.py
```diff
@@ -10,7 +10,7 @@
 from datetime import timedelta
 from functools import cached_property
 from pathlib import Path
-from typing import List, Optional
+from typing import Any, List, Optional, cast
```
sadamov: Finish PEP 585/604: drop the unused `List`/`Optional` here and convert `Optional[str]` (235) and the `List[str]` returns (630/648/668). Same is still pending in `mdp.py`: the `from typing import List, Optional, Union` (line 10), `Union[str, Path]` (39), `List[str]` (179/198/217), `Optional[str]` (256) and `Union[xr.DataArray, None]` (258).

```suggestion
from typing import Any, cast
```

neural_lam/config.py
```diff
@@ -174,7 +172,7 @@ class InvalidConfigError(Exception):
 
 def load_config_and_datastore(
     config_path: str,
-) -> tuple[NeuralLAMConfig, Union[MDPDatastore, NpyFilesDatastoreMEPS]]:
+) -> tuple[NeuralLAMConfig, BaseDatastore]:
```
sadamov: Keep `main`'s contract modernised (one cast in the loader drops the three call-site casts).

```suggestion
) -> tuple[NeuralLAMConfig, MDPDatastore | NpyFilesDatastoreMEPS]:
```

pyproject.toml
```diff
@@ -51,7 +51,15 @@ gpu = ["torch>=2.12,<2.13"]        # CUDA 13.0, default GPU build
 gpu-cu128 = ["torch>=2.11,<2.12"]  # CUDA 12.8, last torch series with cu128 wheels
 
 [dependency-groups]
-dev = ["pre-commit>=3.8.0", "pytest>=8.3.2", "pooch>=1.8.2"]
+dev = [
+    "pre-commit>=3.8.0",
+    "pytest>=8.3.2",
+    "pooch>=1.8.2",
+    "mypy>=2.1.0",
```
sadamov: #670 item 3, the enforcement gate, is missing: you add `mypy` here but not the `[tool.mypy]` block that makes an untyped def fail, so the coverage is neither enforced nor verifiable. Could we add (after `[tool.codespell]`):

```toml
[tool.mypy]
python_version = "3.10"
disallow_untyped_defs = true
disallow_incomplete_defs = true
ignore_missing_imports = true
exclude = ["^tests/", "^docs/", "^build/"]
```

and scope the existing pre-commit mypy hook so it doesn't choke on `tests/` (mypy's `exclude` is ignored for files passed explicitly, only `files:` filters the hook):

```yaml
      - id: mypy
        additional_dependencies: [types-PyYAML, types-Pillow, types-tqdm]
        files: ^neural_lam/
```

CHANGELOG.md
```diff
@@ -74,6 +74,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Maintenance
 
+- Enforce 100% type-hint coverage across neural_lam/ and align all type annotations with PEP 585 and PEP 604 [\#670](https://github.com/mllam/neural-lam/issues/670) @GiGiKoneti
```
sadamov: Link the PR not the issue (matches the #669 / #663 lines), and mention the gate that now does the enforcing.

```suggestion
- Enforce 100% type-hint coverage across `neural_lam/` via a `[tool.mypy]` gate (`disallow_untyped_defs`/`disallow_incomplete_defs`) and align all type annotations with PEP 585 and PEP 604 [\#673](https://github.com/mllam/neural-lam/pull/673) @GiGiKoneti
```

neural_lam/datastore/base.py
```diff
@@ -598,12 +598,12 @@ def stack_grid_coords(
         Parameters
         ----------
         da_or_ds : xr.DataArray or xr.Dataset
-            The dataarray or dataset to stack the grid coordinates of.
+             The dataarray or dataset to stack the grid coordinates of.
```
sadamov: Stray extra leading space (same on 605).

```suggestion
            The dataarray or dataset to stack the grid coordinates of.
```

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -22,7 +23,20 @@
...
     """Wrap :class:`WeatherDataset` to pad samples for distributed runners."""
 
-    def __init__(self, base_dataset, world_size, batch_size):
+    base_dataset: WeatherDataset
+    world_size: int
+    batch_size: int
+    total_samples: int
+    padded_samples: int
+    original_indices: list[int]
+    padded_indices: list[int]
```
sadamov: All seven are plain `__init__` assignments, drop the block.

```suggestion
```

neural_lam/gnn_layers.py
```diff
@@ -18,6 +17,11 @@ class InteractionNet(pyg.nn.MessagePassing):
     from Battaglia et al. (2016)
     """
 
+    num_rec: torch.Tensor | int
+    edge_mlp: nn.Module
+    aggr_mlp: nn.Module
+    update_edges: bool
```
sadamov: `num_rec` and the two MLPs are plain assignments, drop the block.

```suggestion
```

neural_lam/models/step_predictors/graph/hi_lam.py
```diff
@@ -20,6 +21,11 @@ class HiLAM(BaseHiGraphModel):
     The Hi-LAM model from Oskarsson et al. (2023)
     """
 
+    mesh_down_gnns: nn.ModuleList
+    mesh_down_same_gnns: nn.ModuleList
+    mesh_up_gnns: nn.ModuleList
+    mesh_up_same_gnns: nn.ModuleList
```
sadamov: The four `ModuleList`s are plain assignments, drop the block.

```suggestion
```

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
```diff
@@ -23,6 +24,14 @@ class HiLAMParallel(BaseHiGraphModel):
     of Hi-LAM.
     """
 
+    edge_split_sections: list[int]
+    processor: (
+        pyg.nn.Sequential
+        | Callable[
+            [torch.Tensor, torch.Tensor], tuple[torch.Tensor, torch.Tensor]
+        ]
+    )
```
sadamov: Both plain assignments, drop the block (and the `from typing import Callable` it needed).

```suggestion
```

neural_lam/weather_dataset.py
```diff
@@ -22,6 +22,15 @@ class WeatherDataset(torch.utils.data.Dataset):
...
     """
 
+    split: str
+    ar_steps: int
+    datastore: BaseDatastore
+    num_past_forcing_steps: int
+    num_future_forcing_steps: int
+    load_single_member: bool
+    da_state: xr.DataArray
+    da_forcing: xr.DataArray | None
```
sadamov: All eight are plain assignments, drop the block.

```suggestion
```

neural_lam/models/module.py
```diff
@@ -33,11 +33,33 @@ class ForecasterModule(pl.LightningModule):
...
+    loss: Any
+    val_metrics: dict[str, list]
+    test_metrics: dict[str, list]
+    restore_opt: bool
+    n_example_pred: int
+    create_gif: bool
+    plotted_examples: int
+    spatial_loss_maps: list[Any]
+    time_step_int: int
+    time_step_unit: str
```
sadamov: Keep only the eight registered buffers, drop the other thirteen.

```suggestion
    interior_mask_bool: torch.Tensor
    per_var_std: torch.Tensor | None
    state_mean: torch.Tensor
    state_std: torch.Tensor
    forcing_mean: torch.Tensor | None
    forcing_std: torch.Tensor | None
    forcing_mean_tiled: torch.Tensor | None
    forcing_std_tiled: torch.Tensor | None
```

neural_lam/models/step_predictors/base.py
```diff
@@ -18,6 +19,32 @@ class StepPredictor(nn.Module, ABC):
...
+    clamp_lower_idx: torch.Tensor
+    clamp_upper_idx: torch.Tensor
+
+    # Callables for clamping
+    clamp_lower_upper: Callable[[torch.Tensor], torch.Tensor]
+    clamp_lower: Callable[[torch.Tensor], torch.Tensor]
+    clamp_upper: Callable[[torch.Tensor], torch.Tensor]
+    inverse_clamp_lower_upper: Callable[[torch.Tensor], torch.Tensor]
+    inverse_clamp_lower: Callable[[torch.Tensor], torch.Tensor]
+    inverse_clamp_upper: Callable[[torch.Tensor], torch.Tensor]
```
sadamov: Keep the ten clamp/feature buffers, drop the scalars and the six lambda `Callable`s (and the `collections.abc.Callable` import).

```suggestion
    grid_static_features: torch.Tensor
    state_mean: torch.Tensor
    state_std: torch.Tensor
    sigmoid_lower_lims: torch.Tensor
    sigmoid_upper_lims: torch.Tensor
    softplus_lower_lims: torch.Tensor
    softplus_upper_lims: torch.Tensor
    clamp_lower_upper_idx: torch.Tensor
    clamp_lower_idx: torch.Tensor
    clamp_upper_idx: torch.Tensor
```

neural_lam/models/step_predictors/graph/base.py
```diff
@@ -18,6 +19,37 @@ class BaseGraphModel(StepPredictor):
...
+    output_map: nn.Module
+
+    # Graph attributes registered as buffers
+    g2m_features: torch.Tensor
+    m2g_features: torch.Tensor
+    g2m_edge_index: torch.Tensor
+    m2g_edge_index: torch.Tensor
+    mesh_static_features: torch.Tensor | list[torch.Tensor]
+    m2m_features: torch.Tensor | list[torch.Tensor]
+    m2m_edge_index: torch.Tensor | list[torch.Tensor]
```
sadamov: Keep `diff_mean/std` and the seven graph-dict features, drop the scalars and embedders (and the `from torch import nn` import).

```suggestion
    diff_mean: torch.Tensor
    diff_std: torch.Tensor
    g2m_features: torch.Tensor
    m2g_features: torch.Tensor
    g2m_edge_index: torch.Tensor
    m2g_edge_index: torch.Tensor
    mesh_static_features: torch.Tensor | list[torch.Tensor]
    m2m_features: torch.Tensor | list[torch.Tensor]
    m2m_edge_index: torch.Tensor | list[torch.Tensor]
```

neural_lam/models/step_predictors/graph/graph_lam.py
```diff
@@ -21,6 +22,13 @@ class GraphLAM(BaseGraphModel):
     Oskarsson et al. (2023).
     """
 
+    mesh_static_features: torch.Tensor
+    m2m_features: torch.Tensor
+    m2m_edge_index: torch.Tensor
+    mesh_embedder: nn.Module
+    m2m_embedder: nn.Module
+    processor: nn.Module
```
sadamov: Keep the three buffer narrowings, drop the embedders/processor (and `from torch import nn`).

```suggestion
    mesh_static_features: torch.Tensor
    m2m_features: torch.Tensor
    m2m_edge_index: torch.Tensor
```

neural_lam/models/step_predictors/graph/hierarchical.py
```diff
@@ -17,6 +18,26 @@ class BaseHiGraphModel(BaseGraphModel):
...
+    mesh_read_gnns: nn.ModuleList
+
+    # Buffers or graph features that are lists of tensors in hierarchical models
+    mesh_static_features: list[torch.Tensor]
+    m2m_features: list[torch.Tensor]
+    mesh_up_features: list[torch.Tensor]
+    mesh_down_features: list[torch.Tensor]
+    m2m_edge_index: list[torch.Tensor]
+    mesh_up_edge_index: list[torch.Tensor]
+    mesh_down_edge_index: list[torch.Tensor]
```
sadamov: Keep the seven list narrowings, drop the gnn-type scalars and embedders.

```suggestion
    mesh_static_features: list[torch.Tensor]
    m2m_features: list[torch.Tensor]
    mesh_up_features: list[torch.Tensor]
    mesh_down_features: list[torch.Tensor]
    m2m_edge_index: list[torch.Tensor]
    mesh_up_edge_index: list[torch.Tensor]
    mesh_down_edge_index: list[torch.Tensor]
```

pyproject.toml
```diff
@@ -51,7 +51,15 @@ gpu = ["torch>=2.12,<2.13"]        # CUDA 13.0, default GPU build
...
+    "types-pyyaml>=6.0.12.20260518",
+    "types-pillow>=10.2.0.20240822",
+    "types-tqdm>=4.68.0.20260608",
```
sadamov: These date-pinned stub versions are over-specified: they are dev-only tooling, the lock already pins exact versions for reproducibility, and you always want the newest stubs. I suggest to drop the floors (keeping `mypy>=2.1.0` is fine).

```suggestion
    "types-pyyaml",
    "types-pillow",
    "types-tqdm",
```

GiGiKoneti:
Thanks @sadamov  for investing the time into this detailed review; all requested changes have been resolved.

* The unused datastore properties have been dropped.
* Class-level type annotations have been stripped to only retain buffers and subclass narrowings.
* Code changes in `plot_example.py` and `utils.py` have been reverted to keep runtime behavior identical to main.
* Floors on development type stubs in `pyproject.toml` have been removed.

All local checks and test suites pass successfully.

sadamov:
I think you forgot to unhide the 14+ collapsed inline suggestions above ^^ or did you not agree with them?

GiGiKoneti:
> I think you forgot to unhide the 14+ collapsed inline suggestions above ^^ or did you not agree with them?

I had few doubts earlier actually and also forgot later on ..nvm I figured out things..can u review these implementations please

* Class-level type annotations have been stripped to only retain buffers, and the floor dependencies on dev stubs are dropped.
* Code changes in `plot_example.py`, `utils.py`, and datastore coordinate arrays have been reverted to keep runtime behavior identical to main.
* A strict `[tool.mypy]` gate (`disallow_untyped_defs` and `disallow_incomplete_defs`) has been configured, and the PR has been squashed into a single clean commit.

GiGiKoneti:
@sadamov 
just  added a strict `[tool.mypy]` gate as well as requested

sadamov:
@GiGiKoneti nice! could you again make sure to use the full PR template please.
Similar to #252 we should merge this as soon as possible to avoid rebasing over and over.
But introducing mandatory typing is a major change... :thinking: we could ask in slack if everyone agrees to put this in v0.7.0 without dev-meeting. what do you think?

PS: I only just pushed five `List` instances in docstrings to be fully aligned, not directly part of the PR (https://github.com/mllam/neural-lam/pull/673/commits/1c10f1d695ca4390580979a46cab09f620b6d266)

GiGiKoneti:
I updated the pull request description to use the full template and pulled in your docstring alignment commit.

I agree with asking on Slack to include this in v0.7.0 without a dev-meeting so we can merge it quickly.

I'm sure these type hints will help the active work going on

sadamov (approved):
Approving, thanks @GiGiKoneti, this is in good shape. I pushed fda55de to finish PEP 585/604 in the few modules the earlier pass had not touched (metrics, custom_loggers, plot_graph, npyfilesmeps/config, and the base docstrings), so the codebase-wide claim now holds.

One thing for a follow-up rather than a blocker. The gate is coverage-only right now: the pre-commit mypy hook runs isolated with no torch/numpy/xarray, so `ignore_missing_imports` turns every library type into `Any`. It proves every def is annotated (the #670 ask) but will not catch a wrong annotation.

Pointing mypy at the synced deps (`mypy --python-version 3.12 neural_lam`) finds 7 errors:
- `store.py:492` `step_length * np.arange(...)` -> `timedelta * ndarray`
- `utils.py:594` `rank_zero_only.rank` -> overload has no `.rank`
- `utils.py:747` `torch.clamp(x, min=<Tensor>, max=<float>)` -> no matching overload
- `plot_example.py:82` `da.plot(...)` -> xarray plot-accessor `__call__`
- `create_graph.py:67` `LineCollection(edge_lines)` -> ndarray not a Sequence
- `gnn_layers.py:95`/`:103` `edge_mlp`/`aggr_mlp` -> `SplitMLPs` vs `Sequential`

The first five are exactly the lines we reverted to keep runtime identical to main, so the clean fix is a targeted `# type: ignore[code]` each, like `set_powerlimits` in `vis.py`. The gnn_layers one wants the `edge_mlp: nn.Module` / `aggr_mlp: nn.Module` annotations back, a subset of the block I asked to drop.

So the open question: do we want the gate to actually type-check against the deps (install them in the hook or a separate CI step), or keep it coverage-only for now? My lean is to fix this now for v0.7.0.

GiGiKoneti:
@sadamov Let's merge this as the coverage gate for v0.7.0 to secure the 100% annotation baseline without further scope creep. I will open a follow-up issue to configure the strict dependency type-check and address the 7 remaining errors, keeping the `type: ignore` behavioral suppressions logically separated from this milestone.

sadamov:
@GiGiKoneti sounds good, in the follow-up we can also try to properly implement typing for the 6 cases above and not just ignore it. even if that means that the code behavior slightly changes.

GiGiKoneti:
@sadamov Should I create the follow-up issue for this now?

GiGiKoneti:
@sadamov When is our next dev meet btw  ?

sadamov:
@GiGiKoneti let's wait until we discuss typing at the next dev meeting on July 13. no need to keep the PR up to date with main until then

sadamov:
this should consider https://github.com/mllam/neural-lam/pull/688 which will fix some code issues that are also being flagged by Ty. preferrable 688 goes in first

GiGiKoneti:
Hi @sadamov! 

I have updated the branch by merging your `ty` type-checker branch into `chore/type-hint-coverage`. 
To resolve the two GNN type errors flagged by `ty` under `train_model.py`, I also merged the GNN kwargs fix from #688 locally. 

All checks are green:
* All pre-commit hooks (including the new `ty` check) pass successfully.
* The entire pytest test suite (248 tests) runs and passes without errors.

The PR is fully updated and ready for review/merging!

.pre-commit-config.yaml
```diff
@@ -43,12 +43,16 @@ repos:
         description: Ensure documentation coverage stays perfect
         pass_filenames: false
         args: ["-c", "pyproject.toml", "neural_lam"]
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.19.0
+  - repo: local
```
observingClouds: Great that you migrate to ty in this PR as well and make use of the performance improvements. To pin a specific version and keep all linters in the same framework, could you change here to the ty pre-commit hook? (https://github.com/astral-sh/ty-pre-commit)

pyproject.toml
```diff
@@ -51,7 +51,12 @@ gpu = ["torch>=2.12,<2.13"]        # CUDA 13.0, default GPU build
 gpu-cu128 = ["torch>=2.11,<2.12"]  # CUDA 12.8, last torch series with cu128 wheels
 
 [dependency-groups]
-dev = ["pre-commit>=3.8.0", "pytest>=8.3.2", "pooch>=1.8.2"]
+dev = [
+    "pre-commit>=3.8.0",
+    "pytest>=8.3.2",
+    "pooch>=1.8.2",
+    "ty>=0.0.59",
```
observingClouds: `ty` should not be needed when switching to its pre-commit hook (see comment below)

observingClouds (changes requested):

GiGiKoneti:
@observingClouds 

Updated the PR per feedback:
- Switched to the official `astral-sh/ty-pre-commit` hook (`v0.0.69`) in `.pre-commit-config.yaml`.
- Removed `ty` from `pyproject.toml` dev dependencies.
- Added `[tool.ty.src] exclude = ["tests/", "docs/"]` to scope `ty` checking to production code.
- Unskipped `ty` in `.github/workflows/pre-commit.yml` and removed the standalone `uv run ty` step.

All pre-commit hooks, pytest suites, and CI checks are now green.

.gitignore
```diff
@@ -11,6 +11,7 @@ sweeps
 *.html
 *.zarr
 *slurm*
+gigilocal/
```
observingClouds: This seems user-specific ;)

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -174,14 +189,14 @@ def save_stats(
     filename_prefix : str
         Prefix (e.g., ``"parameter"`` or ``"diff"``) for saved tensors.
     """
-    means = (
-        torch.stack(means) if len(means) > 1 else means[0]
+    means_tensor = (
+        torch.stack(list(means)) if len(means) > 1 else means[0]
```
observingClouds: if `means` and `squares` are of type list, why do we need to wrap them here again in `list()`?

observingClouds:
@GiGiKoneti, I had a quick look besides the `ty` introduction and a few additional comments.

.gitignore
GiGiKoneti: Removed  from .

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
GiGiKoneti: Removed the redundant list() wrapping in torch.stack().

GiGiKoneti:
Thanks @observingClouds! Addressed both inline comments and pushed the updates.

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -196,14 +211,16 @@ def save_stats(
 
     if len(flux_means) == 0:
         return
-    flux_means = (
-        torch.stack(flux_means) if len(flux_means) > 1 else flux_means[0]
+    flux_means_tensor = (
+        torch.stack(list(flux_means)) if len(flux_means) > 1 else flux_means[0]
```
observingClouds: so, you removed the list wrapper based on my previous comment, I guess the wrapping here and at other places is then also not needed.

pyproject.toml
```diff
@@ -121,6 +125,12 @@ per-file-ignores = [
 [tool.codespell]
 skip = "requirements/*"
 
+[tool.ty.src]
+exclude = ["tests/", "docs/", "gigilocal/"]
```
observingClouds: now your local directory came in here.

observingClouds (changes requested):
Thanks for the prompt adjustments. I have a few follow up comments.

pyproject.toml
GiGiKoneti: Removed — sorry about that.

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
GiGiKoneti: Good catch, removed remaining list() wrappers from flux_means/flux_squares too.

neural_lam/datastore/npyfilesmeps/store.py
observingClouds: here it states list of int, while the type hint is ndarray. Do we need to include/change the type or the docstring?

observingClouds: type annotations and docstring differ slightly

observingClouds: split can be None according to the new annotations. if this is the case, please document here that it can be optional and what the default behaviour is

neural_lam/datastore/npyfilesmeps/store.py
```diff
@@ -314,7 +328,9 @@ def get_dataarray(
         da = da.rename(dict(feature=f"{category}_feature"))
 
         # stack the [x, y] dimensions into a `grid_index` dimension
-        da = self.stack_grid_coords(da)
+        da_stacked = self.stack_grid_coords(da)
+        assert isinstance(da_stacked, xr.DataArray)
```
observingClouds: Testing the type in a get method looks incorrect to me

neural_lam/datastore/plot_example.py
```diff
@@ -47,21 +50,24 @@ def plot_example_from_datastore(
         Matplotlib figure object.
     """
     da = datastore.get_dataarray(category=category, split=split)
+    assert da is not None
     if standardize:
         da_stats = datastore.get_standardization_dataarray(category=category)
         da = (da - da_stats[f"{category}_mean"]) / da_stats[f"{category}_std"]
-    da = datastore.unstack_grid_coords(da)
+    da_unstacked = datastore.unstack_grid_coords(da)
+    assert isinstance(da_unstacked, xr.DataArray)
```
observingClouds: is this assert needed here?

neural_lam/datastore/plot_example.py
```diff
@@ -73,7 +79,7 @@ def plot_example_from_datastore(
 
     crs = datastore.coords_projection
     col_wrap = min(4, int(da[col].count()))
-    g = da.plot(
+    g = da.plot(  # ty: ignore[missing-argument]
```
observingClouds, line 79: what is missing here? is it just a `kwargs`? If it is just a missing argument, could we just add it instead of ignoring it?

neural_lam/models/step_predictors/graph/graph_efm.py
```diff
@@ -790,11 +798,14 @@ class GraphEFMMultiScale(BaseGraphEFM):
     Graph-based Ensemble Forecasting Model on a flat mesh graph
     (e.g. a multi-scale graph).
 
-    The latent variable lives on the mesh nodes. The prior and variational
-    encoder are ``GraphLatentEncoder``s and the decoder is a
-    ``GraphLatentDecoder``.
+    The latent variable lives on the (single) level of the mesh graph. The
```
observingClouds, line 801: @sadamov this might be worth checking with your expertise

neural_lam/models/step_predictors/graph/hi_lam.py
```diff
@@ -337,6 +352,11 @@ def hi_processor_step(
             self.mesh_up_gnns,
             self.mesh_up_same_gnns,
         ):
+            assert isinstance(down_gnns, nn.ModuleList)
+            assert isinstance(down_same_gnns, nn.ModuleList)
+            assert isinstance(up_gnns, nn.ModuleList)
+            assert isinstance(up_same_gnns, nn.ModuleList)
+
```
observingClouds: This looks like the wrong place to check the type. This should probably tested in the init method

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
```diff
@@ -182,7 +189,7 @@ def hi_processor_step(
             mesh_rep_levels, dim=1
         )  # (B, num_mesh_nodes, hidden_dim)
         mesh_edge_rep = torch.cat(
-            mesh_same_rep + mesh_up_rep + mesh_down_rep, axis=1
+            mesh_same_rep + mesh_up_rep + mesh_down_rep, dim=1
```
observingClouds, line 192: @sadamov , @joeloskarsson is this a bugfix that needs special mention? I am surprised this worked earlier as I cannot seem to find that `axis` was ever an argument, and the default `dim=0`.

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
```diff
@@ -196,13 +203,15 @@ def hi_processor_step(
...
         )
 
-        mesh_same_rep = mesh_edge_rep_sections[: self.num_levels]
-        mesh_up_rep = mesh_edge_rep_sections[
-            self.num_levels : self.num_levels + (self.num_levels - 1)
-        ]
-        mesh_down_rep = mesh_edge_rep_sections[
-            self.num_levels + (self.num_levels - 1) :
-        ]  # Last are down edges
+        mesh_same_rep = list(mesh_edge_rep_sections[: self.num_levels])
```
observingClouds, line 206: is this another case of unnecessary `list()`?

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
```diff
@@ -196,13 +203,15 @@ def hi_processor_step(
...
+        mesh_up_rep = list(
```
observingClouds, line 207: and how about this `list()`?

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
```diff
@@ -196,13 +203,15 @@ def hi_processor_step(
...
+            mesh_edge_rep_sections[
+                self.num_levels : self.num_levels + (self.num_levels - 1)
+            ]
+        )
+        mesh_down_rep = list(
```
observingClouds, line 212: and here?

neural_lam/models/module.py
```diff
@@ -404,6 +431,7 @@ def all_gather_cat(self, tensor_to_gather):
             The concatenated tensor from all processes.
         """
         gathered = self.all_gather(tensor_to_gather)
+        assert isinstance(gathered, torch.Tensor)
```
observingClouds, line 434: is this type check needed here because `.all_gather` could return None?

neural_lam/utils/logging.py
```diff
@@ -117,7 +124,7 @@ def setup_training_logger(
         return pl.loggers.WandbLogger(
             project=args.logger_project,
             name=None if args.wandb_id else run_name,
-            config=dict(training=vars(args), datastore=datastore._config),
+            config=dict(training=vars(args), datastore=datastore.config),
```
observingClouds, line 127: oh, is this another bug that we are fixing here?

observingClouds (changes requested):
Hi @GiGiKoneti,
Thanks for addressing my latest comments. As this is quite an extensive PR that touches a lot of files, I did review it in pieces. With my latest comments now, I should have addressed it all. I tagged @sadamov and @joeloskarsson for specific sections to also get their attention. I think we are getting close to approval. At least from my point of view 🤣

neural_lam/datastore/npyfilesmeps/store.py
GiGiKoneti: Updated docstring to  to match the type hint and boolean mask usage.

GiGiKoneti: Updated docstring to `np.ndarray or None, optional` to match the type hint and boolean mask usage.

GiGiKoneti: Updated docstring to `config_path : str or Path` to align with the type hint.

GiGiKoneti: Updated docstring to document `split : str or None` and noted that `split=None` is allowed for static data.

neural_lam/datastore/npyfilesmeps/store.py
GiGiKoneti: Added `@overload` signatures for `stack_grid_coords` and `unstack_grid_coords` in `BaseDatastore`. Type checkers now know `DataArray` in returns `DataArray`, so this runtime assertion has been removed.

neural_lam/datastore/plot_example.py
GiGiKoneti: Removed along with the other asserts now that `@overload` is added to `unstack_grid_coords`.

neural_lam/datastore/plot_example.py
GiGiKoneti: xarray's type stubs use a `_Wrapped` callable descriptor for `DataArrayPlotAccessor.__call__`, which causes `ty` to report `missing-argument` for `self` when calling `da.plot(...)`. Added an inline comment explaining this stub limitation.

neural_lam/models/step_predictors/graph/hi_lam.py
GiGiKoneti: Replaced the runtime `isinstance` asserts in the loop with `typing.cast(nn.ModuleList, ...)`. This provides static narrowing to `ty` without any runtime overhead inside `mesh_step`.

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
GiGiKoneti: Yes! `torch.cat` expects `dim=1`; `axis` was invalid in PyTorch and threw an error if reached.

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
GiGiKoneti: `torch.split` returns a `tuple`, but `hi_processor_step` in `BaseHiGraphModel` specifies `list[torch.Tensor]` in its return signature. The `list(...)` wrappers convert the tuple slices to lists to satisfy the base class contract.

neural_lam/models/module.py
GiGiKoneti: PyTorch Lightning's `all_gather` signature is stubbed as `Any` / `Union[Tensor, Dict, List]`. The `assert isinstance(gathered, torch.Tensor)` is required by static type checkers to narrow `gathered` before calling `.dim()` / `.flatten()`.

neural_lam/utils/logging.py
GiGiKoneti: Switched from private attribute `datastore._config` to the public property `datastore.config`.

GiGiKoneti:
Thanks @observingClouds for the detailed review! I've addressed all comments:

- Added  signatures for / in  and removed redundant runtime  checks.
- Replaced inner-loop  in  with .
- Fixed docstring types for , , and .
- Confirmed  ->  fix and explained  conversions matching base class interface signatures.

All 278 unit tests and  are passing green.

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
observingClouds: Thanks for checking and explaining

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
observingClouds: @GiGiKoneti you might have missed this and the next comment

GiGiKoneti: These `list(...)` calls convert the `tuple` returned by `torch.split` (and tuple slicing) into `list[torch.Tensor]`. This is required to satisfy the `hi_processor_step` interface contract defined in `BaseHiGraphModel`, which specifies mutable `list[torch.Tensor]` return types (needed because downstream methods like `HiLAM.mesh_step` perform element assignment `mesh_rep[i] = ...` which requires mutable lists).

neural_lam/models/step_predictors/graph/hi_lam_parallel.py
joeloskarsson: HiLAMParallel is in this limbo where it is not really used much (I believe), and (as clear from this) not properly tested nor maintained :/ But in my view there is no clear reason to drop it, so it is of course good if we can fix issues like this as they are spotted.

What confuses me here is how this relates to https://github.com/mllam/neural-lam/pull/680. We should have run into this there? And the test from there should catch this? Investigating.

GiGiKoneti: Correction on my earlier reply: `torch.cat` does silently accept `axis` as a numpy-compatible alias for `dim`, so this wasn't a runtime error. The test from #680 passes because `axis=1` works identically to `dim=1`. The change here is a style fix to use the canonical PyTorch keyword.

joeloskarsson: This bug was introduced with the refactor (https://github.com/mllam/neural-lam/commit/95cc8b0719f3c821509e4a0eb95bf21265a42fe8).

Ok apparently this just works, but seems undocumented 🙃
```
>>> a
tensor([[0, 1, 2]])
>>> b
tensor([[3, 4, 5]])
>>> torch.cat([a, b], dim=0)
tensor([[0, 1, 2],
        [3, 4, 5]])
>>> torch.cat([a, b], axis=0)
tensor([[0, 1, 2],
        [3, 4, 5]])
>>> torch.cat([a, b], dim=1)
tensor([[0, 1, 2, 3, 4, 5]])
>>> torch.cat([a, b], axis=1)
tensor([[0, 1, 2, 3, 4, 5]])
```

So I suppose not even a bugfix per say, just a fix to actually use proper (documented) pytorch. I'd just leave the fix in this PR though. It is good that it is done.

observingClouds: Nice, thanks guys for investigating my concerns.

neural_lam/models/step_predictors/graph/graph_efm.py
joeloskarsson: I can pitch in: This is correct, just a small change that adds a bit of clarity.

neural_lam/utils/logging.py
sadamov: No behaviour change for the two bundled datastores (their `config` property is just `return self._config`), but `_config` reached past the public `BaseDatastore.config` contract and would break any datastore that doesn't happen to store it under that name, e.g. `DummyDatastore` with `--logger wandb`. 
@GiGiKoneti Small follow-up: `tests/test_cli.py` still sets `datastore._config = {}` on the mock, could we switch those to `datastore.config = {}`?

GiGiKoneti: Updated all mock setups in `tests/test_cli.py` to use `datastore.config = {}`.

sadamov:
@observingClouds i just had a look at the comments (especially where you pinged me), all good from my side. If you accept we can merge this one

observingClouds (approved):
LGTM now! Thanks @GiGiKoneti for your contribution and patience 🤣

neural_lam/models/module.py
```diff
@@ -1001,7 +1002,8 @@ def on_test_epoch_end(self) -> None:
             torch.cat(self.spatial_loss_maps, dim=0)
         )
         if self.trainer.is_global_zero:
-            mean_spatial_loss = torch.mean(spatial_loss_tensor, dim=0)
+            # `nanmean` because boundary nodes are NaN-masked in `test_step`
```
observingClouds: Could we remove this boilerplate comment?

sadamov: @GiGiKoneti i am just merging in some other bugfixes first, it was me who introduced the comment above. @observingClouds will remove it before merge.

sadamov: removed in https://github.com/mllam/neural-lam/pull/673/commits/5156c3109b31e14916486725dd6c6caa5e0a83ba

sadamov:
@observingClouds maybe you want to have a quick look at https://github.com/mllam/neural-lam/pull/673/commits/6b18cf7a76847c296022d00f01da1578ec595def before I finally and actually merge :D
I merged in #688 which this branch here depends on and had a legacy-implementation of.

PR #620, opened 2026-04-22
Add comprehensive type hints to utils.py

GiGiKoneti:
## Describe your changes

Added type hints to all previously untyped functions and class methods in `neural_lam/utils.py`, including the `BufferList` class and utility functions (`load_graph`, `make_mlp`, `setup_training_logger`, `inverse_softplus`, `inverse_sigmoid`, etc.).

Uses lowercase generics (Python 3.10+) to match the existing style in this file.

No new dependencies required.

## Issue Link

closes #619

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [x] the code is readable
- [x] the code is well tested
- [x] the code is documented (including return types and parameters)
- [x] the code is easy to maintain

## Author checklist after completed review

- [x] I have added a line to the CHANGELOG describing this change, in a section
  reflecting type of change (add section where missing):
  - *added*: when you have added new functionality
  - *changed*: when default behaviour of the code has been changed
  - *fixes*: when your contribution fixes a bug
  - *maintenance*: when your contribution is relates to repo maintenance, e.g. CI/CD or documentation

## Checklist for assignee

- [x] PR is up to date with the base branch
- [x] the tests pass
[...]

PR #677, opened 2026-06-20
Add PersistencePredictor baseline model

### LINKED ISSUE #676: Add baseline persistence model forecaster
To support evaluating weather models against a persistence baseline, we should implement a baseline persistence model. This model will simply output the initial state repeated for all future steps, allowing standard evaluation metrics to be computed for a persistence baseline.

GiGiKoneti:
## Describe your changes

- Add `PersistencePredictor`, a baseline `StepPredictor` model that returns the previous state unchanged at each autoregressive step.
- Register `PersistencePredictor` as `persistence` in the `MODELS` registry inside `neural_lam/models/__init__.py`.
- Add validation logic in `neural_lam/train_model.py` to prevent training the persistence model, raising a ValueError if training is attempted.
- Add comprehensive unit tests covering the forward pass, kwargs absorption, forecaster unrolling, predicts_std behavior, and training error prevention.

No changes to any existing prediction pathways.

None.

## Issue Link

closes #676

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `PersistencePredictor` baseline model (`--model persistence`) that returns the previous state unchanged, enabling standard evaluation of a persistence baseline through the existing pipeline [\#676](https://github.com/mllam/neural-lam/issues/676) @GiGiKoneti
```
sadamov: Our convention is to link the PR number, not the issue (the issue is already covered by `closes #676` in the body).

```suggestion
- Add `PersistencePredictor` baseline model (`--model persistence`) that returns the previous state unchanged, enabling standard evaluation of a persistence baseline through the existing pipeline [\#677](https://github.com/mllam/neural-lam/pull/677) @GiGiKoneti
```

CHANGELOG.md
```diff
@@ -112,12 +114,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Explicitly cleanup the temporary directory in `DummyDatastore` by adding a `__del__` method, preventing `ResourceWarning` during garbage collection in tests [\#487](https://github.com/mllam/neural-lam/pull/487) @sohampatil01-svg
 
-- Add comprehensive type hints to all functions and class methods in `utils.py` [\#620](https://github.com/mllam/neural-lam/pull/620) @GiGiKoneti
 
 - Add probabilistic objective regression coverage for weighted losses and `pred_std` broadcasting semantics [\#504](https://github.com/mllam/neural-lam/pull/504) @kshirajahere
 
 - Add comprehensive type hints to `neural_lam/create_graph.py` [\#618](https://github.com/mllam/neural-lam/pull/618) @GiGiKoneti
 
+- Add comprehensive type hints to all functions and class methods in `utils.py` [\#620](https://github.com/mllam/neural-lam/pull/620) @GiGiKoneti
```
sadamov: This relocates the unrelated #620 entry, looks like a merge/rebase artifact. Could we rebase on `main` and drop the move so the diff only adds the persistence line?

tests/test_persistence.py
```diff
@@ -0,0 +1,107 @@
...
+    # Interior nodes should equal init_states[:, 1] (persistence),
+    # boundary nodes should equal boundary_states
+    interior_mask = forecaster.interior_mask.squeeze(0).squeeze(-1).bool()
+    for t in range(pred_steps):
+        interior_pred = prediction[:, t, interior_mask, :]
+        # After first step interior is init_states[:, 1], then persists
+        interior_boundary = boundary_states[:, t, interior_mask, :]
+        # Verify interior nodes are NOT equal to boundary_states
+        # (they should be the persisted initial state)
+        assert interior_pred.shape == interior_boundary.shape
```
sadamov: This loop only asserts `shape ==`, which is always true, so it does not actually check persistence. Interior nodes stay at `init_states[:, 1]` and boundary nodes equal `boundary_states[:, t]` for every step (I verified locally), so we can assert the real behaviour:

```suggestion
    # Persistence keeps every interior node at init_states[:, 1] for all
    # steps, while boundary nodes are overwritten with boundary_states.
    interior_mask = forecaster.interior_mask.squeeze(0).squeeze(-1).bool()
    boundary_mask = forecaster.boundary_mask.squeeze(0).squeeze(-1).bool()
    for t in range(pred_steps):
        assert torch.equal(
            prediction[:, t, interior_mask, :],
            init_states[:, 1, interior_mask, :],
        )
        assert torch.equal(
            prediction[:, t, boundary_mask, :],
            boundary_states[:, t, boundary_mask, :],
        )
```

sadamov (commented):
Thanks @GiGiKoneti, this is clean and well-scoped. I ran it locally: the 5 tests pass, pre-commit is clean, and `--eval test --model persistence` runs end-to-end with the loss growing by lead time...

A few small asks inline, none of them block the logic.
- the training guard in `train_model.py` compares against the literal `"persistence"`. Fine for a single baseline, though a `trainable = False` class attribute would be more robust if more analytical baselines arrive later.
- `--output_std` is silently ignored with persistence (documented, harmless). A `logger.warning` would make the no-op explicit.
- README is still unchecked in the PR checklist. A one-liner noting `--model persistence` needs `--eval val/test` would help discoverability.

GiGiKoneti:
@sadamov 
- Replaced the hardcoded training guard with a `trainable` class attribute on `StepPredictor`.
- Fixed the cascading `test_eval_without_load_warning` failure caused by the new `MODELS` dictionary lookup in `train_model.py`.
- Addressed all inline suggestions for the changelog, test assertions, std warning, and README.

neural_lam/train_model.py
```diff
@@ -457,6 +462,13 @@ def main(input_args=None):
         mesh_up_gnn_type=args.mesh_up_gnn_type,
         mesh_down_gnn_type=args.mesh_down_gnn_type,
     )
+
+    if args.output_std and not predictor.predicts_std:
```
joeloskarsson: Is this not a bit confusing logic for this check? `predictor.predicts_std` generally tells us that a predictor is configured to predict std-dev, not that it is not capable of predicting std. In all other cases `predictor.predicts_std` is just set from `args.output_std`.
In this case I think it might be better to do the check in the PersistencePredictor constructor.

neural_lam/train_model.py
```diff
@@ -355,6 +355,11 @@ def main(input_args=None):
         ),
     )
     args = parser.parse_args(input_args)
+    if not args.eval and not MODELS[args.model].trainable:
```
joeloskarsson: An idea would be to just derive the `trainable` property from checking if the model has trainable parameters? That way we avoids an additional property that is just a duplicated proxy for this. 
(This would only cause issues if we have a StepPredictor that has trainable parameters, but they are not used during the forward pass and can not be trained in practice. This would not make sense and we should avoid such a case)

joeloskarsson:
Looking very good, I added some ideas above that could be small improvements :)

neural_lam/train_model.py
GiGiKoneti: Moved the standard deviation warning directly into `PersistencePredictor.__init__.`

neural_lam/train_model.py
GiGiKoneti: Replaced the class-level `trainable` attribute with a dynamic property on the `StepPredictor` base class that checks `requires_grad` on instantiation.

joeloskarsson (approved):
Nice! LGTM

tests/test_train_model_warnings.py
```diff
@@ -34,13 +35,21 @@ def test_eval_without_load_warning(eval_val, load_val, expect_warning):
...
+                        "neural_lam.train_model.MODELS",
+                        {"graph_lam": mock_model_class},
+                    ):
+                        with pytest.raises(SystemExit):
+                            main()
+                        if expect_warning:
+                            mock_warning.assert_called_once()
+                            assert "--load" in mock_warning.call_args[0][0]
+                        else:
+                            mock_warning.assert_not_called()
```
sadamov: These patches are vestigial since the guard moved past the stubbed `load_config_and_datastore` (which exits before reaching them), so I propose to drop them:

```suggestion
                with pytest.raises(SystemExit):
                    main()
                if expect_warning:
                    mock_warning.assert_called_once()
                    assert "--load" in mock_warning.call_args[0][0]
                else:
                    mock_warning.assert_not_called()
```

tests/test_persistence.py
```diff
@@ -0,0 +1,144 @@
...
+# First-party
+from neural_lam.models import ARForecaster
+from neural_lam.models.step_predictors.persistence import PersistencePredictor
+from tests.conftest import init_datastore_example
+
+
+def test_persistence_predictor_returns_prev_state():
+    """PersistencePredictor.forward returns prev_state unchanged."""
+    datastore = init_datastore_example("mdp")
+    predictor = PersistencePredictor(datastore=datastore)
```
sadamov, line 16: The real `any(p.requires_grad ...)` trigger is never exercised because `test_persistence_training_error` mocks `trainable`, so one line here closes the gap:

```suggestion
    predictor = PersistencePredictor(datastore=datastore)
    assert predictor.trainable is False
```

sadamov (commented):
Clean and correct, already approved. A few small polish asks inline, none block the logic.

GiGiKoneti:
Done!
- Removed the vestigial patches from `test_eval_without_load_warning` in `tests/test_train_model_warnings.py` as suggested.
- Added the missing `predictor.trainable is False` assertion to `test_persistence_predictor_returns_prev_state` in `tests/test_persistence.py` to exercise the dynamic `trainable` property.

sadamov (approved):

PR #706, opened 2026-07-15
Fix legacy graph indexing offset mismatch when boundary grid nodes are unconnected

### LINKED ISSUE #699: Bug: Offset Mismatch in zero_index_m2g / zero_index_g2m when last grid node is unconnected
### **Bug Description**
When `mesh_first` is `False` (grid nodes are indexed before mesh nodes), `zero_index_m2g` and `zero_index_g2m` in `neural_lam/utils.py` dynamically calculate the offset for mesh nodes using the maximum connected grid node index:
*   `num_interior_nodes = m2g_edge_index[1].max() + 1` in `zero_index_m2g`
*   `num_grid_nodes = g2m_edge_index[0].max() + 1` in `zero_index_g2m`

This assumes the highest-indexed grid node is always connected to the mesh. If the highest-indexed grid nodes are unconnected (e.g., due to boundary masking or spatial selection criteria), the computed offset is smaller than the true grid size. This corrupts the zero-indexing of all subsequent mesh nodes, causing index alignment errors during encoding and decoding.

---

### **Minimal Reproducible Example**
The following code reproduces the bug by creating a mock edge index where the highest active grid node connected in `m2g` is `90` (while the true grid size is `100`):

```python
import torch
from neural_lam.utils import zero_index_m2g

# True configuration: 100 grid nodes (0..99), 10 mesh nodes (100..109)
true_num_grid_nodes = 100

# Edge index where the highest active grid node connected in m2g is 90
m2g_edge_index = torch.tensor([
    [100, 101, 102, 109],  # mesh node indices
    [  0,  10,  50,  90]   # grid node indices
])

# Computes offset as 90 + 1 = 91 instead of 100
zeroed_m2g = zero_index_m2g(m2g_edge_index, mesh_static_features=[], mesh_first=False)

print(f"Original mesh indices: {m2g_edge_index[0].tolist()}")
print(f"Expected zero-indexed: {[x - true_num_grid_nodes for x in m2g_edge_index[0].tolist()]}")
print(f"Actual zero-indexed:   {zeroed_m2g[0].tolist()}")
# Actual zero-indexed:   [9, 10, 11, 18] -> Does not start at 0
```

---

### **Impact**
This causes subsequent mesh node index accesses to point to the wrong locations in the mesh representation array (or crash with an `IndexError` if the corrupted index exceeds the array bounds), leading to silent spatial coordinate misalignment or runtime crashes.

---

### **Proposed Fix**
Pass the true grid node count (e.g. `num_grid_nodes` from the datastore) directly into `load_graph` and use it to offset the indices rather than inferring it from the edge indices.

GiGiKoneti:
## Describe your changes

This PR fixes the offset mismatch in `zero_index_m2g` and `zero_index_g2m` when `mesh_first` is `False`. 
Specifically, instead of dynamically computing the number of grid nodes from the maximum active node index in the edge index (which is incorrect if some boundary grid nodes are unconnected to the GNN mesh), it computes the total number of nodes at the bottom level and subtracts the number of mesh nodes at that level:
1. `max_idx` is calculated over `g2m`, `m2g`, and the bottom level `m2m` edge indices.
2. `num_grid_nodes` is calculated as `(max_idx + 1) - num_mesh_nodes`.
3. Passes the correctly computed `num_grid_nodes` down to the zero-indexing helpers.

## Issue Link

closes #699

## Type of change

- [x] I have updated the [README](README.MD) to cover introduced code changes (no README changes needed for this bug fix)

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

- [x] I have added a line to the CHANGELOG describing this change, in a section reflecting type of change:
  - *fixes*: `- Fix offset mismatch in legacy graph zero-indexing...` under `## [unreleased]`


neural_lam/utils/graph.py
```diff
@@ -318,11 +329,33 @@ def load_graph_spec_version() -> str:
...
+        num_grid_nodes = None
+        if not mesh_first:
+            # Under legacy layout, mesh nodes are at indices:
+            # num_grid_nodes to num_grid_nodes + num_mesh_nodes - 1.
+            # We determine total number of nodes at the bottom level
+            # to find num_grid_nodes.
+            max_idx = max(
+                g2m_edge_index.max().item(),
+                m2g_edge_index.max().item(),
+                max(ei.max().item() for ei in m2m_edge_index),
```
Sir-Sloth-The-Lazy, line 344: `m2m_edge_index` gets reassigned to the locally zero-indexed version (each level independently rebased to start at 0 via `zero_index_edge_index`) several lines before the new `max_idx` block runs. So by the time it's used, it no longer lives in the same raw/combined grid+mesh index space as the still-raw `g2m_edge_index/m2g_edge_index`, mixing them in one `max()` is comparing apples to oranges. In practice this doesn't blow up on typical graphs, because grid counts vastly exceed mesh counts, so the raw g2m/m2g values dominate the max and the (small, rebased) m2m term is inert. But it means the m2m term contributes nothing to the robustness the fix is meant to provide, if the mesh's topmost-indexed node were ever the one lacking a g2m/m2g edge (the same failure mode as #699, just mirrored onto the mesh side), this fix wouldn't catch it either, silently.

There's also a smaller mismatch: the comment says "we determine total number of nodes at the bottom level," but the code does `max(ei.max().item() for ei in m2m_edge_index)` over the entire list (all mesh levels), not just `m2m_edge_index[0]`. For non-hierarchical graphs this is moot, but it's inconsistent with the stated intent and the` mesh_static_features[0] ` reference right below it.

Suggested fix to flag: compute `max_idx` from the raw `m2m_edge_index` (before the `zero_index_edge_index` reassignment) or move the whole num_grid_nodes computation up before that reassignment and use `m2m_edge_index[0]` explicitly to match the "bottom level" comment.

Sir-Sloth-The-Lazy (changes requested):
Just a small problem here, also great eye for finding this issue !  Just a ping no tests are added till now however the descirption suggests there should be one. I would suggest  the exact scenario from #699 (legacy format, grid-first layout, unconnected boundary grid node) has zero coverage before or after this PR  you can add a regression test here with a synthetic legacy  graph that has a diconneted top-indexed grid node, mirroring the reproduction in the issue. 

Also **the new parameter inserted before the existing one changes the public call signature**. Both `zero_index_m2g` and `zero_index_g2m` are exported from `neural_lam/utils/__init__.py `(public API). The PR adds `num_grid_nodes: int | None = None `before the existing `restore: bool = False` parameter. All current in-repo callers use keywords, so nothing breaks internally, but any external caller invoking these positionally (e.g. `zero_index_m2g(idx, features, mesh_first, True)`) would now silently bind `Tru`e to `num_grid_nodes` instead of `restore`, `restore` would default to `False` while `num_grid_nodes `becomes truthy, corrupting the result without raising. Safer to append new optional params after `restore`, or make them keyword-only.

GiGiKoneti:
Thanks for the thorough review and catch @Sir-Sloth-The-Lazy!

1. **API Signature:** Updated `zero_index_m2g` and `zero_index_g2m` to place `restore: bool = False` before `num_grid_nodes: int | None = None`, preserving positional parameter order for external callers.
2. **Regression Test:** Added `test_legacy_graph_unconnected_boundary_grid_node_zero_indexing` in `tests/test_graph_creation.py` to cover legacy graphs (`mesh_first=False`) with top-indexed unconnected boundary grid nodes.

All 250 tests and pre-commit checks are passing cleanly!

PR #710, opened 2026-07-17
Fix logger config for datastores without _config

### LINKED ISSUE #709: `setup_training_logger` assumes every datastore exposes a private `_config`
`neural_lam/utils/logging.py:120` and `:143` build the logger config with:

```python
config=dict(training=vars(args), datastore=datastore._config)
```

`_config` is an MDP-specific private attribute, not part of the datastore interface, so any datastore that doesn't happen to carry one raises `AttributeError` the moment training starts:

```
AttributeError: 'DummyDatastore' object has no attribute '_config'. Did you mean: 'config'?
```

We already work around this in the test suite rather than fixing it: `tests/test_cli.py:84`, `:112` and `:145` each inject `datastore._config = {}` onto a mock before calling the training entry point.

The practical effect is that an in-memory datastore (e.g. `tests/dummy_datastore.py::DummyDatastore`) cannot be driven through the real training entry point without patching it first.

I suggest the minimal fix:

```python
config=dict(training=vars(args), datastore=getattr(datastore, "_config", None))
```

Logging `None` for a datastore with no raw config file is fine. the cleaner option might be smth like a public accessor on the datastore interface, not sure...

Mayank-Bajoria:
## Summary

This PR updates `setup_training_logger` to avoid assuming that every datastore exposes a private `_config` attribute.

It replaces direct access to:

```python
datastore._config
```

with:

```python
getattr(datastore, "_config", None)
```

This prevents an `AttributeError` for datastores that do not implement `_config` while preserving the existing behavior for datastores that do.

## Issue

Closes #709

### DIFF
--- a/neural_lam/utils/logging.py
+++ b/neural_lam/utils/logging.py
@@ -117,7 +117,7 @@ def setup_training_logger(
         return pl.loggers.WandbLogger(
             project=args.logger_project,
             name=None if args.wandb_id else run_name,
-            config=dict(training=vars(args), datastore=datastore._config),
+            config=dict(training=vars(args), datastore=getattr(datastore, "_config", None)),
             resume=wandb_resume,
             id=args.wandb_id,
             save_dir=run_dir,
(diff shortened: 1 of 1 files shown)


sadamov:
#723 and #673 fixed this issue at the root instead of the symptom. #709 can therefore be considered resolved. thanks for your contribution @Mayank-Bajoria

Issue #385, opened 2026-03-11
[Design discussion] Supporting `pyg.HeteroData` use internally in `neural-lam`


leifdenby:
The aim of this issue to continue the discussion on whether and how to support use of `pyg.HeteroData` objects to represent the graph when working with `neural-lam` (rather than the [dictionary of edge-indexes and node/edge features currently used](https://github.com/mllam/neural-lam/blob/main/neural_lam/utils.py#L188))

To start of the discussion here are some things I have been wondering about:

1. My view is that we will eventually want to use `pyg.HeteroData` objects to represent the graph internally in `neural-lam`. This is what I did in https://github.com/leifdenby/weatherduck which was a prototype implementation trying that approach out (and also trying out fiddle-based config, which is a different conversation). It is also what @matschreiner does in https://github.com/matschreiner/equicast/tree/main?tab=readme-ov-file#data-handler. I think it simplifies the code a lot (only a single object is passed around representing the entire graph) and enables us to eventually also put the model inputs (state/forcing etc) as node features *on* this graph object, rather than as separate tensors being passed around (which I think is a nice design for implementing new architectures in `neural-lam`)
    - Is there an interest in moving towards this?
    - Have some people already started this work?
    - @SimonKamuk you and I have discussed this, have you looked at refactoring `neural-lam` in this way?
    - If we agree this should be done, maybe we could plan a first step in this direction to be loading the tensor-based format into a `pyg.HeteroData`object that is then passed to the `torch.nn.Module`-derived classes that implement the forecasting model (currently ARModel). What do people think of that?
2. Does it make sense trying to export from `weather-model-graphs` to `pyg.HeteroData` objects and pickling these to disk?
    - If so, what would the drawbacks of these be? To me a clear drawback is that the format (pickled python objects) isn't a self-describing format, unlike what a `xr.DataTree` `netCDF`/`zarr` dataset could be. And it relies on `torch-geometric` which seems a bit overkill for something that just describes a static graph with edge and node features.
    - If we instead opt for writing `xr.DataTrees` to disk to represent the graph what would the pros/cons there be? Is the `xr.DataTree` functionality I have already drafted (https://github.com/mllam/weather-model-graphs/pull/47) fit-for-purpose or is there something wrong/missing?

This issue is meant as a continuation that started on #339 on bridging the gap between `weather-model-graphs` and `neural-lam`.

Joltsy10:
[...]

PR #711, opened 2026-07-22
Load flat graphs into pyg.HeteroData and wire into BaseGraphModel

prajwal-tech07:
First step of the HeteroData migration (#385), scoped to **flat** (non-hierarchical) graphs.

- Adds `neural_lam.utils.heterodata` with `graph_dict_to_heterodata` and its inverse `heterodata_to_graph_dict`. Grid nodes are the `grid` node type and mesh nodes the `mesh` node type, matching the terminology used throughout neural-lam (`grid_static_features`, `mesh_static_features`, `g2m`/`m2m`/`m2g`); `g2m`/`m2m`/`m2g` become typed edges. The grid node count is taken from the datastore rather than inferred from edge indices, and the edge-feature width is read dynamically (3 for 2D, 4 for 3D). The node/edge-type names are module-level constants; the reference implementations linked from the issue (`leifdenby/weatherduck`, `matschreiner/equicast`) use `data`/`hidden` - happy to switch if preferred.
- Wires into `BaseGraphModel` behind a `use_heterodata` flag (default `False`): when enabled the graph is represented as a `HeteroData` object and the model's graph tensors are taken from it. The extracted tensors are identical to the dict ones, so the model builds and trains identically either way, and existing checkpoints remain compatible.
- Threads the flag through `GraphLAM`.

Hierarchical (HiLAM) support is deliberately left as a follow-up - there is an open question on #385 about how to represent mesh levels as node/edge types.

## Issue Link

Part of #385 (first implementation step; the design discussion and hierarchical follow-up remain open).

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/neural_lam/models/step_predictors/graph/base.py
+++ b/neural_lam/models/step_predictors/graph/base.py
@@ -116,11 +116,12 @@ class BaseGraphModel(StepPredictor):
             mesh_node_features_scaling=grid_xy_max_span,
         )
 
-        # Optionally represent the graph as a pyg.HeteroData object and take
-        # the model's graph tensors from it, so that everywhere the graph is
-        # used the data now originates from the HeteroData datastructure
-        # (issue #385). The extracted tensors are identical to the dict ones,
-        # so the model builds and trains identically either way.
+        # Optionally represent the graph as a pyg.HeteroData object (issue
+        # #385). The loaded tensors are placed on the typed node/edge stores
+        # of that object, and every graph tensor the model uses below is then
+        # read back out of it, so the HeteroData is the source of the graph
+        # data. The tensors themselves are unchanged, so the model builds and
+        # trains identically either way.
         self.use_heterodata = use_heterodata
         if use_heterodata:
             if self.hierarchical:
@@ -132,7 +133,7 @@ class BaseGraphModel(StepPredictor):
                 graph_ldict,
                 num_grid_nodes=self.num_grid_nodes,
             )
-            graph_ldict = utils.heterodata_to_graph_dict(self.graph)
+            graph_ldict = utils.graph_tensors_from_heterodata(self.graph)
 
         for name, attr_value in graph_ldict.items():
             # Make BufferLists module members and register tensors as buffers
--- a/neural_lam/models/step_predictors/graph/base.py
+++ b/neural_lam/models/step_predictors/graph/base.py
@@ -33,6 +33,7 @@ class BaseGraphModel(StepPredictor):
         output_clamping_upper: dict[str, float] | None = None,
         g2m_gnn_type: str = "InteractionNet",
         m2g_gnn_type: str = "InteractionNet",
+        use_heterodata: bool = False,
     ) -> None:
         """
         Initialize the BaseGraphModel.
@@ -62,6 +63,10 @@ class BaseGraphModel(StepPredictor):
             Lower clamping limits for state variables.
         output_clamping_upper : dict, optional
             Upper clamping limits for state variables.
+        use_heterodata : bool, default False
+            If True, represent the loaded graph as a ``pyg.HeteroData``
+            object and take the model's graph tensors from it. Only supported
+            for flat (non-hierarchical) graphs.
         """
         super().__init__(
             datastore=datastore,
@@ -110,6 +115,24 @@ class BaseGraphModel(StepPredictor):
             graph_dir_path=graph_dir_path,
             mesh_node_features_scaling=grid_xy_max_span,
         )
+
+        # Optionally represent the graph as a pyg.HeteroData object and take
+        # the model's graph tensors from it, so that everywhere the graph is
+        # used the data now originates from the HeteroData datastructure
+        # (issue #385). The extracted tensors are identical to the dict ones,
+        # so the model builds and trains identically either way.
+        self.use_heterodata = use_heterodata
+        if use_heterodata:
+            if self.hierarchical:
+                raise NotImplementedError(
+                    "use_heterodata is currently only supported for flat "
+                    "(non-hierarchical) graphs."
+                )
+            self.graph = utils.graph_dict_to_heterodata(
+                graph_ldict,
+                num_grid_nodes=self.num_grid_nodes,
+            )
+            graph_ldict = utils.heterodata_to_graph_dict(self.graph)
 
         for name, attr_value in graph_ldict.items():
             # Make BufferLists module members and register tensors as buffers
--- a/neural_lam/models/step_predictors/graph/base.py
+++ b/neural_lam/models/step_predictors/graph/base.py
@@ -4,6 +4,7 @@
 
 # Third-party
 import torch
+from torch_geometric.data import HeteroData
 
 # Local
 from .... import utils
@@ -176,6 +177,36 @@ class BaseGraphModel(StepPredictor):
 
         # Compute indices and define clamping functions
         self.prepare_clamping_params(datastore)
+
+    @property
+    def graph(self) -> HeteroData:
+        """The model's graph as a ``pyg.HeteroData`` object.
+
+        Built on access from the graph tensors currently registered on the
+        model, so it always agrees with their device and dtype. It is not
+        stored on the model, because a ``HeteroData`` attribute would not be
+        moved by ``.to(...)`` and would go stale.
+
+        Returns
+        -------
+        torch_geometric.data.HeteroData
+            Typed graph referencing the model's current graph tensors.
+
+        Raises
+        ------
+        AttributeError
+            If the model was not constructed with ``use_heterodata=True``.
+        """
+        if not self.use_heterodata:
+            raise AttributeError(
+                "graph is only available on models constructed with "
+                "use_heterodata=True"
+            )
+        return utils.heterodata_from_module(
+            self,
+            num_grid_nodes=self.num_grid_nodes,
+            hierarchical=self.hierarchical,
+        )
 
     def get_num_mesh(self) -> tuple[int, int]:
         """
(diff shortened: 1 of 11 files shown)


prajwal-tech07:
@leifdenby @joeloskarsson could I get a review on this when you have a chance? It's been rebased onto main after #648 - that PR moved graph loading into `utils.load_and_register_graph`, so the `use_heterodata` wiring now lives there. #713 (hierarchical) stacks on top of this one.

**Correction:** I originally wrote here that this means the wiring "applies to any module loading a graph through that helper, including the new Graph-EFM predictors". That was wrong - the helper takes the flag, but `BaseGraphEFM` never passes it, so it isn't reachable there. Thanks @Sir-Sloth-The-Lazy for catching it; see the review thread below for what has been fixed.

neural_lam/utils/graph.py
```diff
@@ -447,17 +456,39 @@ def load_and_register_graph(
...
         mesh_node_features_scaling=mesh_node_features_scaling,
     )
+
+    if use_heterodata:
+        if hierarchical:
+            raise NotImplementedError(
+                "use_heterodata is currently only supported for flat "
+                "(non-hierarchical) graphs."
+            )
+        module.graph = graph_dict_to_heterodata(
```
Sir-Sloth-The-Lazy: `HeteroData` is neither `Tensor`, `Parameter`, nor `Module`, so this `setattr` lands in `module.__dict__` and `nn.Module._apply` skips it. After `.cuda()`, the registered buffers move but `module.graph` keeps the original CPU tensors, a stale second copy of `m2m_features` etc. held on host memory. Harmless today since nothing reads `module.graph` back, but it'll bite the moment a later PR actually uses it in `forward`.

neural_lam/utils/graph.py
```diff
@@ -436,6 +438,13 @@ def load_and_register_graph(
     buffer and each non-tensor (e.g. ``BufferList``) as a plain attribute on
     ``module``.
 
+    With ``use_heterodata`` the loaded tensors are first placed on the typed
+    node/edge stores of a ``pyg.HeteroData`` object (issue #385), which is
+    stored as ``module.graph``, and the tensors registered on ``module`` are
+    then read back out of that object, making it the source of the module's
+    graph data. The tensors themselves are unchanged, so the module behaves,
+    and therefore trains, identically either way.
+
```
Sir-Sloth-The-Lazy: I think this is an overstatement, if you would like to clarify this ? I am sorry, I did not understand this

Sir-Sloth-The-Lazy:
**`use_heterodata` is unreachable outside a direct `GraphLAM(...)` call.**

- `BaseHiGraphModel.__init__`  doesn't accept it and doesn't forward it to `super().__init__()`, so `HiLAM`/`HiLAMParallel` can never set it (moot anyway since they're hierarchical, but the flag isn't even reachable to raise the informative error).
- `BaseGraphEFM.__init__` calls `load_and_register_graph` without `use_heterodata` at all.
- `train_model.py` defines no `--use_heterodata` CLI flag, so it can't be set from the training entrypoint either.

The PR thread says *"the use_heterodata wiring now lives [in load_and_register_graph] and applies to any module loading a graph through that helper, including the new Graph-EFM predictors"*  that's true of the helper's signature, but no caller actually passes it through, so in practice the only way to exercise this flag today is constructing `GraphLAM` directly (which is what the test does).

neural_lam/utils/graph.py
prajwal-tech07: Fair, that was an overstatement. It read as if the `HeteroData` were the module's durable source of graph data, when it's really a pass-through at load time. Reworded: the tensors are routed through the `HeteroData` object and read back out of it, and the docstring now also says explicitly that the object isn't stored on the module and why.

prajwal-tech07:
You're right on all three, and the claim I made in the thread above was wrong, not just imprecise. Thanks for checking it rather than taking it at face value.

Fixed the parts that are in scope here: `use_heterodata` is now threaded through `BaseHiGraphModel`, `HiLAM` and `HiLAMParallel` so the informative error is actually reachable, and there's a --use_heterodata CLI flag in train_model.py passed at both construction sites, so it can be exercised from the training entrypoint rather than only by constructing `GraphLAM` directly. The checkpoint-reload path uses a getattr default so older checkpoints still load.

I left `BaseGraphEFM` alone deliberately: it isn't in the `MODELS` registry yet, so the flag couldn't be reached from the CLI there in any case, and it seemed better not to churn that code while it's still landing. Happy to add it if you'd prefer. Either way the "including the Graph-EFM predictors" line was wrong; I have corrected that comment.

neural_lam/utils/graph.py
prajwal-tech07: Good catch, and thanks for the precise diagnosis. You're right and I reproduced it: after `.to(torch.float64)` the registered buffer became float64 while `module.graph` stayed float32, so `.cuda()` would leave a stale copy stranded on the host exactly as you describe.

Fixed by not storing the object at all. There's now `utils.heterodata_from_module`, and `BaseGraphModel.graph` is a property that builds the HeteroData on access from the tensors currently registered on the module. It can't go stale, it holds no second copy (the stores reference the module's own tensors), and it stays correct for the case you were worried about, a later PR reading it in `forward`. Added a test that moves the model and asserts the graph follows.

neural_lam/models/step_predictors/graph/base.py
```diff
@@ -33,6 +34,7 @@ def __init__(
         output_clamping_upper: dict[str, float] | None = None,
         g2m_gnn_type: str = "InteractionNet",
         m2g_gnn_type: str = "InteractionNet",
+        use_heterodata: bool = False,
```
leifdenby, line 37: I think we can just make the assumption that we will *only* use herodata going forward instead. I think that will be simpler

neural_lam/models/step_predictors/graph/base.py
```diff
@@ -105,11 +111,16 @@ def __init__(
             grid_xy_extent[1] - grid_xy_extent[0],
             grid_xy_extent[3] - grid_xy_extent[2],
         )
+        # With use_heterodata the loaded graph is represented as a
+        # pyg.HeteroData object (issue #385) that the model's graph tensors
+        # are then read out of; see load_and_register_graph.
+        self.use_heterodata = use_heterodata
```
leifdenby, line 117: Rather than pass in `self` below (and then modify self) I think we should introduce a torch module that will store the required graph information and attach that (i.e. `self.graph_store = utils.load_graph(datastore, ...)`. Modifying `self` here makes the side-effect of loading the graph hidden

neural_lam/models/step_predictors/graph/base.py
```diff
@@ -167,6 +178,36 @@ def __init__(
...
+        -------
+        torch_geometric.data.HeteroData
+            Typed graph referencing the model's current graph tensors.
+
+        Raises
+        ------
+        AttributeError
+            If the model was not constructed with ``use_heterodata=True``.
+        """
+        if not self.use_heterodata:
```
leifdenby, line 200: remove to just assume that we *only* use herodata

neural_lam/utils/heterodata.py
```diff
@@ -0,0 +1,216 @@
...
+    "g2m_features",
+    "m2g_features",
+    "m2m_features",
+    "mesh_up_features",
+    "mesh_down_features",
+    "mesh_static_features",
+)
+
+
+def graph_dict_to_heterodata(
```
leifdenby, line 70: I don't think we should have the "graph dict" at all in `neural-lam` going forward. Instead we should load directly into `HeteroData` datastructure and use that

leifdenby (changes requested):
This in the right direction, but I think we should be more bold:

1. I would assume that going forward we *only* use `HeteroData` for storing the graph internally in `neural-lam`, and so
2. I would remove the notion of a "graph dict" completely from the `neural-lam` code. We should be able to contain everything we need within a `HeteroData` data structure, this means
3. Rewriting the functions that take part of the graph's properties to use the heterodata data structure. For example edges in `BaseGraphModel`) accessing `self.g2m_features` this should be accessed via an object that holds everything related to the graph, e.g. access `self.graph.g2m_features`
4. Instead of passing in `self` (the `BaseGraphModel` instance) and assigning attributes (and thus modifying `self`) with the "load graph" call (currently called `load_and_register_graph()` https://github.com/mllam/neural-lam/blob/main/neural_lam/utils/graph.py#L464, we create a graph storage module (if we need it). Modifying the arguments with the graph loading function feels hacky to me and hides the side effect. Instead we could have all these `BufferList` members stored in a separate object (for example a `self.graph_store` or simply `self.graph` object) and get the graph properties from there

prajwal-tech07:
Sounds good, I'll go this way. One object owns the graph, no dict, no flag.

On point 4, I'll make the graph store an `nn.Module` that holds the buffers, rather than a plain `HeteroData` attribute. Reason is that a `HeteroData` object isn't an `nn.Module`, so it doesn't get moved when you call `.to()` or `.cuda()` on the model, and the graph would end up still pointing at tensors on the old device. That's the bug @Sir-Sloth-The-Lazy caught earlier in this PR. Making the store a module keeps the `BufferList` members moving with the model, and the properties still come off one object like you describe, so `self.graph.g2m_features`, with a typed HeteroData view available on it as well.

One thing I'd like your call . Removing the dict path also removes the test that checks training is identical with the old and new representation, since there's nothing left to compare against. That's the check you asked for originally and it passes today. Two options:

1. Merge this PR as it stands, with that test in it, and do the removal in a follow-up PR. The proof stays in the repo.
2. I run the comparison one last time, paste the result here as a record, and then delete the dict path and the test together in this PR.

I'm happy with either. I'll start on the rest either way since it doesn't change the bulk of the work, just let me know which you'd prefer for that last bit
Scope wise I'll keep it all in #711 since that's where you reviewed it. #713 is stacked on this one, so I'll rework it once this settles.

PR #568, opened 2026-04-01
Mask boundary nodes in spatial loss maps during test_step

RajdeepKushwaha5:
## Describe your changes

The `spatial_loss` call in `test_step` did not pass `mask=self.interior_mask_bool`, unlike every other `self.loss()` call in `training_step`, `validation_step`, and `test_step` itself. This caused boundary grid nodes to be included in the spatial loss maps and the saved `mean_spatial_loss.pt`, inconsistent with the loss the model optimises.

Set boundary node values to NaN after computing the full-grid loss (to keep the `num_grid_nodes` shape needed by `plot_spatial_error`) and switch to `torch.nanmean` when averaging over samples in `on_test_epoch_end`.

## Issue Link

closes #569

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form
- [ ] I have requested a reviewer and an assignee

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented (including return types and parameters)
- [ ] the code is easy to maintain

## Author checklist after completed review

- [x] I have added a line to the CHANGELOG describing this change, in a section reflecting type of change

## Checklist for assignee

- [ ] PR is up to date with the base branch

PR #720, opened 2026-08-15
Exclude boundary nodes from test_step spatial loss maps

### LINKED ISSUE #569: [Bug] spatial_loss_maps in test_step computed without interior_mask, inconsistent with all other loss calls
**Description**

In `ar_model.py`, `test_step` computes `spatial_loss_maps` at L452-454 without passing `mask=self.interior_mask_bool`:

```python
spatial_loss = self.loss(
    prediction, target, pred_std, average_grid=False
)
```

Every other call to `self.loss` in the module passes `mask=self.interior_mask_bool`:

- `training_step` (L311)
- `validation_step` (L354)
- `test_step` main loss (L408)

The per-variable metrics (`mse`, `mae`) computed just above (L435-441) also pass the mask. This means the spatial loss maps include boundary grid nodes that are excluded everywhere else.

**Impact**

- The saved `mean_spatial_loss.pt` contains loss values at boundary nodes that the model never optimises for, which can mislead downstream error analysis.
- `plot_spatial_error` picks `vmin`/`vmax` from the full tensor, so boundary-node loss values can skew the colour scale.
- For LAM domains with boundary forcing the inconsistency is most pronounced; for global domains (no boundary) it has no practical effect.

**Minimal example**

```bash
python -m neural_lam.train_model --eval test --config_path <config> --load <ckpt>
# spatial_loss_maps saved to <logger_dir>/spatial_loss_maps/ include boundary values
```

**Proposed fix**

After computing spatial loss over the full grid (to preserve the `num_grid_nodes` shape needed by plotting), mask boundary nodes to NaN:

```python
spatial_loss = self.loss(
    prediction, target, pred_std, average_grid=False
)
spatial_loss[..., ~self.interior_mask_bool] = float("nan")
```

And use `torch.nanmean` instead of `torch.mean` when averaging the maps in `on_test_epoch_end`.

NoiceHax:
## Describe your changes

`test_step` built its spatial loss map over the whole grid, while every other
loss and metric call in `ForecasterModule` restricts itself to the interior via
`interior_mask_bool`. `ARForecaster` overwrites the boundary nodes with the true
boundary state, so their loss is always ~0, and those zeros were dragging down
the saved `mean_spatial_loss.pt` and flattening the colour range of the plotted
loss maps.

This NaN-masks the boundary right after the loss call and reduces with
`torch.nanmean` in `on_test_epoch_end`. Passing `mask=self.interior_mask_bool`
into `self.loss` is not an option: `metrics.mask_and_reduce_metric`
index-selects with `metric_entry_vals[..., mask, :]`, which shrinks the grid
dimension and breaks the shape `vis.plot_spatial_error` reshapes back onto the
grid. `vis.py` needs no change, it already uses `np.nanmin`/`np.nanmax` and
matplotlib draws NaN cells transparent.

The fix is @RajdeepKushwaha5's from #568, following @sadamov's review there.
#568 edits `neural_lam/models/ar_model.py`, which no longer exists after the
`ForecasterModule` split in #208, so it cannot be rebased onto main. The new
test is the first coverage of `test_step` / `on_test_epoch_end`. It fails on
main and passes here.

No new dependencies. Written with help from Claude Code.

## Issue Link

closes #569

## Type of change

- [x] ðŸ› Bug fix (non-breaking change that fixes an issue)
- [ ] âœ¨ New feature (non-breaking change that adds functionality)
- [ ] ðŸ’¥ Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] ðŸ“– Documentation (Addition or improvements to documentation)

- [ ] I have updated the [README](README.MD) to cover introduced code changes - not needed, nothing user-facing changes

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Exclude boundary nodes from the spatial loss maps computed in `test_step`, so the plotted loss maps and the saved `mean_spatial_loss.pt` cover the interior only, consistent with every other loss and metric call in `ForecasterModule` [\#720](https://github.com/mllam/neural-lam/pull/720) @NoiceHax
+- Exclude boundary nodes from the spatial loss maps computed in `test_step`, so the plotted loss maps and the saved `mean_spatial_loss.pt` cover the interior only, consistent with every other loss and metric call in `ForecasterModule` [\#720](https://github.com/mllam/neural-lam/pull/720) @RajdeepKushwaha5 @NoiceHax
 
 - Fix `RuntimeError` in `HiLAMParallel` forward pass on hierarchical graphs by offsetting edge indices into the global mesh node index space ([#679](https://github.com/mllam/neural-lam/issues/679))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Exclude boundary nodes from the spatial loss maps computed in `test_step`, so the plotted loss maps and the saved `mean_spatial_loss.pt` cover the interior only, consistent with every other loss and metric call in `ForecasterModule` [\#720](https://github.com/mllam/neural-lam/pull/720) @NoiceHax
+
 - Fix `RuntimeError` in `HiLAMParallel` forward pass on hierarchical graphs by offsetting edge indices into the global mesh node index space ([#679](https://github.com/mllam/neural-lam/issues/679))
 
 - Fix `IndexError` in HiLAM forward pass by offsetting grid nodes in `zero_index_g2m`/`zero_index_m2g` by the total mesh-node count across all levels ([#642](https://github.com/mllam/neural-lam/issues/642)) @Sir-Sloth-The-Lazy
--- a/neural_lam/models/module.py
+++ b/neural_lam/models/module.py
@@ -611,6 +611,11 @@ class ForecasterModule(pl.LightningModule):
         spatial_loss = self.loss(
             prediction, target_states, pred_std, average_grid=False
         )
+        # Exclude boundary nodes, consistent with the loss used in
+        # training/validation. NaN-mask in place so the tensor keeps its
+        # (B, pred_steps, num_grid_nodes) shape for downstream plotting,
+        # and is reduced with `torch.nanmean` in `on_test_epoch_end`.
+        spatial_loss[..., ~self.interior_mask_bool] = float("nan")
         log_spatial_losses = spatial_loss[
             :,
             [
@@ -935,7 +940,8 @@ class ForecasterModule(pl.LightningModule):
             torch.cat(self.spatial_loss_maps, dim=0)
         )
         if self.trainer.is_global_zero:
-            mean_spatial_loss = torch.mean(spatial_loss_tensor, dim=0)
+            # `nanmean` because boundary nodes are NaN-masked in `test_step`
+            mean_spatial_loss = torch.nanmean(spatial_loss_tensor, dim=0)
 
             loss_map_figs = [
                 vis.plot_spatial_error(
--- a/neural_lam/models/module.py
+++ b/neural_lam/models/module.py
@@ -611,10 +611,6 @@ class ForecasterModule(pl.LightningModule):
         spatial_loss = self.loss(
             prediction, target_states, pred_std, average_grid=False
         )
-        # Exclude boundary nodes, consistent with the loss used in
-        # training/validation. NaN-mask in place so the tensor keeps its
-        # (B, pred_steps, num_grid_nodes) shape for downstream plotting,
-        # and is reduced with `torch.nanmean` in `on_test_epoch_end`.
         spatial_loss[..., ~self.interior_mask_bool] = float("nan")
         log_spatial_losses = spatial_loss[
             :,
--- a/tests/test_training.py
+++ b/tests/test_training.py
@@ -1,15 +1,12 @@
 # Standard library
 import warnings
-from functools import cached_property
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -50,6 +50,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Exclude boundary nodes from the spatial loss maps computed in `test_step`, so the plotted loss maps and the saved `mean_spatial_loss.pt` cover the interior only, consistent with every other loss and metric call in `ForecasterModule` [\#720](https://github.com/mllam/neural-lam/pull/720) @NoiceHax
```
sadamov: The fix is originally @RajdeepKushwaha5's from #568, could we credit you both here?

```suggestion
- Exclude boundary nodes from the spatial loss maps computed in `test_step`, so the plotted loss maps and the saved `mean_spatial_loss.pt` cover the interior only, consistent with every other loss and metric call in `ForecasterModule` [\#720](https://github.com/mllam/neural-lam/pull/720) @RajdeepKushwaha5 @NoiceHax
```

sadamov (approved):
Thanks for picking this up. I verified locally: the new test fails on main with the boundary nodes sitting at exactly `0.0`, and passes with the fix, so it is real regression coverage.

This stays relevant after #636, which only deprecates `crop_to_interior` and always crops the axes to the interior bbox. The `vrange=None` call sites still take `vmin`/`vmax` from the full grid, so the boundary zeros keep pinning `vmin` to 0, and `mean_spatial_loss.pt` is still saved over the whole grid.

One request on the CHANGELOG below, and then this is ready from my side.

CHANGELOG.md
NoiceHax: Absolutely, Its only fair to give credit where its due.

neural_lam/models/module.py
```diff
@@ -611,6 +611,11 @@ def test_step(self, batch, batch_idx):
         spatial_loss = self.loss(
             prediction, target_states, pred_std, average_grid=False
         )
+        # Exclude boundary nodes, consistent with the loss used in
+        # training/validation. NaN-mask in place so the tensor keeps its
+        # (B, pred_steps, num_grid_nodes) shape for downstream plotting,
+        # and is reduced with `torch.nanmean` in `on_test_epoch_end`.
```
sadamov: remove
```suggestion
```

tests/test_training.py
```diff
@@ -227,3 +231,115 @@ def all_gather(self, tensor, sync_grads=False):
...
+        mask_2d = np.zeros((n_points_1d, n_points_1d), dtype=int)
+        mask_2d[0, :] = 1
+        mask_2d[-1, :] = 1
+        mask_2d[:, 0] = 1
+        mask_2d[:, -1] = 1
+        # `grid_index` is the stacked ("x", "y") dimension, so a C-order
+        # flatten of an (x, y) array lines up with it.
+        return xr.DataArray(mask_2d.reshape(-1), dims=["grid_index"])
+
+
```
sadamov:
```suggestion

@pytest.mark.slow
```

tests/test_training.py
```diff
@@ -227,3 +231,115 @@ def all_gather(self, tensor, sync_grads=False):
...
+    def boundary_mask(self) -> xr.DataArray:
+        n_points_1d = int(np.sqrt(self.num_grid_points))
+        mask_2d = np.zeros((n_points_1d, n_points_1d), dtype=int)
+        mask_2d[0, :] = 1
+        mask_2d[-1, :] = 1
+        mask_2d[:, 0] = 1
+        mask_2d[:, -1] = 1
+        # `grid_index` is the stacked ("x", "y") dimension, so a C-order
+        # flatten of an (x, y) array lines up with it.
+        return xr.DataArray(mask_2d.reshape(-1), dims=["grid_index"])
```
sadamov: `test_plotting.py:365` already frames a boundary this way, by assigning datastore.ds["boundary_mask"] and popping the cached_property. Could we lift that into tests/dummy_datastore.py and drop the subclass? #636 rewrites that plotting test too, so a shared helper saves a third copy.

something like_
```
def set_framed_boundary(
    datastore: DummyDatastore, width: int = 1
) -> np.ndarray:
    """Replace the random boundary mask with a frame of `width` cells."""
    shape = (datastore.grid_shape_state.x, datastore.grid_shape_state.y)
    mask = np.ones(shape, dtype=int)
    mask[width:-width, width:-width] = 0
    # `grid_index` is the stacked ("x", "y") dimension, so a C-order flatten
    # of an (x, y) array lines up with it.
    datastore.ds["boundary_mask"] = xr.DataArray(
        mask.reshape(-1), dims=["grid_index"]
    )
    datastore.__dict__.pop("boundary_mask", None)  # bust the cached_property
    return mask
```
Then here, which also drops the now unused cached_property, numpy and xarray imports:


```
def test_test_step_excludes_boundary_from_spatial_loss(tmp_path):
    ...
    datastore = DummyDatastore(n_grid_points=400, n_timesteps=10)
    set_framed_boundary(datastore)
```

sadamov (commented):
@NoiceHax before merging with main I carried out my default list of review points and found some additional nitpicks. could you please have a look

NoiceHax:
Addressed all the nitpicks.

- Dropped the redundant comment in module.py.
- Lifted set_framed_boundary into tests/dummy_datastore.py and dropped the subclass plus its now-unused numpy / xarray / cached_property imports. The plotting test can reuse the helper too.
- Marked the new test @pytest.mark.slow.
- Also added @RajdeepKushwaha5 to the CHANGELOG entry from the earlier request.

black and isort pass on the touched files. I still can't run the suite locally, so the fix and test logic are unchanged from what you verified.

sadamov (approved):
great! merging this fix now, thanks @NoiceHax and @RajdeepKushwaha5

PR #741, opened 2026-09-02
Add property-based tests for output clamping

### LINKED ISSUE #602: Add property-based testing with hypothesis

KumarShivam1908:
## Describe your changes

Adds `hypothesis` as a dev dependency and covers the output clamping in
`tests/test_clamping.py` with property-based tests, replacing the hand-built
out-of-bounds cases they subsume.

The design follows @aliviahossain's comment on #602 from 2026-07-13, which
lays out this approach point for point: the invariant is exact and
magnitude-independent, arbitrary starting states are valid inputs because
`inverse_sigmoid`/`inverse_softplus` clamp internally, the unclamped features
and the zero-delta identity are separate properties worth keeping, and the
model construction has to move into a wider-scoped fixture.

**Why this is worth it.** I mutation-tested the new tests rather than assuming
they work. Applying the sigmoid clamp to only the first grid point
(`new_state[:, 0, idx]`) — a realistic broadcasting slip that would silently
corrupt every grid point but one in a live forecast — leaves the old
`test_clamping` **passing**, because it only ever uses a `(1, 1, 4)` tensor
with a single grid point. The property test catches it and shrinks to a
minimal two-grid-point reproducer. That is a bug class the example-based test
is structurally unable to see.

### Changes after review

- Deleted both `invalid_state` blocks. As @sadamov points out, the property
  test already draws `prev_state` anywhere in ±1e3, so hand-building an
  out-of-range state adds nothing. The suite stays green and still catches the
  injected grid bug.
- Split the chained `lower <= x <= upper` comparisons in the remaining
  example-based test. Python expands those to `and`, which calls `bool()` on
  the tensor and raises as soon as a second feature gets both bounds.
- `unscaled_prediction` is now recomputed after the negative-delta loop. It
  was computed once from the positive-delta prediction, so the physical-space
  assertions below it were re-checking the earlier result.
- Dropped `derandomize=True`. The falsifying example is printed on failure, so
  a red run stays reproducible without freezing the drawn examples — that
  argument is better than the one I had.
- **Source fix:** the clamp index buffers now use `dtype=torch.long`.
  `torch.tensor([])` yields a float tensor, so any config leaving a clamping
  category unused raised
  `IndexError: tensors used as indices must be long, int, byte or bool tensors`.
  Found by @sadamov while checking one of the latent issues. Covered by
  `test_empty_clamp_category_indices_are_long`, which fails without the fix.
- Added a `Property-based tests` section to `CONTRIBUTING.md` so this sets the
  pattern rather than being a one-off.

On the third latent issue I raised: @sadamov is right that it does not hold.
`not (...).any()` is already `all(not ...)`, so it asserts what was intended.
Two of the three were real, and both are fixed here.

`uv.lock` is regenerated because CI installs with `--locked`; it adds only
`hypothesis` and `sortedcontainers`. Plain `hypothesis` is enough —
`hypothesis.extra.numpy` just needs numpy, already a core dependency. No CI or
workflow changes.

**Verification**

```
pytest --doctest-modules tests/test_clamping.py -q            # 4 passed
pytest tests/test_prediction_model_classes.py tests/test_graph_efm_predictor.py \
       tests/test_probabilistic_objectives.py tests/test_clamping.py -q   # 24 passed
pre-commit run --files <changed files>                        # all hooks pass
```

Remaining follow-up: dataset slicing (`test_time_slicing.py`), where the
`if/elif` ladder raising `NotImplementedError` is what blocks the parametrize
list from growing. Left out here — replacing it means computing the expected
values, and if that formula restates `weather_dataset`'s own offset logic the
test goes green while proving nothing. Worth agreeing the approach first.

## Issue Link

Part of #602

## Type of change

- [x] If this PR changes contributor workflow, I've updated CONTRIBUTING.md / AGENTS.md

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


tests/test_clamping.py
```diff
@@ -290,3 +315,109 @@ class ModelArgs:
...
+    allow_infinity=False,
+)
+
+# `derandomize` fixes the drawn examples, so a given commit always runs the
+# same inputs. That trades hypothesis' open-ended search for determinism,
+# on the reasoning that a property test which can fail on an unrelated PR
+# tends to get disabled rather than fixed.
+# `deadline=None` because per-example timing of torch ops on shared CI
+# runners is unreliable, and the 200 ms default would only add flakiness.
+_SETTINGS = settings(deadline=None, derandomize=True)
```
sadamov: Dropping `derandomize` (False is the default, so the kwarg just goes).

```suggestion
# The search is left random on purpose: a derandomized property test only
# ever runs the examples drawn at authoring time, which gives up the one
# thing hypothesis is for. When it does fail, the falsifying example is
# printed, so reproduce with that rather than disabling the test.
# `deadline=None` because per-example timing of torch ops on shared CI
# runners is unreliable, and the 200 ms default would only add flakiness.
_SETTINGS = settings(deadline=None)
```

CHANGELOG.md
```diff
@@ -95,6 +95,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Maintenance
 
+- Add `hypothesis` as a dev dependency and cover the output clamping in `tests/test_clamping.py` with property-based tests, asserting that the clamped state stays within its configured bounds and that unbounded features keep the plain residual update, for any previous state and any delta. The existing example-based test is unchanged; its setup moved to a module-scoped fixture so the model is built once rather than per generated example. [\#741](https://github.com/mllam/neural-lam/pull/741) @KumarShivam1908
```
sadamov: One entry, not two. I put it under `Fixed` rather than `Maintenance` because the dtype bug is the only half a user can hit; the test work is discoverable from the PR link.

```suggestion
- Fix `IndexError` when an output-clamping config leaves a clamping category unused: the index buffers were registered from an empty `torch.tensor([])`, which is float32 and cannot index. Also adds `hypothesis` as a dev dependency and covers the clamping bounds with property-based tests in `tests/test_clamping.py` [\#741](https://github.com/mllam/neural-lam/pull/741) @KumarShivam1908
```

sadamov (commented):
Nice PR @KumarShivam1908 , the mutation test reproduces: with the clamp on `new_state[:, 0, idx]` the old test still passes and yours fails and shrinks.

Since this PR introduces hypothesis, I would like it to set the pattern too, so I suggest to add a short CONTRIBUTING.md section on when a property test is worth writing. Block 7 below.

I would drop `derandomize=True`, suggestion inline. Frozen examples mean the test only ever runs what was drawn at authoring time, and the falsifying example is printed anyway, so a red run stays reproducible.

It also grows `tests/test_clamping.py` from 292 to 423 lines, where #602 argued hypothesis would shrink it. The two `invalid_state` blocks are redundant by your own reasoning: the property test already draws an out-of-range `prev_state`, so hand-building one tests nothing extra. Deleting both (72 lines) keeps the suite green and still catches the injected grid bug. Could we do that here?

Two of your three latent issues are real and mostly land inside those blocks. The third does not hold, `not (...).any()` is already `all(not ...)`. Checking the second turned up a source bug worth folding in: the clamp index buffers come from `torch.tensor(...)`, and an empty list gives float32, which cannot index, so any config that leaves a clamping category unused raises `IndexError`.

Could you also credit @aliviahossain in the description? Their comment on #602 from 2026-07-13 lays out this design point for point.

GitHub will not let me attach suggestions to lines outside the diff, so the rest are plain blocks below rather than one-click suggestions.

1. Delete lines 201-256. Subsumed by `test_clamped_features_stay_within_bounds`, which draws `prev_state` anywhere in +/-1e3 and already covers out-of-bounds starting states for all three clamp categories.

2. Delete lines 295-317, same for the negative direction.

3. Lines 153-161. Chained `lower <= x <= upper` raises as soon as a second feature gets both bounds, since Python expands it to `and` and calls `bool()` on the tensor. Confirmed by adding `v100m` to both limit dicts.

```python
    sigmoid_prediction = prediction[:, :, model.clamp_lower_upper_idx]
    assert (sigmoid_prediction >= model.sigmoid_lower_lims).all().item()
    assert (sigmoid_prediction <= model.sigmoid_upper_lims).all().item()
```

4. Lines 267-277. Same split, plus `unscaled_prediction` is computed at line 174 from the positive-delta prediction and never recomputed, so the physical-space assertions just below re-check the earlier result.

```python
    sigmoid_prediction = prediction[:, :, model.clamp_lower_upper_idx]
    assert (sigmoid_prediction >= model.sigmoid_lower_lims).all().item()
    assert (sigmoid_prediction <= model.sigmoid_upper_lims).all().item()

    # Check that prediction is within bounds in original non-normalized space
    unscaled_prediction = prediction * model.state_std + model.state_mean
```

5. `neural_lam/models/step_predictors/base.py` lines 285-293.

```python
        self.register_buffer(
            "clamp_lower_upper_idx",
            torch.tensor(sigmoid_lower_upper_idx, dtype=torch.long),
        )
        self.register_buffer(
            "clamp_lower_idx",
            torch.tensor(softplus_lower_idx, dtype=torch.long),
        )
        self.register_buffer(
            "clamp_upper_idx",
            torch.tensor(softplus_upper_idx, dtype=torch.long),
        )
```

6. Regression test for it. 

```python
def test_empty_clamp_category_indices_are_long(datastore_and_graph):
    """Clamp index buffers stay integer tensors when a category is empty."""
    datastore, graph_name = datastore_and_graph

    # Only lower limits, so both the sigmoid and the upper softplus
    # categories come out empty.
    model = GraphLAM(
        datastore=datastore,
        graph_name=graph_name,
        hidden_dim=4,
        hidden_layers=1,
        processor_layers=2,
        mesh_aggr="sum",
        num_past_forcing_steps=1,
        num_future_forcing_steps=1,
        output_std=False,
        output_clamping_lower={"t2m": 0.0},
        output_clamping_upper={},
    )

    assert model.clamp_lower_upper_idx.numel() == 0
    assert model.clamp_upper_idx.numel() == 0

    for idx in (
        model.clamp_lower_upper_idx,
        model.clamp_lower_idx,
        model.clamp_upper_idx,
    ):
        assert idx.dtype == torch.long

    # Indexing with an empty buffer must work rather than raise IndexError.
    features = datastore.get_vars_names(category="state")
    state = torch.zeros(1, 1, len(features))
    assert state[:, :, model.clamp_upper_idx].numel() == 0

    new_state = model.get_clamped_new_state(torch.ones_like(state), state)
    assert torch.isfinite(new_state).all()
```

7. CONTRIBUTING.md, new section after "Before you push".

```markdown
## Property-based tests

Some tests use [hypothesis](https://hypothesis.readthedocs.io/) to generate
inputs rather than hardcode them. `tests/test_clamping.py` is the reference
example.

Write one when a rule must hold across a whole input space and the existing
tests pin a single arbitrary point of it. Do not write one for specific
input/output pairs, physical reference values or regression cases: there is no
invariant to state, and the generator scaffolding costs more than it returns.

- Set `deadline=None`; torch op timing on shared CI runners is too unreliable
  for the 200 ms default.
- Do **not** set `derandomize=True`. It freezes the examples to those drawn at
  authoring time, so the test stops finding anything new.
- Fixtures used by a `@given` test must be module-scoped or wider. Hypothesis
  rejects function-scoped ones, and the body re-runs per generated example.

When a property fails, reproduce with the falsifying example hypothesis prints
and fix the cause. Do not disable the test or pin a seed.
```

sadamov:
also make sure to use the full pr template, the current one is cutoff

leifdenby:
Maybe we could build on https://hypothesis-torch.readthedocs.io/en/latest/quick_start_guide/ and make the amount of changes fewer and clearer?

sadamov:
note from the dev-meeting: the general consensus was that this is not a priority and currently on hold.
- The test files actually got larger not smaller with hypothesis
- We didn't see the benefit of randomized test. (WeatherGenerator allows for very diverse inputs and output, maybe that's where they get the benefit and we don't)
This could be revisited if someone champions the PR and clearly shows which cases benefit from hypothesis in what fashion.

Issue #409, opened 2026-03-17
[Bug] AttributeError in PaddedWeatherDataset.get_original_window_indices when using --distributed


RajdeepKushwaha5:
### Description

`PaddedWeatherDataset.get_original_window_indices()` in `neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py`
passes `step_length.total_seconds()` (a `float`) to `get_integer_time()`, which internally calls `.total_seconds()` on its argument — expecting a
`datetime.timedelta`. This causes a crash when running standardization stats computation with `--distributed` on multi-GPU / SLURM setups.

### Error

```
AttributeError: 'float' object has no attribute 'total_seconds'
```

### Root cause

The call on line 51 redundantly converts the `timedelta` to a `float` before passing it to `get_integer_time()`, which then tries to call `.total_seconds()` on the float and crashes:

```python
# Line 51 — BUGGY: passes float to get_integer_time()
step_int, _ = get_integer_time(step_length.total_seconds())

# Line 310 — CORRECT: passes timedelta directly
time_step_int, time_step_unit = get_integer_time(step_length)
```

`get_integer_time()` in `utils.py` always calls `tdelta.total_seconds()` internally, so it must receive a `timedelta`, not a `float`.

### Steps to reproduce

Run `compute_standardization_stats` with `--distributed` on a multi-process SLURM setup using the `npyfilesmeps` datastore. The crash occurs at rank 0 inside the diff-stats gathering block (line 366).

### Fix

Remove the redundant `.total_seconds()` call so the `timedelta` is passed directly, consistent with the other call site at line 310:

```python
step_int, _ = get_integer_time(step_length)
```

AshNicolus:
This was fixed incidentally by #411 - PaddedWeatherDataset.get_original_window_indices was renamed to get_original_indices, and the only remaining get_integer_time call site (compute_standardization_stats.py:428) now passes step_length directly, not step_length.total_seconds(). No .total_seconds() call remains anywhere in the file. Closing as resolved.

PR #743, opened 2026-09-04
Fix wrong lead-time unit in plot_error_heatmap axis label

nikhil3495:
## Describe your changes

`plot_error_heatmap` (`neural_lam/vis.py`) built the lead-time x-axis label from `time_step_unit[0]` - the first character of the unit word returned by `utils.get_integer_time` (`weeks`/`days`/`hours`/`minutes`/`seconds`/`milliseconds`/`microseconds`/`unknown`). That is only correct for `weeks`/`days`/`hours`/`seconds`:

- `minutes` -> `Lead time (m)` (should be `min`; `m` reads as metres/months)
- `unknown` (step length not an exact whole number of seconds) -> `Lead time (u)` (meaningless)
- `milliseconds` / `microseconds` -> `Lead time (m)`

Every other call site (`neural_lam/models/module.py:780`, `:1019`) uses the full unit word; only this one truncated.

This maps the full unit name to a proper abbreviation via a module-level `_LEAD_TIME_UNIT_ABBREVIATIONS` dict, and labels the `unknown` case `steps` since the tick values are then plain step indices. Display-only change - no numbers change.

Added a parametrized test over `hours`/`days`/`minutes`/`seconds`/`unknown` step lengths and a drift-guard test that every unit name `get_integer_time` can return has an abbreviation.

### Testing / verification

- `pre-commit run --all-files` passes.
- `pytest tests/test_plotting.py --doctest-modules neural_lam/vis.py` passes, including the new parametrized abbreviation test and the drift-guard test.
- Manually inspected `plot_error_heatmap` output for `minutes`- and `unknown`-unit datasets to confirm the axis label now reads `Lead time (min)` / `Lead time (steps)` instead of the truncated single-letter label.

## Issue Link

closes #742

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
[...]

PR #748, opened 2026-09-12
Fix silent stat corruption in distributed standardization-stats depadding

### LINKED ISSUE #749: compute_standardization_stats.py --distributed silently corrupts saved stats via positional depadding after gather
## Description

`compute_standardization_stats.py`'s `--distributed` path pads `WeatherDataset`
(via `PaddedWeatherDataset`) so its length divides evenly across ranks, then
after the multi-rank gather "depads" by taking a positional prefix:

```python
means, squares = (
    [means_gathered_tensor[i] for i in original_indices],
    [squares_gathered_tensor[i] for i in original_indices],
)
```

(and equivalently `diff_means_gathered_tensor[:n_original_windows]` for the
diff stats), where `original_indices = list(range(total_samples))`. This
assumes the padded rows land at the tail of `gathered_tensor`.

They don't. `DistributedSampler(shuffle=False)` assigns this rank the dataset
indices `range(len(dataset))[rank::world_size]` — a stride of `world_size`,
not a contiguous block. So whenever `total_samples % world_size != 0`, the
padded indices (`>= total_samples`) land on different ranks depending on
`padded_index % world_size`, not all on the same rank. After
`all_gather_object` concatenates the per-rank results rank-major, the
padding is scattered through the middle of the tensor, not sitting at the
tail.

Taking the first `total_samples` entries then silently keeps a few padded
(duplicated last-sample) rows from one rank while dropping an equal number
of real rows from another — corrupting the saved `parameter_mean.pt`,
`parameter_std.pt`, `diff_mean.pt` and `diff_std.pt` used to standardize the
whole dataset for training.

## Reproduction

Using this repo's actual `torch.utils.data.distributed.DistributedSampler`
with `total_samples=101, world_size=4, batch_size=8`
(`padded_samples=3`, padded length 104):

```python
from torch.utils.data.distributed import DistributedSampler

class DS:
    def __init__(self, n): self.n = n
    def __len__(self): return self.n

total_samples, world_size = 101, 4
padded_len = 104  # PaddedWeatherDataset's computed length
ds = DS(padded_len)

per_rank = [list(DistributedSampler(ds, num_replicas=world_size, rank=r, shuffle=False))
            for r in range(world_size)]
flattened = [idx for rank_list in per_rank for idx in rank_list]

naive_selected = flattened[:total_samples]
print("padded rows wrongly INCLUDED:", [i for i in naive_selected if i >= total_samples])
print("real rows wrongly DROPPED:", [i for i in range(total_samples) if i not in naive_selected])
```

Output:

```
padded rows wrongly INCLUDED: [101, 102]
real rows wrongly DROPPED: [95, 99]
```

Not covered by existing tests — `tests/test_compute_standardization_stats.py`
(added in #411) only checks that slicing preserves tensor *shape* on
synthetic data; it never simulates an actual multi-rank
`DistributedSampler`, which is why this gap went uncaught by that PR's
otherwise-thorough fix for the neighboring shape/`IndexError` bugs
(#409/#412/#413).

## Fix

Identify real vs. padded rows locally, per rank, *before* the gather —
using the sampler's own (already deterministic) per-rank index order — and
filter each minibatch's contribution as soon as it's computed, instead of
trying to reconstruct sample identity positionally after the gather. Order
doesn't matter for a mean/std reduction, so no reordering is needed once
padding is dropped at the source.

nikhil3495:
## Describe your changes

`compute_standardization_stats.py --distributed` pads `WeatherDataset` (via `PaddedWeatherDataset`) so its length divides evenly across ranks, then after the multi-rank gather "depads" by taking `gathered[:total_samples]` (equivalently `gathered[i] for i in original_indices]`). That assumes padded rows land at the tail of the gathered, rank-major-concatenated tensor.

They don't: `DistributedSampler(shuffle=False)` stripes dataset indices across ranks (rank `r` gets `r, r+world_size, r+2*world_size, ...`), so whenever `total_samples % world_size != 0`, the padded rows are scattered across several ranks' local tails rather than one global tail. The prefix-selection then silently keeps a few padded (duplicated last-sample) rows while dropping an equal number of real ones — corrupting the saved `parameter_mean/std.pt` and `diff_mean/std.pt` tensors used to standardize the whole dataset.

I confirmed this with a standalone simulation using this repo's actual `DistributedSampler`: for `total_samples=101, world_size=4, batch_size=8`, the old logic swaps real samples `95` and `99` for padded samples `101` and `102`.

**Fix**: identify real vs. padded rows locally, per rank, *before* the gather, instead of trying to reconstruct sample identity after it. `real_sample_mask_per_batch()` derives — from pure index arithmetic on the sampler's already-deterministic iteration order — which rows of each minibatch are real, with no dependency on gather order. Both the parameter-stats and diff-stats accumulators are filtered by this mask as soon as each batch is computed, so the post-gather concatenation already contains exactly the real samples, in whatever order (order doesn't matter for a mean/std reduction). This also removes the need for the previous `get_original_indices()` / `n_original_windows` positional bookkeeping in `main()`.

Not a duplicate of #409/#412/#413 (all closed via #411): those fixed a shape/`IndexError` bug in the same file's gather logic, not this sample-identity mismatch. The tests #411 added only check that slicing preserves shape/values on synthetic tensors — they never simulate an actual multi-rank `DistributedSampler`, which is why this gap went uncaught.

No new dependencies required.

## Issue Link

closes #749

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation (Addition or improvements to documentation)

- [ ] I have requested a reviewer and an assignee. I don't have write access to the repo — happy to have a maintainer add these.

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

- [x] I have added a line to the CHANGELOG describing this change, under *Fixed*.

## Checklist for assignee

- [ ] the tests pass

---

Generated with Claude Code


neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -73,6 +73,57 @@ def get_original_indices(self) -> list[int]:
...
+        One boolean tensor per minibatch, ``True`` for rows that came from a
+        real (non-padded) sample.
+    """
+    rank_indices = list(sampler)
+    return [
+        torch.tensor(
+            [idx < total_samples for idx in rank_indices[i : i + batch_size]]
+        )
+        for i in range(0, len(rank_indices), batch_size)
+    ]
```
sadamov, line 124: I suggest to make this a method: `batch_size` and `total_samples` are already here, and `len % world_size == 0` (needed, else `DistributedSampler` pads by repeating the first real indices, which `idx < total_samples` cannot see) holds by construction. `split` gives the same chunks as `DataLoader(drop_last=False)`. Docstring trimmed, the bug story is in the CHANGELOG.

```suggestion

    def real_sample_masks(
        self, sampler: DistributedSampler
    ) -> list[torch.Tensor]:
        """
        Return one boolean mask per minibatch, ``True`` for non-padded rows.

        ``DistributedSampler`` stripes indices across ranks, so the padded
        tail of this dataset lands on several ranks and cannot be sliced off
        the rank-major gathered tensor. Each rank drops its padded rows with
        these masks instead; the sampler order chunked by ``batch_size`` is
        the order a ``DataLoader(drop_last=False)`` yields batches in.

        Parameters
        ----------
        sampler : DistributedSampler
            Sampler of this rank's ``DataLoader`` over this dataset.

        Returns
        -------
        list[torch.Tensor]
            Boolean mask per minibatch.
        """
        is_real = torch.tensor(list(sampler)) < self.total_samples
        return list(is_real.split(self.batch_size))
```

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -303,8 +354,20 @@ def main(
...
+    # Precomputed so real (non-padded) rows can be dropped locally, per
+    # rank, as soon as each batch is processed -- see
+    # `real_sample_mask_per_batch` for why depadding after the gather below
+    # is not safe to do positionally.
+    real_masks = None
+    if distributed:
+        assert sampler is not None
+        real_masks = real_sample_mask_per_batch(
+            sampler, batch_size, len(ds_base)
+        )
```
sadamov, line 366: The comment here said "real rows can be dropped", the opposite of what the mask does.

```suggestion
    real_masks = None
    if distributed:
        assert isinstance(ds, PaddedWeatherDataset) and sampler is not None
        real_masks = ds.real_sample_masks(sampler)
```

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -316,10 +379,14 @@ def main(
...
         flux_batch = forcing_batch[:, :, :, 0]
         # (B, d_features,)
-        means.append(torch.mean(batch, dim=(1, 2)).cpu())
-        squares.append(
-            torch.mean(batch**2, dim=(1, 2)).cpu()
-        )  # (B, d_features,)
+        batch_means = torch.mean(batch, dim=(1, 2)).cpu()
+        batch_squares = torch.mean(batch**2, dim=(1, 2)).cpu()
+        if real_masks is not None:
+            mask = real_masks[batch_i]
```
sadamov, line 385:
```suggestion
        if real_masks is not None:
            # Drop padded rows per rank, see `real_sample_masks`
            mask = real_masks[batch_i]
```

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -334,19 +401,13 @@ def main(
...
-
-            assert isinstance(ds, PaddedWeatherDataset)
-            original_indices = ds.get_original_indices()
-            means, squares = (
-                [means_gathered_tensor[i] for i in original_indices],
-                [squares_gathered_tensor[i] for i in original_indices],
-            )
+            # Padding rows were already dropped per rank above, so the
+            # gathered tensors already hold exactly the real samples (order
+            # doesn't matter for a mean/std reduction).
```
sadamov, line 406: "order doesn't matter" describes `save_stats`, not this line.

```suggestion
            # Padded rows were already dropped per rank above
```

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -432,9 +493,20 @@ def main(
...
+    # per rank, right after each batch's diffs are computed. Each real row
+    # of `batch` produces `time_step_int` rows in `stepped_batch` /
+    # `batch_diffs` (one per `ss_i` block, stacked along dim 0 in that
+    # order), so the per-batch mask is repeated to match.
+    real_masks_standard = None
+    if distributed:
+        assert sampler_standard is not None
+        real_masks_standard = real_sample_mask_per_batch(
+            sampler_standard, batch_size, len(ds_standard_base)
+        )
```
sadamov, line 506: Same as 2, the `repeat` note moves next to the call (6).

```suggestion
    real_masks_standard = None
    if distributed:
        assert (
            isinstance(ds_standard, PaddedWeatherDataset)
            and sampler_standard is not None
        )
        real_masks_standard = ds_standard.real_sample_masks(sampler_standard)
```

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -456,10 +528,16 @@ def main(
...
         batch_diffs = stepped_batch[:, 1:] - stepped_batch[:, :-1]
         # (B', N_t-1, num_grid_nodes, d_features)
-        diff_means.append(torch.mean(batch_diffs, dim=(1, 2)).cpu())
+        batch_diff_means = torch.mean(batch_diffs, dim=(1, 2)).cpu()
         # (B', d_features,)
-        diff_squares.append(torch.mean(batch_diffs**2, dim=(1, 2)).cpu())
+        batch_diff_squares = torch.mean(batch_diffs**2, dim=(1, 2)).cpu()
         # (B', d_features,)
+        if real_masks_standard is not None:
+            mask = real_masks_standard[batch_i].repeat(time_step_int)
```
sadamov, line 536:
```suggestion
        if real_masks_standard is not None:
            # stepped_batch stacks the time_step_int ss_i blocks along dim 0,
            # so tile the mask (repeat, not repeat_interleave)
            mask = real_masks_standard[batch_i].repeat(time_step_int)
```

neural_lam/datastore/npyfilesmeps/compute_standardization_stats.py
```diff
@@ -473,18 +551,17 @@ def main(
...
-            )
-            assert isinstance(ds_standard, PaddedWeatherDataset)
-            n_original_windows = (
-                len(ds_standard.get_original_indices()) * time_step_int
-            )
-            diff_means = [diff_means_gathered_tensor[:n_original_windows]]
-            diff_squares = [diff_squares_gathered_tensor[:n_original_windows]]
+            # Padding rows were already dropped per rank above (see the
+            # parameter-stats loop for why depadding by position, after the
+            # gather, is not safe to do here).
```
sadamov, line 556:
```suggestion
            # Padded rows were already dropped per rank above
```

tests/test_compute_standardization_stats.py
```diff
@@ -1,9 +1,11 @@
 # Third-party
 import torch
+from torch.utils.data.distributed import DistributedSampler
 
 # First-party
 from neural_lam.datastore.npyfilesmeps.compute_standardization_stats import (
     PaddedWeatherDataset,
+    real_sample_mask_per_batch,
 )
```
sadamov, line 9:
```suggestion
# Third-party
import pytest
import torch
from torch.utils.data import DataLoader
from torch.utils.data.distributed import DistributedSampler

# First-party
from neural_lam.datastore.npyfilesmeps.compute_standardization_stats import (
    PaddedWeatherDataset,
)
```

tests/test_compute_standardization_stats.py
```diff
@@ -114,3 +116,108 @@ def test_slice_handles_larger_step_length(self):
...
+    def test_no_padding_needed_keeps_every_row(self):
+        """When length is already divisible by world_size, nothing pads."""
+        total_samples, world_size, batch_size = 16, 4, 4
+        ds = PaddedWeatherDataset(
+            _IndexEchoDataset(total_samples), world_size, batch_size
+        )
+
+        recovered = _per_rank_real_dataset_indices(ds, world_size, batch_size)
+
+        assert sorted(recovered) == list(range(total_samples))
```
sadamov, line 223: `_per_rank_real_dataset_indices` re-implements the function's own chunking, so it cannot catch a mismatch with what `DataLoader` yields (chunking by `batch_size + 1` in both still passes). Real `DataLoader` instead; pads echo `total_samples - 1`, so a leaked one is a duplicate. The count test is implied by the set equality.

```suggestion
# -- Bug 3: positional depadding after distributed gather ------------------


class _IndexEchoDataset:
    """Dataset stub whose items echo their own index."""

    def __init__(self, n_samples: int):
        self._n = n_samples

    def __len__(self) -> int:
        return self._n

    def __getitem__(self, idx: int) -> torch.Tensor:
        return torch.tensor(idx)


class TestRealSampleMasks:
    """Per-rank masks drop exactly the padded rows the DataLoader yields."""

    @pytest.mark.parametrize(
        "total_samples, world_size, batch_size",
        [
            (16, 4, 4),  # no padding
            (103, 4, 8),  # one padded row, on rank 3 only
            (101, 4, 8),  # old prefix slice kept pads 101, 102, dropped 95, 99
            (97, 4, 8),  # each pad is alone in a 1-row last batch (empty mask)
            (10, 4, 4),  # per-rank count below batch_size
        ],
    )
    def test_masked_batches_recover_each_real_sample_once(
        self, total_samples, world_size, batch_size
    ):
        """Pads echo total_samples - 1, so a leaked pad shows as a duplicate."""
        ds = PaddedWeatherDataset(
            _IndexEchoDataset(total_samples), world_size, batch_size
        )
        recovered = []
        for rank in range(world_size):
            sampler = DistributedSampler(
                ds, num_replicas=world_size, rank=rank, shuffle=False
            )
            loader = DataLoader(ds, batch_size, sampler=sampler)
            masks = ds.real_sample_masks(sampler)
            assert len(masks) == len(loader)
            for batch, mask in zip(loader, masks):
                recovered.extend(batch[mask].tolist())

        assert sorted(recovered) == list(range(total_samples))
```

CHANGELOG.md
```diff
@@ -50,6 +50,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  output: depadding after the multi-rank gather selected
+  `gathered[:total_samples]` (equivalently `gathered[original_indices]`),
+  which assumes padded rows land at the tail of the gathered tensor. They
+  don't -- `DistributedSampler(shuffle=False)` stripes dataset indices
+  across ranks, so a naive prefix selection silently swapped a handful of
+  real samples for padded (duplicated last-sample) ones whenever
+  `total_samples % world_size != 0`, corrupting the saved parameter/diff
+  mean and std tensors. Real vs. padded rows are now identified locally,
+  per rank, before the gather (`real_sample_mask_per_batch`), so order no
+  longer matters. @nikhil3495
```
sadamov, line 63: One line with the PR link, as the other entries.

```suggestion
- Fix `compute_standardization_stats.py --distributed` silently keeping padded rows and dropping real ones when `len(dataset) % world_size != 0`. Padded rows are now masked out per rank before the gather. [\#748](https://github.com/mllam/neural-lam/pull/748) @nikhil3495
```

sadamov (changes requested):
Nice catch, reproduced with a real `DataLoader` (101/4/8 loses samples 95 and 99 and counts sample 100 three times). Suggestions below, please rebase first, the CHANGELOG conflicts with #743.

Outside the diff so not suggestible: `get_original_indices`, `original_indices` and `padded_indices` have no callers left after this PR (`__getitem__` can return `self.base_dataset[min(idx, self.total_samples - 1)]`), I would drop them and `test_original_indices` too.

Below I am mostly talking about shorter comments and docstrings.

nikhil3495:
Thanks for the detailed review! I've addressed all points:
- Converted real_sample_mask_per_batch into the real_sample_masks method 
  on PaddedWeatherDataset
- Removed the now-dead original_indices/padded_indices/get_original_indices 
  and test_original_indices
- Fixed the reversed/misleading comments
- Rewrote the Bug 3 tests to use a real DataLoader with your parametrized cases
- Shrunk the CHANGELOG entry to one line
- Rebased onto main to resolve the CHANGELOG conflict

All pushed. Let me know if anything else needs adjusting!

PR #751, opened 2026-09-15
Keep a copy of the DANRA test data between CI runs

### LINKED ISSUE #257: Test/CI: Implement offline mocking or local caching for "mdp" datastore tests to prevent network-dependent failures
Hi team! I am currently diving into the Neural-LAM codebase and preparing my local development environment for the GSoC 2026 application period.

While running the pytest suite locally, I noticed that all tests relying on the mdp datastore fail if the user's network drops, times out, or blocks external connections. Specifically, the tests hang and eventually fail when trying to load data from the remote ECMWF object store (https://object-store.os-api.cci1.ecmwf.int/.../height_levels.zarr).

The Problem:
Currently, the mdp test suite strictly relies on an active, stable internet connection to fetch real data. Network-dependent tests can lead to flaky CI pipelines and create a frustrating onboarding experience for new contributors whose local firewalls might block these specific requests.

Proposed Solution:
I would love to contribute a PR to make the test suite robust and offline-capable. I see two potential paths to solve this, and I'd like the maintainers' feedback on which is preferred:

Approach A (Mocking): Intercept the network calls during testing using unittest.mock (or a library like responses), returning a minimal, valid xarray/zarr object in memory so the data-loading logic can be tested without actual HTTP requests.

Approach B (Local Dummy Data): Similar to how the dummydata tests pass cleanly offline, we could generate a tiny, static dummy .zarr file (e.g., 10-50 KB) that strictly mimics the mdp structure. This file would be kept in the repository's test fixtures, allowing the mdp datastore to be tested entirely locally.

Impact:
Implementing this will speed up the test suite execution time, eliminate flaky CI runs caused by server timeouts, and make local development much smoother for the community.

I have my local environment fully configured and would love to take this on as my first contribution. Do the maintainers agree this is an issue worth addressing, and if so, do you prefer Approach A or Approach B?

KumarShivam1908:
## Describe your changes

Every CI run downloads the DANRA example data used by the `mdp` tests again, about 92 MB. This is slow, and if the data server has a problem, the tests fail even though the code is fine.

This PR keeps a copy of that data between CI runs. As agreed in #257, it keeps a copy of the **input** data that `mllam-data-prep` reads, not the finished `danra.datastore.zarr`. This way `mllam-data-prep` still runs in every test run.

### How it works

1. The test config now puts `simplecache::` in front of each DANRA address. This tells fsspec (the library that downloads the files): "before you download a file, check if you already have a copy; if not, download it and keep a copy."
2. On CI, the workflow tells fsspec which folder to keep the copies in (`.fsspec-cache`). It asks GitHub to bring that folder back before the tests, and to store it after the tests if it was not there yet.
3. On your own computer nothing changes. fsspec keeps the copies in a temporary folder for that run only.

### What changed

- `tests/datastore_examples/mdp/danra_100m_winds/danra.datastore.yaml`: add `simplecache::` in front of the 4 input paths.
- `tests/conftest.py`: one line, `fsspec.config.conf["https"] = {"asynchronous": True}`. With zarr 3, `simplecache::https://` crashes the first time the data is built unless https is set to async. `mllam-data-prep` has no place in its config for this setting yet, so, as @observingClouds suggested, it is set once for all tests.
- `.github/workflows/install-and-test.yml`:
  - Set `FSSPEC_SIMPLECACHE_CACHE_STORAGE` to the folder for the copies.
  - A step before the tests brings the folder back.
  - A step after the tests stores it, only if it was not brought back.
  - The existing MEPS steps are unchanged.
- `CHANGELOG.md`: one entry under Maintenance.

### When the stored copy is replaced

The stored copy has a name made of three parts:
- the operating system;
- a version number, `v1`. Change it to `v2` to force a fresh download;
- a fingerprint of the files `tests/datastore_examples/mdp/**/*.datastore.yaml`. If someone changes which data the tests use, the name changes and the data is downloaded again.

Only the `mdp` configs are part of the fingerprint. The MEPS data is unpacked during the tests and adds its own `*.datastore.yaml` under `npyfilesmeps/`. If those files were included, the name before the tests would differ from the name after the tests, and the copy would never be reused.

### Timings

Measured on my machine (Windows, CPU), `pytest tests/test_datastores.py -k mdp` (18 tests):

| | Time |
|---|---|
| No copy yet (downloads everything) | 5 min 10 s |
| Copy already there | 8 s |

### How I checked it

- The 18 `mdp` tests pass with no copy and with a copy.
- The full `pytest -vv -s --doctest-modules` passes (287 tests).
- pre-commit passes.

### Good to know

- The first CI run on a branch still downloads everything and then stores the copy. Later runs use it.
- GitHub only shares stored copies from `main` with other branches, so other PRs get faster once this is merged.
- If the tests fail, the copy is not stored in that run. The existing MEPS copy works the same way.
- The CPU and GPU jobs store the copy under the same name. The job that finishes second only shows a warning that the copy already exists. This also already happens for MEPS.

**Dependencies:** none. #606 also edits `install-and-test.yml`, so whichever PR merges second may need a small conflict fix. The ERA5 data from #635 can use the same approach (`simplecache::gs://...`) once it is merged. Its config is under `mdp/`, so it is already part of the fingerprint.

## Issue Link

closes #257

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass

@sadamov @observingClouds @leifdenby  could one of you approve the CI run? This PR can only be fully tested on GitHub.


observingClouds:
This looks very neat. I'd be curious if this implementation is also what @leifdenby had in mind.

observingClouds:
OK, so the actions ran two days ago once and I just allowed them to run again after your latest commit and it now restored the cache it created previously:

Today's run:
image

Two days ago:
image

So this looks good, but there are two caches created (I guess from the CPU and GPU CI) and for some reason their size also differs:
image

KumarShivam1908:
> OK, so the actions ran two days ago once and I just allowed them to run again after your latest commit and it now restored the cache it created previously:
> 
> 
> 
> Today's run:
> 
> image
> 
> 
> 
> Two days ago:
> 
> image
> 
> 
> 
> So this looks good, but there are two caches created (I guess from the CPU and GPU CI) and for some reason their size also differs:
> 
> image

Let me check from my side about this and will inform you.

KumarShivam1908:
> OK, so the actions ran two days ago once and I just allowed them to run again after your latest commit and it now restored the cache it created previously:
> 
> Today's run: image
> 
> Two days ago: image
> 
> So this looks good, but there are two caches created (I guess from the CPU and GPU CI) and for some reason their size also differs: image

 I feel it's because of how GitHub tells caches apart: it looks at the key and the folder path. The CPU and GPU runners have different folder paths (/home/runner/... vs /home/runnerx/...), so each job saved its own cache under the same key.

The sizes probably differ because each machine compresses the files a little differently. The data itself should be the same, since both jobs ran the same tests.

If you'd like a single shared cache, I can switch the cache path to the relative .fsspec-cache.

observingClouds:
Yes, please only create one common cache as the data should be the same.

KumarShivam1908:
> Yes, please only create one common cache as the data should be the same.

Please check now , should work well .

observingClouds:
Thanks for the edit. It seems to work now. Only one cache is created. There is a bit of a race condition when the cache gets first created, but as long as we do not run the test matrix sequentially this is hard to solve.

Screenshot 2026-09-18 at 12 58 34

KumarShivam1908:
> Thanks for the edit. It seems to work now. Only one cache is created. There is a bit of a race condition when the cache gets first created, but as long as we do not run the test matrix sequentially this is hard to solve.
> 
> Screenshot 2026-09-18 at 12 58 34

Yeah, true. Let me know if you have something in mind for this, or we can leave it for now.

observingClouds:
> There is a bit of a race condition when the cache gets first created, but as long as we do not run the test matrix sequentially this is hard to solve.

Keeping this is fair as it does only print a warning and does not stop the CI from running. In addition, this case only appears when the cache needs to be rebuild, which should be rather seldom.

.github/workflows/install-and-test.yml
```diff
@@ -75,6 +77,15 @@ jobs:
...
 
+      # Input data for the mdp examples, downloaded through fsspec simplecache.
+      # The cache resets when an mdp datastore config changes or v1 is bumped.
+      - name: Load mdp input cache
+        id: mdp-input-cache
+        uses: actions/cache/restore@v4
+        with:
+          path: .fsspec-cache
+          key: ${{ runner.os }}-mdp-input-v1-${{ hashFiles('tests/datastore_examples/mdp/**/*.datastore.yaml') }}
+
```
observingClouds, line 87: So in #753 you mention the switch to actions/cache/cache instead of having restore and save. Should this also done in this PR?

KumarShivam1908: Yeah, good call. Switching to `actions/cache@v4` actually makes the PR simpler since it already skips saving when there’s a cache hit. So we can remove the `id:` and `if: cache-hit` checks as well.

KumarShivam1908:
There are two PRs now, so here's what's in each.

#754 is the new one, based on `main`, and this is the actual fix. The MEPS tarball had one step to fetch the cached copy and another step to save it. The save step didn't check if we had already found a cache, which is what caused the duplicate copies. I replaced both with a single `actions/cache@v4` step, which automatically skips saving when it finds a cache. This is what was causing the 18 duplicate copies and the 5.12 GB of storage.

#751 is the existing one, and I've made the same change for the MDP cache there. Nothing was actually wrong with it because it already had the check, but it was using the same two step setup. It seemed better to clean it up now instead of coming back to it later. It also got smaller, going from 13 lines to 5.

Once both are merged, both cache steps in the file will follow the same approach, and there won't be any manual save logic left.

For pre commit, there's nothing to change. I checked and `pre-commit/action` already uses the combined cache step internally. The extra copies are just old ones from before `main` had those cache keys, and they should clear out on their own.

Two things to keep in mind:

The change only stops new duplicates from being created. The 18 existing ones will go away gradually as those PRs become inactive, rather than disappearing as soon as this is merged.

Both PRs remove a step from the same place, so whichever one gets merged second will need a small rebase. I'm happy to do that.

.github/workflows/install-and-test.yml
```diff
@@ -75,6 +77,14 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-meps-reduced-example-data-v0.3.0
 
+      # Input data for the mdp examples, downloaded through fsspec simplecache.
+      # The cache resets when an mdp datastore config changes or v1 is bumped.
```
observingClouds: This is obvious from the code and also too specific as the comment would need updating if the version would be bumped.
```suggestion
```

observingClouds (approved):
LGTM! Let's see what @leifdenby has to add as he proposed the simplecache idea.

KumarShivam1908:
> LGTM! Let's see what @leifdenby has to add as he proposed the simplecache idea.

yeah :-)

PR #115, opened 2025-02-04
Fix how torch version is derived in ci/cd and clean up github actions

leifdenby:
## Describe your changes

Fixes to ci/cd setup to ensure that correct torch version is installed when testing with cpu/gpu and pdm/pip. And refactor to have single github actions script (rather than four as before).

## Issue Link

https://github.com/mllam/neural-lam/issues/114

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [ ] I have performed a self-review of my code
- [ ] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [ ] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have updated the [README](README.MD) to cover introduced code changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging). This applies only if you have write access to the repo, otherwise feel free to tag a maintainer to add a reviewer and assignee.

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [x] the code is readable
- [x] the code is well tested
- [x] the code is documented (including return types and parameters)
- [x] the code is easy to maintain

## Author checklist after completed review

- [ ] I have added a line to the CHANGELOG describing this change, in a section
  reflecting type of change (add section where missing):
  - *added*: when you have added new functionality
  - *changed*: when default behaviour of the code has been changed
  - *fixes*: when your contribution fixes a bug

## Checklist for assignee

- [ ] PR is up to date with the base branch
- [ ] the tests pass
- [ ] author has added an entry to the changelog (and designated the change as *added*, *changed* or *fixed*)
- Once the PR is ready to be merged, squash commits and merge the PR.

PR #754, opened 2026-09-18
Use one actions/cache step so PRs stop storing duplicate test data

### LINKED ISSUE #753: GitHub Action cache approaching limit of 10GB
Looking at https://github.com/mllam/neural-lam/actions/caches it appears that we are approaching GitHub's limit on cache size. This is not immediately an issue as GitHub will automatically delete the least recently used caches.

The issue looks rather that the caches are not properly shared/restored because I assume that all of the `Linux-meps-reduced-example-data-v3.0.0` files should be the same:

Image

KumarShivam1908:
## Describe your changes

The Actions cache is at 9.5 GB of the 10 GB limit, and 5.12 GB of that is 18 byte-identical copies of the same 291 MB MEPS tarball — one per open PR. They have the same size and the same cache version hash, differing only in the ref they are attached to.

The cause is that the workflow hand-writes the restore and the save as two separate steps, and the save step has no condition. GitHub scopes cache writes per branch: a pull request can read `main`'s cache but can only write into its own slot. So every PR restores `main`'s copy, then saves an identical duplicate under `refs/pull/<n>/merge`. It has been that way since #115, and looks accidental rather than deliberate — the `restore-keys` block was a verbatim copy of `key`, so it never provided a fallback.

This is already costing us. At 9.5 GB, least-recently-used eviction is active, and `main`'s copy is the likeliest to go because `main` runs least often. When it does, every PR misses and re-downloads 291 MB from the ECMWF object store.

The fix is to use the combined `actions/cache` step, which does the restore and the save and already skips the save when the restore was an exact hit:

```yaml
      - name: Cache meps example data
        uses: actions/cache@v4
        with:
          path: tests/datastore_examples/npyfilesmeps/meps_example_reduced.v0.3.0.tar.gz
          key: ${{ runner.os }}-meps-reduced-example-data-v0.3.0
```

The separate `Save cache data` step and the redundant `restore-keys` block are removed with it.

I preferred this over adding an `if: cache-hit != 'true'` guard to the existing save step. Both are the same size, but the guard leaves the pattern in place for the next cache someone adds — which is roughly how the missing condition survived unnoticed for so long.

### What does not change

- The cache is still restored at the same point, before the tests.
- Nothing is saved when an earlier step fails. The combined action behaves the same way here as the current implicit `success()` on the save step.
- The CPU and GPU jobs still race to save on a cold cache, and the loser still logs a harmless `cache already exists` warning, as discussed in #751.

### How to verify after merge

Re-run any PR and check the cache list. There should be no new `Linux-meps-reduced-example-data-v0.3.0` entry under its `refs/pull/*/merge` — the 18 existing copies will then age out on their own.

### Related

#751 makes the same change for the mdp input cache it adds, so both caches end up in one consistent form.

Also checked while investigating: `pre-commit/action` already uses the combined `actions/cache@v4` internally, so the per-Python-version caches are doing the right thing and need no change.

## Issue Link

closes #753

## Type of change

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass


GiGiKoneti:
@KumarShivam1908 
Please use PR template

KumarShivam1908:
@GiGiKoneti have Updated , thanks

CHANGELOG.md
```diff
@@ -105,6 +105,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Maintenance
 
+- Use the combined `actions/cache` step for the meps test-data cache in CI, so a pull request that restores the existing cache no longer saves a duplicate 291 MB copy under its own branch [\#754](https://github.com/mllam/neural-lam/pull/754) @KumarShivam1908
```
observingClouds: Let's make this a bit less specific
```suggestion
- Use the combined `actions/cache` step for the meps test-data cache in CI, so a PR that restores the existing cache no longer saves a duplicate copy under its own branch [\#754](https://github.com/mllam/neural-lam/pull/754) @KumarShivam1908
```

observingClouds (approved):
LGTM! Thanks @KumarShivam1908 for your contribution.

